PR #3, opened 2015-12-17
Romanesco task for TIFF creation

zachmullen:


server/create_tiff.py
```diff
@@ -0,0 +1,25 @@
...
+
+out_path = os.path.join(_tempdir, out_filename)
+
+convert_command = (
+    'vips',
+    'tiffsave',
+    '"%s"' % in_path,
+    '"%s"' % out_path,
+    '--compression', 'jpeg',
+    '--Q', '90',
```
brianhelba: Can we make this ``90`` a parameter called quality?

server/create_tiff.py
```diff
@@ -0,0 +1,26 @@
...
+    '--tile',
+    '--tile-width', '256',
```
brianhelba: Can we make this ``256`` a parameter called ``tile_size``?

brianhelba: Both ``256``s should use the same parameter value.

server/create_tiff.py
brianhelba: Also, we should eventually store these parameters, in addition to the fact that it was created with this specific version of this script somewhere in the item (or file?) metadata, to provide provenance.

It's not critical that we get this into the first version of this functionality. We can always add it later.

zachmullen:
Just ran this on girder.neuro using the gauss machine as the romanesco worker, everything seems to be working. Does this tiff seem OK?

https://girder.neuro.emory.edu/#item/56730e9ab0e9575c8ab1c900

cooperlab:
Zach,

I requested a login - if you grant access then I can take a look.

-Lee

________________________________

zachmullen:
@cooperlab you should now have access to that item.

dgutman:
I am wondering if the encode parameters were odd... I can't open thie file
with IrfanView or ScanScope....




[image: pasted1]

zachmullen:
@brianhelba any idea what may be going wrong with this TIFF? I'm going to go ahead and get the provenance stuff sorted out on this branch, I have an idea of how to do it.

dgutman:
sorry ill have to look into it tonight.. got side tracked

brianhelba:
@zachmullen : I spoke with @dgutman about this, and it looks like the way we're building VIPS is causing SVS files to fail. For right now, you can probably test it with a large PNG or the like.

If this branch is ready to merge, please feel free to do so. We can discuss the fixes to the VIPS building directly.

dgutman:
I am actually working on this now... I'll let everyone know when I get the
sever patched

brianhelba:
@dgutman Please keep some notes / documentation of everything you need to do to build VIPS / OpenJPEG / etc. Once we get it working the first time, I want to move my Ansible script from https://github.com/ImageMarkup/isic-archive/blob/master/ansible/roles/vips/tasks/main.yml to this repo and update it to include the latest build procedure.

server/__init__.py
```diff
@@ -17,14 +17,37 @@
...
+    """
+    file = event.info['file']
+    if 'itemId' not in file:
+        return
+
+    itemModel = ModelImporter.model('item')
+    item = itemModel.load(file['itemId'], force=True, exc=True)
+
+    if item.get('expectedLargeImage') and (file['name'].endswith('.tiff') or
+            file.get('mimeType') == 'image/tiff'):
```
brianhelba: If the file doesn't actually have its ``mimeType`` field set to ``image/tiff`` here, we need to set it and update the file.

Logic in the POST endpoint to short-circuit the conversion process depends on this field being correct for TIFF files.

PR #4, opened 2015-12-21
Update to geojs 0.6.0-rc1.

manthey:
For files that have non 256-pixel square tiles, this still needs to be adjusted.

### DIFF
--- a/Gruntfile.js
+++ b/Gruntfile.js
@@ -36,7 +36,7 @@ module.exports = function (grunt) {
                 pnltri: '<%= plugin.large_image.geojs_modules %>/pnltri/pnltri.js',
                 proj4: '<%= plugin.large_image.geojs_components %>/proj4/dist/proj4-src.js',
                 d3: '<%= plugin.large_image.geojs_components %>/d3/d3.js',
-                glmatrix: '<%= plugin.large_image.geojs_components %>/gl-matrix/gl-matrix.js'
+                glmatrix: '<%= plugin.large_image.geojs_components %>/gl-matrix/dist/gl-matrix.js'
             }
         },
         uglify: {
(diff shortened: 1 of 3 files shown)


aashish24:
Yay! 

> For files that have non 256-pixel square tiles, this still needs to be adjusted.

@manthey can you explain bit more? what needs to be adjusted?

manthey:
We get a sizeX and sizeY parameter from the large_image endpoint.  We are not using it yet -- it needs to be set on the layer parameters in an appropriate way.  If we have such a pyramidal tif file, I'd be happy to hook it up, but I didn't want to make it look like it was supported without testing it.

We also can support formats with overlapping tiles, if we have any samples of those.

aashish24:
> We get a sizeX and sizeY parameter from the large_image endpoint. We are not using it yet -- it needs to be set on the layer parameters in an appropriate way. If we have such a pyramidal tif file, I'd be happy to hook it up, but I didn't want to make it look like it was supported without testing it

Ah, I see. I will see if I can generate those tifs for you, unless you are looking into it. 


> We also can support formats with overlapping tiles, if we have any samples of those.

This one can wait I think.

brianhelba:
@manthey So, you'd like me to provide a 128x128 or 512x512 tiled file?

manthey:
I suppose an even better test would be to have tiles that are neither square nor a power of two in size.  This may cause issues in WebGL, which is why it would be a more interesting test.  I'll try this out in the next day or two.

brianhelba:
@manthey We definitely need to support non-256 square tiles. From my perspective, we do not need to support non-square tiles (not that I object to more features).

Note that [TIFF files only support tile sizes that are a multiple of 16](http://www.awaresystems.be/imaging/tiff/tifftags/tilewidth.html), so that's also the only other hard requirement that we need to support in the short-term.

PR #5, opened 2016-01-07
Added additional tile test options.

manthey:
Confirmed that this works with non-square tiles (and tiles of any size).

The test tile server now can take a variety of query parameters:
-  min: minimum tile level (default 0)
-  max: maximum tile level (default levels parameter - 1)
-  fractal: if 'true', draw a simple fractal on the tiles (only on square power-of-two tiles)
-  w: tile width in pixels (default tileSize parameter)
-  h: tile height in pixels (default tileSize parameter)
-  sizex: image width in pixels at maximum level.
-  sizey: image height in pixels at maximum level.

The tile size is percolated through to the viewer javascript as tileWidth and tileHeight, if available.  If additional viewers are extended to handle non-square tiles, they should check if these values are set and fall back to tileSize if not.

The following viewers work with non-square tiles:
-  GeoJS
-  OpenLayers
-  OpenSeaDragon

Leaflet claims it supports non-square tiles by using L.Point for the tileSize, but the code which detects retina displays breaks when it encounters this, so it doesn't work as simply as the documentation claims.  I haven't tried to use non-square tiles in SlideAtlas.

Also, autodetect PNGs and send the right mime type for them (otherwise we assume JPGs).  Using PNGs with the fractal tiles looks a lot better than the compression artifacts on JPGs.

I also adjusted some formatting to make it closer to the girder standards.

### DIFF
--- a/server/rest.py
+++ b/server/rest.py
@@ -25,7 +25,8 @@ from girder.api.describe import describeRoute, Description
 from girder.api.rest import loadmodel, RestException
 from girder.models.model_base import AccessType
 
-from .tilesource import TestTileSource, TiffGirderTileSource, TileSourceException
+from .tilesource import TestTileSource, TiffGirderTileSource, \
+    TileSourceException
 
 
 class TilesItemResource(Item):
@@ -42,21 +43,20 @@ class TilesItemResource(Item):
         apiRoot.item.route('GET', (':itemId', 'tiles', 'zxy', ':z', ':x', ':y'),
                            self.getTile)
 
-
-    def _loadTileSource(self, itemId):
+    def _loadTileSource(self, itemId, params):
         try:
             if itemId == 'test':
-                tileSource = TestTileSource(256)
+                tileSource = TestTileSource(256, params=params)
             else:
                 # TODO: cache the user / item loading too
-                item = self.model('item').load(id=itemId, level=AccessType.READ,
-                                               user=self.getCurrentUser(), exc=True)
+                item = self.model('item').load(
+                    id=itemId, level=AccessType.READ,
+                    user=self.getCurrentUser(), exc=True)
                 tileSource = TiffGirderTileSource(item)
             return tileSource
         except TileSourceException as e:
             # TODO: sometimes this could be 400
             raise RestException(e.message, code=500)
-
 
     @describeRoute(
         Description('Get large image metadata.')
@@ -66,9 +66,8 @@ class TilesItemResource(Item):
     )
     @access.public
     def getTilesInfo(self, itemId, params):
-        tileSource = self._loadTileSource(itemId)
+        tileSource = self._loadTileSource(itemId, params)
         return tileSource.getMetadata()
-
 
     @describeRoute(
         Description('Create a large image for this item.')
@@ -113,7 +112,6 @@ class TilesItemResource(Item):
             'created': True
         }
 
-
     @describeRoute(
         Description('Remove a large image from this item.')
         .param('itemId', 'The ID of the item.', paramType='path')
@@ -133,13 +131,15 @@ class TilesItemResource(Item):
             'deleted': deleted
         }
 
-
     @describeRoute(
         Description('Get a large image tile.')
         .param('itemId', 'The ID of the item or "test".', paramType='path')
-        .param('z', 'The layer number of the tile (0 is the most zoomed-out layer).', paramType='path')
-        .param('x', 'The X coordinate of the tile (0 is the left side).', paramType='path')
-        .param('y', 'The Y coordinate of the tile (0 is the top).', paramType='path')
+        .param('z', 'The layer number of the tile (0 is the most zoomed-out '
+               'layer).', paramType='path')
+        .param('x', 'The X coordinate of the tile (0 is the left side).',
+               paramType='path')
+        .param('y', 'The Y coordinate of the tile (0 is the top).',
+               paramType='path')
         .errorResponse('ID was invalid.')
         .errorResponse('Read access was denied for the item.', 403)
     )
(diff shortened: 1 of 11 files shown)


server/tilesource/test.py
```diff
@@ -35,14 +35,66 @@
 
 
 class TestTileSource(TileSource):
-    # TODO: move this class to its own file, and only import if PIL v3+ is available
-    def __init__(self, tileSize=256, levels=10):
+    # TODO: move this class to its own file, and only import if PIL v3+ is
```
brianhelba: You can remove this comment.

server/rest.py
```diff
@@ -151,11 +151,14 @@ def getTile(self, itemId, z, x, y, params):
...
 
-        tileSource = self._loadTileSource(itemId)
+        tileSource = self._loadTileSource(itemId, params)
         try:
             tileData = tileSource.getTile(x, y, z)
         except TileSourceException as e:
             raise RestException(e.message, code=404)
 
-        cherrypy.response.headers['Content-Type'] = 'image/jpeg'
+        if tileData[:8] == '\x89PNG\r\n\x26\n':
```
brianhelba: I think the TileSource should make the tile encoding type available via a method, instead of peeking into binary data in the API layer.

server/tilesource/test.py
```diff
@@ -35,14 +35,66 @@
...
         super(TestTileSource, self).__init__()
-        self.tileSize = tileSize
-        self.levels = levels
-        self.sizeX = (2 ** (self.levels - 1)) * self.tileSize
-        self.sizeY = (2 ** (self.levels - 1)) * self.tileSize
+        self.tileWidth = self.tileHeight = tileSize
+        self.maxlevel = levels - 1
+        self.minlevel = 0
+        self.fractal = False
+        if params:
```
brianhelba: My understanding of Girder idioms is that it's the job of the API layer to unpack parameters first.

``minLevel`` (please change to be camelCase), ``fractal``, etc. should all be individual keyword parameters to ``TestTileSource.__init__``. The API then does the ``.get`` and passes them each as arguments directly.

If you were trying to keep ``TestTileSource.__init__`` backwards-comparable, don't worry about it, this plugin is still very early in development. Please feel free to aggressively change any APIs that need it.

server/tilesource/test.py
```diff
@@ -62,17 +114,23 @@ def getTile(self, x, y, z):
...
             v=(0.3 + (0.7 * yFraction)),
         )
+        rgbColor = tuple(int(val * 255) for val in backgroundColor)
 
         image = Image.new(
             mode='RGB',
-            size=(self.tileSize, self.tileSize),
-            color=tuple(int(val * 255) for val in backgroundColor)
+            size=(self.tileWidth, self.tileHeight),
+            color=(rgbColor if not self.fractal else (255, 255, 255))
```
brianhelba: Can we make the tile background always use ``rbgColor``, and fractals always be white?

manthey: Fundamentally, they are.  The fractal is drawn by filling the file with the fractal color and then drawing the non-fractal parts recursively.  I don't know of a more efficient way to draw this in PIL.

server/tilesource/test.py
```diff
@@ -84,5 +142,5 @@ def getTile(self, x, y, z):
         )
 
         output = StringIO()
-        image.save(output, 'JPEG')
+        image.save(output, 'PNG')
```
brianhelba: Real tiles are almost always JPEG-encoded in photographic images. I'm fine with PNG as a default for TestTileSource, but let's make it a parameter (something like ``encoding='png'``), so we can use JPEG if necessary.

web_client/js/imageViewerWidget/base.js
```diff
@@ -8,6 +8,8 @@ girder.views.ImageViewerWidget = girder.View.extend({
         }).done(_.bind(function (resp) {
             this.levels = resp.levels;
             this.tileSize = resp.tileSize;
```
brianhelba: I think we should just get rid of ``tileSize`` everywhere now. A viewer can always choose to fail to initialize if it doesn't support non-square tiles.

server/tilesource/test.py
brianhelba: Ah, I think I misunderstood which parts were "inside" the fractal.

web_client/js/imageViewerWidget/base.js
brianhelba: Accordingly, go ahead and put a check in the initialization of Leaflet and have it fail if ``this.tileWidth != this.tileHeight``.

manthey: Will do.

manthey:
The remaining issue that you saw (not loading in FF) is caused by using $.getScript to load vgl.  This will require updating vgl to fix properly.  vgl attempts to determine the url of where it was loaded from in case you use a function to load GL shader code from relative paths.

brianhelba:
@manthey Any possibility that my issue with Chrome+Windows was caused by the fact that Windows was running in a VM with garbage graphics hardware?

manthey:
If the VM doesn't have access to webGL I could see that being the problem.
Otherwise, I could easily image that different versions of Chrome (or even
loading in Chrome with the development tools open) could affect the
behavior based on the vgl issue I found.

On Thu, Jan 7, 2016 at 3:08 PM, Brian Helba <notifications@github.com>
wrote:

> @manthey <https://github.com/manthey> Any possibility that my issue with
> Chrome+Windows was caused by the fact that Windows was running in a VM with
> garbage graphics hardware?
>
> —
> Reply to this email directly or view it on GitHub
> <https://github.com/DigitalSlideArchive/large_image/pull/5#issuecomment-169791702>
> .
>



-- 
David Manthey
R&D Engineer
Kitware Inc.
(518) 881-4439

jbeezley:
@brianhelba I don't what issue you had exactly, but I have had nearly zero success getting webgl to work inside a VM.  It is nearly hopeless with virtualbox, but you might be able to get it to work with VMWare.

brianhelba:
@jbeezley Thanks for the info. I have no need / expectation for WebGL to work in a VM, but it should fall back to a Canvas render gracefully whenever WebGL fails. If that is failing to occur here (as a separate problem from the vgl + ``$.getScript`` issue), it should be fixed.

brianhelba:
@manthey Once you change the ``console.log`` to ``console.error``, this LGTM. Feel free to merge your own stuff.

PR #6, opened 2016-01-09
Fix issues with tileSize

brianhelba:
* Validate TestTileSource parameters in the REST API layer
 * Internal APIs should mostly assume that they are being called with
the correct types.
 * Necessary type conversion that's required due the limitations of HTTP
parameters is really an additional part of HTTP deserialization, and
should be performed by the REST API layer.
* Make TIFFTileSource use tileWidth and tileLength, instead of tileSize
 * This fixes a bug where the TileSource.getMetadata was returning "null"
for the tileWidth and tileHeight.
 * Since non-square tiles are now supported, this is simpler than trying to
also sometimes still use tileSize.
*  Use a BytesIO to store image data

### DIFF
--- a/server/rest.py
+++ b/server/rest.py
@@ -46,15 +46,24 @@ class TilesItemResource(Item):
     def _loadTileSource(self, itemId, params):
         try:
             if itemId == 'test':
-                tileSource = TestTileSource(
-                    minLevel=params.get('minLevel'),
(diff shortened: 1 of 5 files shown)


brianhelba:
@manthey PTAL. Let me know if you disagree with anything.

manthey:
LGTM.

PR #8, opened 2016-01-13
Setup testing infrastructure

jbeezley:
@brianhelba Here is the boilerplate testing setup you asked for.  A few notes:
* The travis config might not work.  I can't test it until travis integration is turned on.
* All but one of the style tests currently fail.  You may want to adjust the style configs (to for example, add global variables), but I wanted to wait on the ESLint PR before worrying about it.
* The client test doesn't work because the plugin source files aren't loaded into the test html.  I haven't dealt with client side plugin tests, so I'm not sure what is wrong with it.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,0 +1,41 @@
+language: python
+python:
+    - "2.7"
+
+cache:
+    directories:
+        - $HOME/.cache
+
+sudo: false
+
+compiler:
+    - gcc
+
+before_install:
+    - main_path=$PWD
+    - build_path=$PWD/build
+    - girder_path=$build_path/girder
+    - mkdir -p $build_path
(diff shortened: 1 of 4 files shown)


brianhelba:
That was fast, thanks! 

I see that the Git SHA of upstream Girder is hardcoded into the Travis script. Is that the current tip of ``master``? I certainly want some of the most recent Girder changes available, but after the next release, I can change it to be a tag.

I'll work on enabling Travis soon.

jbeezley:
Yes, that is the most recent girder master.  It's best to checkout a specific version to avoid tests passing one day and failing the next.

brianhelba:
@jbeezley FYI, I got this working last night. I made some additional commits to [get the build working](https://github.com/DigitalSlideArchive/large_image/commit/c10616975dd452554f053e301142e1982a9d5905) and [tests passing](https://github.com/DigitalSlideArchive/large_image/compare/c10616975dd452554f053e301142e1982a9d5905...22ecd28d2c242dbf1bf3b47bd24542aaabe4f876),

jbeezley:
Great! :clap:

PR #13, opened 2016-02-10
Fix a few bugs.

manthey:
The lru cache doesn't work with @staticmethod without some additional checks.

The element name was changed to an erroneous value.

Also, made the version for numpy >= rather than strictly == to avoid conflict with another plugin.


brianhelba:
LGTM. Merge away.

PR #14, opened 2016-02-10
Upgrade geojs version

manthey:


brianhelba:
LGTM, but it appears that this is based on #13 . Merge away, as long as you can keep Git clean.

manthey:
I cleaned up the history so this change is by itself.

PR #15, opened 2016-02-15
Fix some issues in creating and deleting tile sets.

manthey:
Instead of testing if a file is a tile file, check if it has pyramidal tiles.  If not, run it through the conversion process.

Only check that there are at least 3 channels rather than exactly 3.  The PHOTOMETRIC check ensures that the first three channels are what we expect.  This allows, for instance, uploading a PNG and using it despite having an unused opacity channel.

When we fail to find tiles in a file, include the last exception as part of the error message.  This makes it easier to determine why we couldn't handle the tiles (for instance, because it is in the wrong color space).

### DIFF
--- a/server/rest.py
+++ b/server/rest.py
@@ -122,9 +122,10 @@ class TilesItemResource(Item):
                 raise RestException('"fileId" must be a file on the same item'
                                     'as "itemId".')
             try:
-                TiffGirderTileSource(None, largeImageFile['_id'])
                 item['largeImage'] = largeImageFile['_id']
+                TiffGirderTileSource(item)
             except TileSourceException:
+                del item['largeImage']
                 job = self._createLargeImageJob(largeImageFile, item)
                 item['expectedLargeImage'] = True
                 item['largeImageOriginalId'] = largeImageFileId
--- a/server/rest.py
+++ b/server/rest.py
@@ -241,7 +241,7 @@ class TilesItemResource(Item):
                 }
 
             # If this file was created by the worker job, delete it
-            if 'largeImageJobId':
+            if 'largeImageJobId' in item:
                 # The large image file should not be the original file
                 assert item['largeImageOriginalId'] != item['largeImage']
 
@@ -292,6 +292,9 @@ class TilesItemResource(Item):
             x, y, z = int(x), int(y), int(z)
         except ValueError:
             raise RestException('x, y, and z must be integers', code=400)
+        if x < 0 or y < 0 or z < 0:
+            raise RestException('x, y, and z must be positive integers',
+                                code=400)
 
         tileSource = self._loadTileSource(itemId, params)
         try:
--- a/server/rest.py
+++ b/server/rest.py
@@ -24,7 +24,7 @@ from girder.api import access
 from girder.api.v1.item import Item
 from girder.api.describe import describeRoute, Description
 from girder.api.rest import filtermodel, loadmodel, RestException
-from girder.models.model_base import AccessType
+from girder.models.model_base import AccessType, ValidationException
 from girder.plugins.romanesco import utils as romanescoUtils
 
 from .tilesource import TestTileSource, TiffGirderTileSource, \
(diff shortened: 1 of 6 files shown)


server/tilesource/cache.py
```diff
@@ -87,7 +87,8 @@ def __call__(cls, *args, **kwargs):
         instance = cache.get(key, _MARKER)
         if instance is _MARKER:
             instance = super(LruCacheMetaclass, cls).__call__(*args, **kwargs)
-            cache.put(key, instance)
+            if key is not None:
```
brianhelba: Why do we need this? In theory, ``keyFunc`` could return ``None`` as its key value (which is why we instead use ``_MARKER`` to represent the absence of a key).

server/rest.py
```diff
@@ -119,12 +121,10 @@ def createTiles(self, item, params):
                 # a file ID.
                 raise RestException('"fileId" must be a file on the same item'
                                     'as "itemId".')
-
-            if largeImageFile['mimeType'] == 'image/tiff':
-                # TODO: we should ensure that it is a multiresolution tiled TIFF
+            try:
+                TiffGirderTileSource(None, largeImageFile['_id'])
```
brianhelba: Why not just do:
```
try:
    item['largeImage'] = largeImageFile['_id']
except TileSourceException:
    del item['largeImage']
    ... # create job
```
That way, we can avoid the whole extra parameter. It's not a big deal if this gets added to the cache, as it will never be used again and eventually flushed.

server/tilesource/cache.py
brianhelba: I see now why you have this, but I think there's a cleaner way to support the functionality. See my comments above.

server/tilesource/tiff_reader.py
```diff
@@ -139,7 +139,7 @@ def _validate(self):
 
         :raises: ValidationTiffException
         """
-        if self._tiffFile.GetField('SamplesPerPixel') != 3:
+        if self._tiffFile.GetField('SamplesPerPixel') < 3:
```
brianhelba: Do we have an RGBA JPEG image to test with? I want to make sure that viewers handle them sanely before allowing the server to send one.

server/rest.py
```diff
@@ -241,18 +241,25 @@ def deleteTiles(self, item, params):
                 }
 
             # If this file was created by the worker job, delete it
-            if 'largeImageJobId':
+            if 'largeImageJobId' in item:
```
brianhelba: Thanks for fixing this whole block. I threw this in at the last moment, and I don't know what I was thinking...this didn't even make sense before.

server/tilesource/tiff_reader.py
manthey: The python-tests branch uses a png which results in 4 channel image.  When converted by vips, the TIF still reports four channels, but the JPEGs served for the tiles are all 3 channel.  This is the use case I really want to cover.

JPEGs can by more than 3 channel without the fourth channel being alpha (CMYK is common).  I've also seen (years ago) hyper-spectral JPEGs where 3 channels are YCbCr and additional channels are each luminance in different spectral bands.

server/rest.py
manthey: Much nicer than what I had done.

manthey:
ping @brianhelba

server/tilesource/tiff.py
```diff
@@ -38,23 +38,28 @@ class TiffGirderTileSource(GirderTileSource):
     @staticmethod
     def cacheKeyFunc(args, kwargs):
         item = args[0]
-        return item.get('largeImage')
+        return item.get('largeImage') if item else None
```
brianhelba: I'm not sure why you have this. If ``args[0]`` (``self``) isn't found, we've probably done something wrong, so it's fine to raise an exception. Additionally, ``None`` is a perfectly valid cache key, so there could easily be an invisible key collision.

manthey: I should have removed this.  It was added when I was testing if a file had a large image before adding 'largeImage' to the item (since refactored).

brianhelba:
LGTM. Please merge when tests pass.

PR #20, opened 2016-02-25
Python tests

manthey:
Install a later version of libtiff.  This requires using the travis trusty distribution container, which, in turn, requires the sudo version of the container, which is scheduled less frequently and much slower.

We start Romanesco so that we can test conversion jobs.

Fetch external data files from Midas for a real-world image sample.

This also has a tiny monochrome jpeg (which can't be converted), and large 3-channel and 4-channel pngs, both of which can be converted via jobs.

Properly throw errors on test tiles outside of the specified size.  Fix a typo in an exception.


manthey:
@brianhelba This should be ready, but it takes a long time to run the tests and MIDAS sometimes fails to respond in a timely fashion.

brianhelba:
LGTM. We will change the sample TIFF image soon though.

PR #22, opened 2016-03-02
Switch test images.

manthey:
The new test image is sparsely populated (not all tiles exist at higher levels), so the test was modified to accommodate that condition.


brianhelba:
Great, thanks. However, it now looks like even that image may not be appropriate for long-term release, and the attempt to scan a fresh test image this morning resulted in something that's 86MB, which is still too large for my liking. I'll let you know when we finally decide on something.

brianhelba:
LGTM for now. I reserve the right to be obnoxious and change it again at some point.

Issue #23, opened 2016-03-02
Test and upgrade to Numpy 1.10.4


brianhelba:
Numpy 1.10.4 is now released on Conda. If it works, we should upgrade (``requirements.txt`` and Travis build) to use it.

I have noticed some weird behavior with ``pylibtiff`` on a machine where Numpy 1.10.4 was installed, but I haven't been able to definitively blame it on Numpy 1.10.4 yet. If it turns out there is a problem, we need to affirmatively blacklist this version of Numpy and document this issue.

brianhelba:
@manthey Can you take a look at this? It should be fast, and I'm trying to ensure that this and https://github.com/DigitalSlideArchive/HistomicsTK are using the same Numpy.

manthey:
I've been using numpy 1.10.4 on my machine all along, as I had a single instance of Girder with several plugins I was working on and one of them required it.  What sort of weird behavior were you having?

brianhelba:
I was getting segfaults somewhere inside the initial ``import libtiff``. It's totally possible that this is unrelated to Numpy, but I had started to get the error after upgrading Numpy.

If 1.10.4 works for you, just go ahead and formally change ``requirements.txt`` and Travis.

manthey:
Done in PR #24.

PR #24, opened 2016-03-02
Change numpy version to 1.10.4.

manthey:
Resolves issue #23.


brianhelba:
I notice we're not pre-installing Numpy anymore in the Travis build. This was added because the ``setup.py`` for Pylibtiff does not list Numpy as a requirement (even though it is). Is there a reason why this is working on Travis now?

manthey:
Yes.  We list libtiff and numpy in the requirements.txt file.  This works in Ubuntu 14.04 (which we had to switch to in order to support libtiff5).

brianhelba:
LGTM.

PR #26, opened 2016-03-03
Add a bower postinstall step to install bower deps

jbeezley:
@brianhelba I think this is what you wanted to do.  It adds `bower install` as a post install step just like `geojs` does.  The `bower_components` will be set up automagically after running `grunt init`.  I added `angular` as a dependency just to show off how it works; you can add whatever dependencies you need in the `bower.json` file.

### DIFF
--- a/bower.json
+++ b/bower.json
@@ -12,11 +12,7 @@
     "url": "https://github.com/DigitalSlideArchive/large_image.git"
   },
   "dependencies": {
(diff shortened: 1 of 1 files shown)


brianhelba:
@jbeezley Thanks for the sample. We don't need this for ``large_image``, but I will reuse it for my other plugin.

PR #27, opened 2016-03-03
Raise a distinct exception for non-filesystem assetstores.

manthey:


brianhelba:
LGTM once tests pass.

PR #32, opened 2016-03-12
Abstract complex logic into a new Model

brianhelba:
This should allow all important operations to be called internally from Python, as well as via the HTTP API.

### DIFF
--- a/server/models/image_item.py
+++ b/server/models/image_item.py
@@ -1,0 +1,242 @@
+#!/usr/bin/env python
+# -*- coding: utf-8 -*-
+
+###############################################################################
+#  Copyright Kitware Inc.
+#
+#  Licensed under the Apache License, Version 2.0 ( the "License" );
+#  you may not use this file except in compliance with the License.
+#  You may obtain a copy of the License at
+#
+#    http://www.apache.org/licenses/LICENSE-2.0
+#
+#  Unless required by applicable law or agreed to in writing, software
+#  distributed under the License is distributed on an "AS IS" BASIS,
+#  WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+#  See the License for the specific language governing permissions and
+#  limitations under the License.
+###############################################################################
+
+import collections
+import os
+
+from girder.api.rest import RestException
+from girder.models.model_base import ValidationException
+from girder.models.item import Item
+from girder.plugins.worker import utils as workerUtils
+from girder.plugins.jobs.constants import JobStatus
+
+from ..tilesource import TestTileSource, TiffGirderTileSource, \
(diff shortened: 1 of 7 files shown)


brianhelba:
@manthey Can you take a look at this? I'm having trouble setting up the tests locally, so I'm not sure exactly why they're failing.

server/rest.py
```diff
@@ -116,196 +55,58 @@ def createTiles(self, item, params):
...
+    @access.public
+    def getTilesInfo(self, itemId, params):
+        if itemId == 'test':
+            item = 'test'
+            imageArgs = self._parseTestParams(params)
+        else:
+            item = self.model('item').load(
+                id=itemId, level=AccessType.READ,
+                user=self.getCurrentUser(), exc=True)
+            imageArgs = dict()
```
manthey: This needs to be the params dict.  We support additional parameters for both getTilesInfo and getTile, such as asking for encoding='PNG' (the SVS tiles specifically support encoding, jpegSubsampling and jpegQuality as additional options).

server/rest.py
```diff
@@ -330,11 +131,30 @@ def getTile(self, itemId, z, x, y, params):
...
-        cherrypy.response.headers['Content-Type'] = tileSource.getTileMimeType()
+        if itemId == 'test':
+            item = 'test'
+            imageArgs = self._parseTestParams(params)
+        else:
+            # TODO: cache the user / item loading too
+            item = self.model('item').load(
+                id=itemId, level=AccessType.READ,
+                user=self.getCurrentUser(), exc=True)
+            imageArgs = dict()
```
manthey: This needs to be the params dict.  See above.

server/models/image_item.py
```diff
@@ -0,0 +1,242 @@
...
+            job = self._createLargeImageJob(item, fileObj, user, token)
+            item['expectedLargeImage'] = True
+            item['largeImageOriginalId'] = fileObj['_id']
+            item['largeImageJobId'] = job['_id']
+
+        self.save(item)
+        return job
+
+    def _createLargeImageJob(self, item, fileObj, user, token):
+        path = os.path.join(os.path.dirname(__file__), 'create_tiff.py')
```
manthey: This path is no longer correct.  create_tiff.py is in the parent directory, so you either need to move it or change this to `path = os.path.join(os.path.dirname(__file__), '..', 'create_tiff.py')`

brianhelba:
@manthey PTAL.

manthey:
@brianhelba I changed what exceptions are thrown, since we don't want Rest exceptions in the general model.  I'll merge this in; if you want changes, we'll do it as another PR.

brianhelba:
This is perfect, thanks @manthey.

PR #34, opened 2016-03-17
Add GET image/{id}/tiles/thumbnail endpoint.

manthey:
@brianhelba This is done except for tests.  I've pushed it now in case you need it right away.  I will have tests soon.

### DIFF
--- a/server/rest.py
+++ b/server/rest.py
@@ -105,7 +105,6 @@ class TilesItemResource(Item):
         .errorResponse('ID was invalid.')
         .errorResponse('Read access was denied for the item.', 403)
     )
-    @access.cookie
     @access.public
     def getTilesInfo(self, itemId, params):
         if itemId == 'test':
@@ -194,6 +193,7 @@ class TilesItemResource(Item):
         .param('encoding', 'Thumbnail output encoding', required=False,
                enum=['JPEG', 'PNG'], default='JPEG')
     )
+    @access.cookie
     @access.public
     @loadmodel(model='item', map={'itemId': 'item'}, level=AccessType.READ)
     def getTilesThumbnail(self, item, params):
--- a/server/rest.py
+++ b/server/rest.py
@@ -39,6 +39,8 @@ class TilesItemResource(Item):
         apiRoot.item.route('POST', (':itemId', 'tiles'), self.createTiles)
         apiRoot.item.route('GET', (':itemId', 'tiles'), self.getTilesInfo)
         apiRoot.item.route('DELETE', (':itemId', 'tiles'), self.deleteTiles)
+        apiRoot.item.route('GET', (':itemId', 'tiles', 'thumbnail'),
+                           self.getTilesThumbnail)
         apiRoot.item.route('GET', (':itemId', 'tiles', 'zxy', ':z', ':x', ':y'),
                            self.getTile)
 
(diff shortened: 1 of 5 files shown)


server/tilesource/base.py
```diff
@@ -55,6 +69,55 @@ def getTile(self, x, y, z):
     def getTileMimeType(self):
         return 'image/jpeg'
 
+    def getThumbnail(self, width=None, height=None, **kwargs):
```
brianhelba: Can we document the available ``kwargs`` here? We don't even need detail, but a comprehensive list is important.

server/tilesource/base.py
```diff
@@ -15,14 +15,28 @@
...
+
+from girder import logger
 from girder.models.model_base import ValidationException
 from girder.utility import assetstore_utilities
 from girder.utility.model_importer import ModelImporter
+from six import BytesIO
 
 from ..models.base import TileGeneralException
 
+# Not having PIL disables thumbnail creation, but isn't fatal
```
brianhelba: This is fine for now, but we should start thinking about how to handle graceful degradation in a more organized way.

server/rest.py
```diff
@@ -167,3 +176,42 @@ def deleteTiles(self, item, params):
...
+               'image, which means that asking for a large thumbnail will not '
+               'be a high-quality image.')
+        .param('itemId', 'The ID of the item.', paramType='path')
+        .param('width', 'The maximum width of the thumbnail in pixels.',
+               required=False, dataType='int')
+        .param('height', 'The maximum height of the thumbnail in pixels.',
+               required=False, dataType='int')
+        .param('encoding', 'Thumbnail output encoding', required=False,
+               enum=['PNG', 'JPEG'], default='PNG')
+        .param('jpegQuality', 'Quality used for generating JPEG images',
```
brianhelba: Let's drop ``jpegQuality`` and ``jpegSubsampling`` from the thumbnail HTTP API for now. Leave them available internally in the model / tilesource.

We will definitely want to expose these in the HTTP API for a more generalized "region cutout" endpoint in the future.

server/rest.py
```diff
@@ -97,6 +105,7 @@ def _parseTestParams(params):
         .errorResponse('ID was invalid.')
         .errorResponse('Read access was denied for the item.', 403)
     )
+    @access.cookie
```
brianhelba: Why do we need cookie access here? The metadata endpoint should normally only be called via Ajax, which will be able to set headers.

The image-returning endpoints might be fetched by the DOM (e.g. in ``<img src=``), so they must allow cookie auth.

brianhelba:
See my inline comments.

Otherwise, this looks good and is quite a thoughtful implementation. Thanks @manthey .

manthey:
@brianhelba I addressed all of your comments and added testing.

I removed cookie access.  If we change our minds (because we want to fetch the thumbnails via `` tags), we can always put it back.

dgutman:
Oh we will almost definitely be injecting at least thumbnails using the
 tag for image galleries.,

brianhelba:
@manthey Sorry, I was slightly ambiguous in my last comment about cookies. I was asking why you added the ``@access.cookie`` decorator to the ``getTilesInfo`` endpoint, which just returns JSON metadata. I put an inline comment right next to the change, but my pronouns were unclear.

As @dgutman said, ``getTilesThumbnail`` definitely needs cookie access.

brianhelba:
LGTM. Thanks @manthey .

PR #36, opened 2016-03-21
Make the 'largeImage' field into a dict, holding all related values

brianhelba:
* Make the 'largeImage' field into a dict, holding all related values
  * ``item['largeImage']`` now potentially contains the fields:
    * ``'fileId'``
    * ``'sourceName'``
    * ``'jobId'``
    * ``'expected'``
* Remove the entire 'largeImage' dict field when a large image is deleted


brianhelba:
This is a breaking change, but I think it's necessary to keep the item document clean.

I personally don't have any existing deployments that require migration at the moment.

brianhelba:
PTAL @manthey

manthey:
LGTM.  A good improvement.

PR #37, opened 2016-03-21
Only allow one job per item; make fileId optional.

manthey:
If we have scheduled a conversion job, don't schedule another one.

Also, make fileId optional in POST image/{id}/tiles when there is exactly one file.


brianhelba:
LGTM.

I've rebased it on master. Let's merge when the tests pass.

PR #38, opened 2016-03-21
Refactor how the "test" tile source is handled

brianhelba:
This allows us to use the ``@loadmodel`` decorator in several places, and removes some special-case code.

It does remove the ability to permanently mark a specific item as a "test" largeImage. This functionality could be re-added if it turns out to be useful.

### DIFF
--- a/plugin_tests/tiles_test.py
+++ b/plugin_tests/tiles_test.py
@@ -120,12 +120,6 @@ class LargeImageTilesTest(base.TestCase):
                       message.
         :returns: the tile information dictionary.
         """
-        resp = self.request(path='/item/%s/tiles' % itemId, method='DELETE',
-                            user=self.admin)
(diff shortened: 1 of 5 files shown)


brianhelba:
I'm not sure if I like the 2nd commit. It reduces code duplication, but is a bit magical in the way that it bypasses the ``@loadmodel`` decorator.

Thoughts?

brianhelba:
@manthey PTAL.

manthey:
@brianhelba I don't really care for the `__wrapped__` calls that avoid `@loadmodel`.  It works just fine, but I think it is harder to maintain.  I'll push a change that moves the functionality to a different method, then both end points can call that.

Otherwise, LGTM.

brianhelba:
LGTM.

PR #42, opened 2016-03-25
Region endpoint.

manthey:
Regions can optionally be scaled as part of the output.

This even works on sparse PTIF files at all resolutions.

There is currently no explicit guard on over-using memory.  We allocate memory for the PIL image in Python as a single block, which will gracefully fail if we ask for more memory than is in the system.  However, there are cases where python will fail to reuse memory after it has been used.  The memory is freed, but this still occurs.

We might want to use something like psutil to query how much memory the system has and not allow regions that use more than some factor of that (say 1/8th).  We could implement a lock so that a maximum of such a quantity of regions can be generated at a time, Otherwise, concurrent requests could OOM-kill the system.

### DIFF
--- a/server/rest.py
+++ b/server/rest.py
@@ -41,6 +41,8 @@ class TilesItemResource(Item):
         apiRoot.item.route('DELETE', (':itemId', 'tiles'), self.deleteTiles)
         apiRoot.item.route('GET', (':itemId', 'tiles', 'thumbnail'),
                            self.getTilesThumbnail)
+        apiRoot.item.route('GET', (':itemId', 'tiles', 'region'),
+                           self.getTilesRegion)
         apiRoot.item.route('GET', (':itemId', 'tiles', 'zxy', ':z', ':x', ':y'),
                            self.getTile)
 
@@ -198,6 +200,8 @@ class TilesItemResource(Item):
                required=False, dataType='int')
         .param('encoding', 'Thumbnail output encoding', required=False,
                enum=['JPEG', 'PNG'], default='JPEG')
+        .errorResponse('ID was invalid.')
+        .errorResponse('Read access was denied for the item.', 403)
     )
     @access.cookie
     @access.public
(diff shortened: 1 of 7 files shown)


brianhelba:
Is this ready for review?

manthey:
@brianhelba Yes.  The only caveat is that there is no guard against exhausting memory.

server/rest.py
```diff
@@ -252,3 +256,79 @@ def getTilesThumbnail(self, item, params):
...
+        .param('bottom', 'The bottom row (0-based from the top) of the region '
+               'to return.  The region will not include this row.  Negative '
+               'values are offsets from the bottom edge.',
+               required=False, dataType='float')
+        .param('regionWidth', 'The width of the region to return.',
+               required=False, dataType='float')
+        .param('regionHeight', 'The height of the region to return.',
+               required=False, dataType='float')
+        .param('units', 'Units used for left, top, right, bottom, '
+               'sourceWidth, and sourceHeight.  Note that output width and '
```
brianhelba: This mentions ``sourceWidth, and sourceHeight``, but those are not parameters.

server/tilesource/base.py
```diff
@@ -54,6 +59,104 @@ def __init__(self):
...
+        :param sourceHeight: the height of the source data.
+        :returns: the width and height that is no larger than that specified
+                  and preserves aspect ratio.
+        """
+        if sourceWidth == 0 or sourceHeight == 0:
+            return 0, 0
+        # Constrain the maximum size if both width and height weren't
+        # specified, in case the image is very short or very narrow.
+        if height and not width:
+            width = height * 16
```
brianhelba: I know this is existing code, but what is the worst that happens if we didn't effectively bound ``width <= height * 16``? If the image is naturally very short or narrow, shouldn't we just return it as-is?

server/tilesource/base.py
```diff
@@ -54,6 +59,104 @@ def __init__(self):
...
+            output = BytesIO()
+            image.save(output, encoding, quality=jpegQuality,
+                       subsampling=jpegSubsampling)
+            tileData = output.getvalue()
+        return tileData, self.outputMimeTypes[encoding]
+
+    def _getRegionBounds(self, metadata, **kwargs):
+        """
+        Given a set of arguments that can include left, right, top, bottom,
+        sourceWidth, sourceHeight, and units, generate actual pixel values for
```
brianhelba: Again, ``sourceWidth, sourceHeight``.

server/tilesource/base.py
```diff
@@ -54,6 +59,104 @@ def __init__(self):
...
+                if key in kwargs:
+                    kwargs[key] = kwargs[key] * metadata['sizeY']
+        # convert negative references to right or bottom offsets
+        for key in ('left', 'right', 'top', 'bottom'):
+            if key in kwargs and kwargs.get(key) < 0:
+                kwargs[key] += metadata[
+                    'sizeX' if key in ('left', 'right') else 'sizeY']
+        # Calculate the region we need to fetch
+        left = kwargs.get(
+            'left', kwargs.get('right') - kwargs.get('regionWidth')
```
brianhelba: I have no problem with this logic being inlined, but it's very difficult to mentally parse as it's currently formatted. Can you add some additional whitespace (dropping the 2nd argument down to its own line) and some parentheses (making it obvious that the ternary operator applies to the entire subtraction statement, not just the 2nd term).

server/tilesource/base.py
```diff
@@ -54,6 +59,104 @@ def __init__(self):
         self.sizeX = None
         self.sizeY = None
 
+    def _calculateWidthHeight(self, width, height, sourceWidth, sourceHeight):
```
brianhelba: It actually is called ``sourceWidth, sourceHeight`` here, which is internally consistent, but doesn't match the other functions.

server/tilesource/base.py
```diff
@@ -54,6 +59,104 @@ def __init__(self):
...
+        :param encoding: a valid PIL encoding (typically 'PNG' or 'JPEG').
+                         Must also be in the outputMimeTypes map.
+        :param jpegQuality: the quality to use when encoding a JPEG.
+        :param jpegSubsampling: the subsampling level to use when encoding a
+                                JPEG.
+        """
+        if encoding not in self.outputMimeTypes:
+            raise ValueError('Invalid encoding "%s"' % encoding)
+        if image.width == 0 or image.height == 0:
+            tileData = b''
```
brianhelba: Is a 0x0 image really an empty string in both PNG and JPEG? There's no magic number or headers?

server/rest.py
manthey: I'll change it so that it is regionWidth and regionHeight everywhere.

server/tilesource/base.py
manthey: This is only for when the user hasn't specified both dimensions for limitation.  I don't like the idea of unbounded results when the user may not realize that is what they are asking for.  For instance, supposed you have an original image that is extremely short and wide. say 1x16000, and the user asks for a height of 1000 pixels, it seems imprudent to try to generate that.  If they really want it, they can ask for (a very large number)x1000.

server/tilesource/base.py
manthey: Sure.

server/tilesource/base.py
manthey: Neither PNG nor JPEG support 0 values for width or height.  We could raise an error instead.

server/tilesource/base.py
```diff
@@ -107,28 +205,124 @@ def getThumbnail(self, width=None, height=None, **kwargs):
...
+        #   image = PIL.Image.new('RGB', (regionWidth, regionHeight))
+        # but, for large images (larger than 4 Megapixels), PIL allocates one
+        # memory region per line.  Although it frees this, the memory manager
+        # often fails to reuse these smallish pieces.  By allocating the data
+        # memory ourselves in one block, the memory manager does a better job.
+        # Furthermode, if the source buffer isn't in RGBA format, the memory is
+        # still often inaccessible.
+        image = PIL.Image.frombuffer(
+            mode, (regionWidth, regionHeight),
+            '\x00' * (regionWidth * regionHeight * 4), 'raw', 'RGBA', 0, 1)
```
brianhelba: This should be a ``bytes``.

brianhelba: Can we add a code comment that PIL requires

> 'RGBA', 0, 1)

for ``raw`` images?

tymiao1220:
Hi @manthey, instead of appending '-region' in the download view name, will that makes sense to append coordinate because that makes easier to the user to remember and find from WSI.

tymiao1220:
Good to know, thanks!

Issue #39, opened 2016-03-23
Add client settings


manthey:
Add plugin settings that allow turning on or off showing the client viewer.

Also allow setting the default viewer.

manthey:
Added in PR #43.

Issue #41, opened 2016-03-23
Add button to mark item as large image.


manthey:

brianhelba:
I'm thinking this could just be a button added on to each [``.g-file-list-entry``](https://github.com/girder/girder/blob/master/clients/web/src/templates/widgets/fileList.jade) to "mark this as a large image", which will kick off the POST request. Since MIME types are a bit unreliable in Girder at the moment, I'm thinking it should be available for every listed file.

Once the large image exists, the button should change to only be present on the active large image, and allow a DELETE request to be sent (as well as to signify which image is marked).

It would also be nice to have a special signifier for when an image creation task is pending (i.e. ``item.largeImage.expected == true``)/

manthey:
Should there be an option to show this or not (like controlling whether we show thumbnails and the viewer via settings)?

brianhelba:
IMO, no. These should be tiny, non-intrusive icons that directly map to the REST endpoints that the plugin provides.

manthey:
Added in PR #43.

Issue #40, opened 2016-03-23
Add thumbnails.


manthey:
Add thumbnails to the item list.  Make this an option via the settings (depends on issue #39).

manthey:
Added in PR #43.

PR #43, opened 2016-03-28
 Settings and thumbnails list

manthey:
This currently has the region endpoints PR in the commit history; it can be separated from that, though it would be easier if that PR is resolved first instead.

This add a plugin settings page that allows turning on and off the image viewer on item pages, thumbnails in item lists, and setting the default viewer.  It adds thumbnails on item lists.  It adds additional actions on file lists: if a large image file is set, it can be removed.  If not set, it can be added to any file in the item.  If expected, the large image action icon is shown with some transparency to indicate that it is in progress.

There aren't any client tests yet.

Resolves issues #39 (client settings), #40 (thumbnails), and #41 (button to mark large image).

### DIFF
--- a/server/__init__.py
+++ b/server/__init__.py
@@ -55,6 +55,8 @@ def validateSettings(event):
 
     if key in (constants.PluginSettings.LARGE_IMAGE_SHOW_THUMBNAILS,
                constants.PluginSettings.LARGE_IMAGE_SHOW_VIEWER):
+        if str(val).lower() not in ('false', 'true', ''):
+            return
         val = (str(val).lower() != 'false')
     elif key == constants.PluginSettings.LARGE_IMAGE_DEFAULT_VIEWER:
         val = str(val).strip()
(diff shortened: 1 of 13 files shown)


manthey:
@brianhelba I still need to write tests for this, but the functionality is complete.  See the PR description.

brianhelba:
I just merged the ``test`` endpoint refactor. Please rebase this when you get a chance.

server/__init__.py
```diff
@@ -49,6 +50,20 @@ def _postUpload(event):
         Item.save(item)
 
 
+def validateSettings(event):
+    key, val = event.info['key'], event.info['value']
+
+    if key in (constants.PluginSettings.LARGE_IMAGE_SHOW_THUMBNAILS,
+               constants.PluginSettings.LARGE_IMAGE_SHOW_VIEWER):
+        val = (str(val).lower() != 'false')
```
brianhelba: Let's only allow ``'true'``, ``'false'`` and ``''`` as valid.

brianhelba:
LGTM when tests pass and setting validation is stricter.

PR #45, opened 2016-04-04
Allow public access of large image settings.

manthey:
Otherwise, unless you are logged in as an admin, the client can't read them.

I think Girder ultimately needs access control on individual settings, but that is a bigger issue.


brianhelba:
LGTM.

PR #46, opened 2016-04-05
Better handling of exceptions when loading tile soruces

manthey:
When we installed the large_image plugin in an environment that had openslide-python but not the libtiff module, the DummyTileSource replaced the tiff tile source (as intended), but then this "processed" SVS files and prevented them from working.  This change will allow the file to be properly handed off to the SVS tile source.

Improve the error messages for import failures.

Added some tests for import failures.

### DIFF
--- a/server/tilesource/__init__.py
+++ b/server/tilesource/__init__.py
@@ -20,21 +20,42 @@
 # Deal with a bug where PEP257 crashes when parsing __all__
 # flake8: noqa
 
+import collections
+from girder.constants import TerminalColor
+from girder import logger
+
 from .base import TileSource, TileSourceException, \
     TileSourceAssetstoreException
 
-try:
-    from .test import TestTileSource
-except ImportError:
-    from .dummy import DummyTileSource as TestTileSource
-try:
-    from .tiff import TiffGirderTileSource
-except ImportError:
-    from .dummy import DummyTileSource as TiffGirderTileSource
-try:
-    from .svs import SVSGirderTileSource
-except ImportError:
-    from .dummy import DummyTileSource as SVSGirderTileSource
+AvailableTileSources = collections.OrderedDict()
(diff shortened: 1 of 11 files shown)


brianhelba:
As I discussed with @manthey , we can fix this in a more general way.

server/tilesource/__init__.py
```diff
@@ -20,21 +20,42 @@
...
-try:
-    from .svs import SVSGirderTileSource
-except ImportError:
-    from .dummy import DummyTileSource as SVSGirderTileSource
+AvailableTileSources = collections.OrderedDict()
+all = [TileSource, TileSourceException, TileSourceAssetstoreException,
+       AvailableTileSources]
+
+sourceList = [
+    {'moduleName': '.tiff', 'className': 'TiffGirderTileSource'},
```
brianhelba: Eventually, I'd like to exclusively use MIME types rather than file extensions, but we need to make several other steps (improve Girder's handling of MIME, reach a consensus on what the MIME of an SVS should be) before this can effectively happen.

server/tilesource/__init__.py
```diff
@@ -20,21 +20,42 @@
 # Deal with a bug where PEP257 crashes when parsing __all__
 # flake8: noqa
```
brianhelba: I wonder if PEP257 will still crash now that we build `__all__` in this way. I'll file a bug to explore it.

brianhelba:
LGTM. Great work, I expect that this can be the final mechanism we use for graceful degradation for the foreseeable future.

server/tilesource/__init__.py
manthey: This list is our python module name and has nothing to do with extensions.

I agree, relying on file extensions is a bad idea.  We currently work our way through the readers, asking each in turn if they can handle the file.  We could use mime type (or even file extension) to change the order of the trials, but I think the current method is sound enough.

PR #50, opened 2016-04-08
Annotations RESTful interface

manthey:


manthey:
@brianhelba I've added IDs to the shapes and added work-arounds for the lack of transactions in Mongo.  I think this is done, though there are no tests.

I wonder if we should allow element IDs to be arbitrary strings.  Currently, they need to be of the same form as Mongo IDs (24 hex digits), and are assigned from the Mongo IDs if they are not specified.  I could see a client finding it useful to use their own ids, in which case we could make this an arbitrary string (it is still required to be unique), or a string with strong limitations.

server/models/annotation.py
```diff
@@ -0,0 +1,534 @@
...
+                        'enum': ['rectangle']
+                    },
+                },
+                'required': ['type'],
+                'patternProperties': baseRectangleShapePatternProperties,
+                'additionalProperties': False
+            }
+        ]
+    }
+    rectangleGridShapeSchema = {
```
cdeepakroy: What is the difference between rectangleGridShapeSchema and rectangleShapeSchema?

server/models/annotation.py
```diff
@@ -0,0 +1,534 @@
...
+                        # If we include the baseShapeSchema, then shapes that
+                        # are as-yet invented can be included.
+                        # baseShapeSchema,
+                        arrowShapeSchema,
+                        circleShapeSchema,
+                        pointShapeSchema,
+                        polylineShapeSchema,
+                        rectangleShapeSchema,
+                        rectangleGridShapeSchema,
+                    ]
```
cdeepakroy: I think we would also need an (orientable) ellipse.

server/models/annotation.py
manthey: Charles has found it useful in SlideAtlas to have a grid (a regular array of rectangles) at a specific scale and orientation to add in scale analysis of slides.  Rather than just be the frame of a rectangle (as in rectangleShapeSchema), the grid adds `widthSubdivisions` and `heightSubdivisions`, and, as its name suggests, renders as a grid.

server/models/annotation.py
manthey: No problem.  We can add any additional shapes as needed.  Would it be the same form as the rectangle?  That is, we give a center, width, height, rotation, and, optionally for 3-D scenarios, a normal?  Or would it be more useful to have it be specified via foci?

cooperlab: I would stick with center, rotation, and some elliptical scale [0-1]. Math an be tricky with foci.


Sent from my iPhone

On Apr 12, 2016, at 5:40 PM, David Manthey <notifications@github.com<mailto:notifications@github.com>> wrote:


In server/models/annotation.py<https://github.com/DigitalSlideArchive/large_image/pull/50#discussion_r59398991>:

> +            'elements': {
> +                'type': 'array',
> +                'items': {
> +                    # Shape subtypes are mutually exclusive, so for
> +                    #  efficiency. don't use 'oneOf'
> +                    'anyOf': [
> +                        # If we include the baseShapeSchema, then shapes that
> +                        # are as-yet invented can be included.
> +                        # baseShapeSchema,
> +                        arrowShapeSchema,
> +                        circleShapeSchema,
> +                        pointShapeSchema,
> +                        polylineShapeSchema,
> +                        rectangleShapeSchema,
> +                        rectangleGridShapeSchema,
> +                    ]


No problem. We can add any additional shapes as needed. Would it be the same form as the rectangle? That is, we give a center, width, height, rotation, and, optionally for 3-D scenarios, a normal? Or would it be more useful to have it be specified via foci?

-
You are receiving this because you are subscribed to this thread.
Reply to this email directly or view it on GitHub<https://github.com/DigitalSlideArchive/large_image/pull/50/files/149c5ffc59cebe59cea00b1f051dbecc2db4c549#r59398991>

________________________________

This e-mail message (including any attachments) is for the sole use of
the intended recipient(s) and may contain confidential and privileged
information. If the reader of this message is not the intended
recipient, you are hereby notified that any dissemination, distribution
or copying of this message (including any attachments) is strictly
prohibited.

If you have received this message in error, please contact
the sender by reply e-mail message and destroy all copies of the
original message (including attachments).

server/models/annotation.py
```diff
@@ -0,0 +1,534 @@
...
+            'lineWidth': {
+                'type': 'number',
+                'minimum': 0
+            }
+        },
+        'required': ['type'],
+        'additionalProperties': True
+    }
+    baseShapePatternProperties = {
+        '^%s$' % properyName: {}
```
cdeepakroy: propertyName

server/models/annotation.py
manthey: It is pretty easy to convert between width and height and eccentricity and semi-major axis (or even semi-inter-foci distance and semi major axis).  Regardless, we'll have a rotation.

I think width and height are convenient, in that they make it easy to show the circum-rectangle, and the algorithms I've used in the past to fit ellipses generally output semi-major and semi-minor axes, but if eccentricity is a more useful metric, I'm happy to use it.

cooperlab: Width and height is better - I just think the actual foci and conversions that are less intuitive.

Sent from my iPhone

On Apr 12, 2016, at 5:59 PM, David Manthey <notifications@github.com<mailto:notifications@github.com>> wrote:


In server/models/annotation.py<https://github.com/DigitalSlideArchive/large_image/pull/50#discussion_r59402485>:

> +            'elements': {
> +                'type': 'array',
> +                'items': {
> +                    # Shape subtypes are mutually exclusive, so for
> +                    #  efficiency. don't use 'oneOf'
> +                    'anyOf': [
> +                        # If we include the baseShapeSchema, then shapes that
> +                        # are as-yet invented can be included.
> +                        # baseShapeSchema,
> +                        arrowShapeSchema,
> +                        circleShapeSchema,
> +                        pointShapeSchema,
> +                        polylineShapeSchema,
> +                        rectangleShapeSchema,
> +                        rectangleGridShapeSchema,
> +                    ]


It is pretty easy to convert between width and height and eccentricity and semi-major axis (or even semi-inter-foci distance and semi major axis). Regardless, we'll have a rotation.

I think width and height are convenient, in that they make it easy to show the circum-rectangle, and the algorithms I've used in the past to fit ellipses generally output semi-major and semi-minor axes, but if eccentricity is a more useful metric, I'm happy to use it.

-
You are receiving this because you commented.
Reply to this email directly or view it on GitHub<https://github.com/DigitalSlideArchive/large_image/pull/50/files/149c5ffc59cebe59cea00b1f051dbecc2db4c549#r59402485>

________________________________

This e-mail message (including any attachments) is for the sole use of
the intended recipient(s) and may contain confidential and privileged
information. If the reader of this message is not the intended
recipient, you are hereby notified that any dissemination, distribution
or copying of this message (including any attachments) is strictly
prohibited.

If you have received this message in error, please contact
the sender by reply e-mail message and destroy all copies of the
original message (including attachments).

server/models/annotationelement.py
```diff
@@ -0,0 +1,116 @@
...
+        """
+        Given an annotation, fetch the elements from the database and add them
+        to it.
+
+        :param annotation: the annotation to get elements for.  Modified.
+        """
+        elementCursor = self.find({'annotationId': annotation['_id'],
+                                   '_version': annotation['_version']},
+                                  sort=[('_id', SortDir.ASCENDING)])
+        annotation['annotation']['elements'] = elements = []
```
cdeepakroy: Use single assignment for better readability?

server/rest/annotation.py
```diff
@@ -0,0 +1,165 @@
...
+        self.route('PUT', (':id',), self.updateAnnotation)
+        self.route('DELETE', (':id',), self.deleteAnnotation)
+
+    @describeRoute(
+        Description('Search for annotations.')
+        .responseClass('Annotation')
+        .param('itemId', 'List all annotations in this item.', required=False)
+        .param('text', 'Pass this to perform a full text search for '
+               'annotation names and descriptions.', required=False)
+        .param('name', 'Pass to llokup an annotation by exact name match.',
```
cdeepakroy: llokup --> lookup

server/models/annotationelement.py
manthey: That was the intent.  When I was developing the method, I thought I would need to refer to `annotation['annotation']['elements']` multiple times, and this would make it easier to follow.  The method ended up being simple enough, it now seems unnecessary.

cdeepakroy:
@manthey LGTM.

Next, can you:
* add tests that show how to interact with the REST API to create, find, or update annotations
* add tests that show how the backend analytics code interact with the Annotation API to create, find, or update annotations. Do we do something like importing the Annotation class and interact with its member functions to achieve that?

server/models/annotation.py
cdeepakroy: Ok!

manthey:
@cdeepakroy I just added an ellipse to the schema.

server/models/annotation.py
cdeepakroy: Yes i would stick with center, width, height, and rotation similar to rectangle as they are intuitive and given that other code will have to provide these things while creating/editing the annotation. 

I was thinking maybe we should use major axis length and minor axis length instead of width and height, respectively but come to think of it leaving them as width (for major axis length) and height (minor axis length) seems more intuitive.

manthey:
Initially, I am expecting that the analytics code will interact with the annotations much like other Girder models.  That is, calling `createAnnotation`, `find`, `load`, and `save` from the annotations model.  Eventually there will be a method to filter the returned elements of the annotation to spatial regions (or perform downsampling).  For now, it is expected that you get and set all of them at once.

server/models/annotation.py
cooperlab: Good thought - major/minor axis is the other common way people would deal with ellipses

________________________________

cdeepakroy:
I think to begin with, the backend analytics code will largely generate annotations. Maybe when we delve into implementing the active learning stuff, we might have to do find/search operations in specific regions/tiles.

cooperlab:
Might want to think about a streamlined way to ingest machine-generated annotations in bulk. Many of these will share significant metadata tags that would be repeated up to 1M times per slide.
________________________________

server/models/annotation.py
```diff
@@ -278,6 +278,23 @@ class AnnotationSchema:
...
+                        'type': 'string',
+                        'enum': ['ellipse']
+                    },
+                },
+                'required': ['type'],
+                'patternProperties': baseRectangleShapePatternProperties,
+                'additionalProperties': False
+            }
+        ]
+    }
```
cdeepakroy: LGTM!

manthey:
@cooperlab As part of writing tests, I'll write some for performance and try out sets up to the millions of elements.  If what we have isn't performant enough, we'll adjust it.

PR #53, opened 2016-04-13
Add some documentation about annotations.

manthey:
Added an endpoint to return the official schema.

### DIFF
--- a/docs/annotations.md
+++ b/docs/annotations.md
@@ -1,0 +1,229 @@
+# Annotation Schema
+
+The full schema can be obtained by calling the Girder endpoint of `GET` `/annotation/schema`.
+
+An annotation consists of a basic structure which includes a free-form `attributes` object and a list of `elements`.  The elements are strictly specified by the schema and are limited to a set of defined shapes.
+
+Partial annotations are shown below with some example values.  Note that the comments are not part of a valid annotation:
+
+```
+{
+  "name": "MyAnnotationName",              # Non-empty string.  Optional
+  "description": "This is a description",  # String.  Optional
+  "attributes": {                          # Object.  Optional
+    "key1": "value1",
+    "key2": ["any", {"value": "can"}, "go", "here"] 
+  },
+  "elements": []                           # A list.  Optional.  
+                                           # See below for valid elements.
+}
+```
+
+## Elements
+
+Currently, all defined elements are shapes.  All of these have some properties that they are allowed.  Each shape is listed below:
+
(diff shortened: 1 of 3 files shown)


docs/annotations.md
```diff
@@ -0,0 +1,229 @@
...
+
+```
+{
+    "type": "point",                  # Exact string for the specific shape.  Required
+    "id": "0123456789abcdef01234567", # String, 24 lowercase hexadecimal digits.  Optional.
+    "label": {                        # Object.  Optional
+      "value": "This is a label",     # String.  Optional
+      "visability": "hidden",         # String.  One of "always", "hidden", "onhover".  Optional
+      "fontSize": 3.4,                # Number.  Optional
+      "color": "#0000FF"              # String.  See note about colors.  Optional
```
jbeezley: Does this mean the color of the font, the label background, or something else?

docs/annotations.md
```diff
@@ -0,0 +1,229 @@
...
+    "type": "rectanglegrid",           # Exact string.  Required
+    <id, label, lineColor, lineWidth>  # Optional general shape properties
+    "center": [10.3, -40.0, 0],        # Coordinate.  Required
+    "width": 5.3,                      # Number >= 0.  Required
+    "height": 17.3,                    # Number >= 0.  Required
+    "rotation": 0,                     # Number.  Counterclockwise radians around normal.  Required
+    "normal": [0, 0, 1.0],             # Three numbers specifying normal.  Default is positive Z.  Optional
+    "widthSubdivisions": 3,            # Integer > 0.  Required
+    "heightSubdivisions": 4,           # Integer > 0.  Required
+    "fillColor": "rgba(0, 255, 0, 1)"  # String.  See note about colors.  Optional
```
jbeezley: Are grids supposed to have fill colors?

docs/annotations.md
manthey: Font color (though we might have to as @brianhelba to be certain).  I don't think that we currently have a concept of a label bound box and background (or even relative positioning).

docs/annotations.md
manthey: I doubt @law12019 is using a fill color at all in his use of rectanglegrids.  I could see wanting the whole region to have a color with a partial opacity.

jbeezley: Yes, I suppose if the grid can be a subregion of the full image, than it could be useful.

cdeepakroy:
LGTM!

PR #54, opened 2016-04-14
Refactor the tile sources so that they can be used without Girder.

manthey:
Assuming you have the prerequisites, the tilesource can be used without girder.  To facilitate importing it from the source tree, when server is imported without girder, it will still work (though, naturally, you can't import server.models or server.rest).

The plugin_tests/girderless_test.py file shows some access of file sources without girder.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -3,3 +3,7 @@ node_modules
 _build
 .*.swp
 .*.swo
+env
+dist
+build
+large_image.egg-info
--- a/large_image.py
+++ b/large_image.py
@@ -1,0 +1,21 @@
+#!/usr/bin/env python
+# -*- coding: utf-8 -*-
+
+##############################################################################
+#  Copyright Kitware Inc.
+#
+#  Licensed under the Apache License, Version 2.0 ( the "License" );
+#  you may not use this file except in compliance with the License.
+#  You may obtain a copy of the License at
+#
+#    http://www.apache.org/licenses/LICENSE-2.0
+#
+#  Unless required by applicable law or agreed to in writing, software
+#  distributed under the License is distributed on an "AS IS" BASIS,
+#  WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+#  See the License for the specific language governing permissions and
+#  limitations under the License.
+##############################################################################
+
+from server import tilesource
+getTileSource = tilesource.getTileSource  # noqa
--- a/large_image/__init__.py
+++ b/large_image/__init__.py
@@ -17,5 +17,8 @@
 #  limitations under the License.
 ##############################################################################
 
+import server
 from server import tilesource
 getTileSource = tilesource.getTileSource  # noqa
+
+__all__ = [server, tilesource, getTileSource]
--- a/plugin.cmake
+++ b/plugin.cmake
@@ -36,7 +36,16 @@ add_python_test(tiles PLUGIN large_image BIND_SERVER EXTERNAL_DATA
   "sample_image.ptif" "plugins/large_image/sample_image.ptif"
   "sample_svs_image.svs" "plugins/large_image/sample_svs_image.TCGA-DU-6399-01A-01-TS1.e8eb65de-d63e-42db-af6f-14fefbbdf7bd.svs"
   )
-add_python_test(import PLUGIN large_image)
 set_property(TEST server_large_image.tiles APPEND PROPERTY ENVIRONMENT
   "LARGE_IMAGE_DATA=${PROJECT_BINARY_DIR}/data/plugins/large_image")
+
+add_python_test(import PLUGIN large_image)
+
+add_python_test(girderless PLUGIN large_image BIND_SERVER EXTERNAL_DATA
+  "sample_image.ptif" "plugins/large_image/sample_image.ptif"
+  "sample_svs_image.svs" "plugins/large_image/sample_svs_image.TCGA-DU-6399-01A-01-TS1.e8eb65de-d63e-42db-af6f-14fefbbdf7bd.svs"
+  )
+set_property(TEST server_large_image.girderless APPEND PROPERTY ENVIRONMENT
+  "LARGE_IMAGE_DATA=${PROJECT_BINARY_DIR}/data/plugins/large_image")
+
 add_web_client_test(example "${CMAKE_CURRENT_LIST_DIR}/plugin_tests/exampleSpec.js" PLUGIN large_image)
--- a/plugin_tests/girderless_test.py
+++ b/plugin_tests/girderless_test.py
@@ -1,0 +1,147 @@
+#!/usr/bin/env python
+# -*- coding: utf-8 -*-
+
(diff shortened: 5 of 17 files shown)


cdeepakroy:
@manthey This is neat! 

Would it be better to move [TileSource](https://github.com/DigitalSlideArchive/large_image/blob/tilesources-without-girder/server/tilesource/base.py#L56), [FileTileSource](https://github.com/DigitalSlideArchive/large_image/blob/tilesources-without-girder/server/tilesource/base.py#L348), [SvsFileTileSource](https://github.com/DigitalSlideArchive/large_image/blob/tilesources-without-girder/server/tilesource/svs.py#L40), and [TiffTileSource](https://github.com/DigitalSlideArchive/large_image/blob/tilesources-without-girder/server/tilesource/tiff.py#L45) from `large_image/server` to `large_image/large_image`? 

Then eventually when we support `pip install large_image` then people get access only to the stuff in `large_image/large_image` and dont get anything in `large_image/server`. That way when people use it as a pure python package to load local files they dont see any server stuff (from **large_image.server** import tilesource).

Will do a more detailed code review and get back to you with comments

manthey:
Ideally, we'd just move the tilesource module to the top level, but I think that Girder requires that plugins' python code be under the `server` module.  I'm not immediately sure how to move it so that some of the code is outside of that.  It may require a change to girder (or I could just be unaware of how to do it).

I'll look into doing so soon.

zachmullen:
One could certainly import code outside the `server` directory using relative paths, in a number of different ways.

jbeezley:
You can move the loading function to an arbitrary place in the package using a setuptools `entry_point`.  See the [test plugin](https://github.com/girder/girder/tree/master/tests/packaging/entry_point_plugin) for an example.

cdeepakroy:
@ manthey It would be nice to implement something like a [factory pattern](http://www.tutorialspoint.com/design_pattern/factory_pattern.htm) to enable the following:

```python
from large_image.server.tilesource import TileSource

largeImg = TileSource('a.svs')
assert( type(largeImg).__name__ = 'SvsFileTileSource' )

largeImg = TileSource('a.tif')
assert( type(largeImg).__name__ = 'TiffFileTileSource' )

largeImg = TileSource('girder:<item-id-of-svs-image>') 
# If the file-name string stats with `girder:` then it is a girder item
assert( type(largeImg).__name__ == 'GirderSVSTileSource' )

largeImg = TileSource('girder:<item-id-of-tiff-image>')
assert( type(largeImg).__name__ == 'GirderTiffTileSource' )
```
slicer [uses](https://github.com/Slicer/Slicer/blob/master/Libs/MRML/IDImageIO/itkMRMLIDImageIO.cxx#L159) 'slicer:<image-address-in-memory>' to enable itk::ImageFileReader to load an image in memory. 

The benefit of the 'girder:' thingy is that the person writing a CLI will be agnostic to where the CLI is located and in which format it is in.

cdeepakroy:
I guess you will have to implement the static CanRead() member function in each of the concrete classes and have someway to register them into a factory/list of concrete classes to achieve that.

cdeepakroy:
@manthey See [this](https://github.com/InsightSoftwareConsortium/ITK/blob/0d5801bc8776fdf228b906682f0c34c8ddd72889/Modules/IO/ImageBase/include/itkImageFileReader.hxx#L105-L110) and [this](https://github.com/InsightSoftwareConsortium/ITK/blob/0d5801bc8776fdf228b906682f0c34c8ddd72889/Modules/IO/ImageBase/src/itkImageIOFactory.cxx#L45-L62) to learn how itk::ImageFileReader does this.

manthey:
We're doing something similar in the plugin's model -- iterating through the available sources and asking each one if it can read the girder item.  If it can, we use that.  We then store which a reference to the class that was used, so that we don't have to check when we access that item in the future.

Checking which source can be used can be expensive.  Also some files can be read by multiple sources, but one source may be more desirable than another.  We can certainly make a single tile source entry point, where you can optionally specify which class should be used (as is functionally done in the girder model).

I don't really like the prefixing of the filename with 'girder:' to indicate that the name refers to an item, as a file could start with that string.  I suppose one could make it a protocol spec and have it start with 'girder://'.

cdeepakroy:
I see your concern with `girder:`. Prefixing with `girder://` is fine by me. 

Bottomline -- When i write some code that uses the `TileSource` class to load a large image, i wouldnt like to put anything in the code where i have to specify whether it is on girder or not, whether it is .svs or .tiff. Once i write the code it should work irrespective of where the image is located and which format it is in.

When you say `Checking which source can be used can be expensive. Also some files can be read by multiple sources, but one source may be more desirable than another.` can you tell me which specific scenarios are these and what makes it expensive so i could understand. Can we solve this by maintaining a priority/preference internally depending on the file extension?

cdeepakroy:
In anycase, when you are checking for available sources, maybe you should first check for girder sources (if used as girder plugin) and then file sources.

cdeepakroy:
Maybe we can use `<girder>:` and check if girder sources (SvsGirderTileSource, TiffGirderTileSource) can read first and then check if file sources (SvsFileTileSource, TiffFileTileSource)  can read it.

If multiple tile sources can read it, then maybe we can have the classes internally assign a preference value for all the extensions it can read and pick the best source based on the highest preference value.

manthey:
@cdeepakroy I've added a function `getTileSource(<uri or path>[, user=<girder user>], **kwargs)` that should do just what you want.  For files, you can call `getTileSource(<path>)`.  For girder items, you can call `getTileSource('girder_item://<item id>', user=<user>)`.  A user isn't required if the item is publicly readable.  The user can be passed in all cases; it will be ignored for file sources.

I've also added a root-level module `large_image`, so you can
```
import large_image
source = large_image.getTileSource(<path>)
```

The new code is covered by testing.

For now, we have a fixed priority for the tile sources, and they are tried in order (first ptif, then svs) unless they are known to be a specific type.

cdeepakroy:
@manthey Great! This works for now.

I kind of don't like any code to be at the root level but that is something we can collectively discuss.

The way it is now, to use large_image we will have to git clone the large_image repo and add the root directory to the python path. 

Probably the next steps are to create setup.py to facilitate `python setup.py install` of large_image and eventually support `pip install large_image` which should only give me access to non-girder classes/functionality. 

When someone wants to use large_image as a girder plugin they should `git clone` it.

@jbeezley and @zachmullen could you also take a look?

zachmullen:
As far as package creation, we can now install third-party plugins pretty easily using the `girder.plugin` entrypoint that @jbeezley added. @jbeezley do you have any projects that are currently making use of that feature in their setup.py that we could use as an example?

jbeezley:
The only example currently is the test inside girder.  Also important to note, the entry_point method for registering plugins won't work for client side code.

cdeepakroy:
@manthey Great! LGTM!

cdeepakroy:
Fixes DigitalSlideArchive/HistomicsTK#47

PR #56, opened 2016-04-29
Cache sources based on available memory.

manthey:
Cache getTile and getThumbnail as well.  These caches are per instance of a tile source; they really should be a single cache for all.

Similarly, the tile source class cache should be one cache, not one cache per class.

Switched to Pillow 3.1.2.

Added a change to the geojs viewer to (when geojs is updated) not request tiles we know don't exist in certain circumstances.

TODO:
- [x] Testing
- [x] Use single caches rather than per-instance and per-class caches


dgutman:
Dave--

Check this out:
https://pythonhosted.org/Flask-Cache/


I added a simple decorator to some of my own code and I hooked it right up
to memcached, and it supports multipel variants/cacheing strategies..

manthey:
@dgutman There are many lru caching libraries.  The Flask one has a lot more overhead than the repoze.lru one we are using.  As of Python 3.2, there is a standard library cache (functools.lru_cache), and there is a backport for earlier versions of Python.  

Most of these cache program work okay with single functions, but, what we really want is a single cache for all of the tile responses rather than one cache for tiles from each image file.

The reason for this is that we know (roughly) how much memory is available, so we know a safe number of tiles to cache.  But, all of the off-the-shelf cache programs are intended for single instances of functions, not for classes or multiple instances of methods that should share a cache.

dgutman:
I just used the Flask one and pointed it to MEMCACHED--- so all machines
would be hitting the same cache.   Since there's no obvious reason why a
thread is likely to get the request for the same cached tile... a
decentralized (aka Memcached or equiv)  seems to be the better way..



On Fri, Apr 29, 2016 at 8:54 AM David Manthey <notifications@github.com>
wrote:

> @dgutman <https://github.com/dgutman> There are many lru caching
> libraries. The Flask one has a lot more overhead than the repoze.lru one we
> are using. As of Python 3.2, there is a standard library cache
> (functools.lru_cache), and there is a backport for earlier versions of
> Python.
>
> Most of these cache program work okay with single functions, but, what we
> really want is a single cache for all of the tile responses rather than one
> cache for tiles from each image file.
>
> The reason for this is that we know (roughly) how much memory is
> available, so we know a safe number of tiles to cache. But, all of the
> off-the-shelf cache programs are intended for single instances of
> functions, not for classes or multiple instances of methods that should
> share a cache.
>
> —
> You are receiving this because you were mentioned.
>
>
> Reply to this email directly or view it on GitHub
> <https://github.com/DigitalSlideArchive/large_image/pull/56#issuecomment-215703132>
>
-- 
David A Gutman MD PhD
Assistant Professor of Neurology, Psychiatry & Biomedical Informatics
Emory University School of Medicine
Staff Physician, Mental Health Service Line
Atlanta VA Medical Center

zachmullen:
Are tiles the right thing to be caching? Subsequent requests shouldn't be requesting the same tiles, as Dave G. mentioned, but there are elements of consecutive requests that we could cache that would reduce latency, like database lookups for authentication or other database lookups that might be repeated for multiple tiles from the same image.

Sorry if I'm misunderstanding what this code is trying to do, I'm just looking at it for the first time.

zachmullen:
I also may be misunderstanding the locality of tiles to the tile server, are they not on the same machine?

salamb:
There is a single cache for all tiles and thumbnails. The cache can be implemented either via cachetools or the tiles can be sent to Memcached. Right now I am assuming Memcached is on the local machine but that can be changed. Also besides the tiles, the data that Dave was caching with lru repoze is being cached (Jpeg tables ).

salamb:
@brianhelba 
I still need a means to decide whether to use memcached or not
PTAL

salamb:
@brianhelba 
Will not be able to test memcached without a proper config file  
PTAL

large_image/__init__.py
```diff
@@ -19,6 +19,7 @@
 
 import server
 from server import tilesource
+from server import cache_util
 getTileSource = tilesource.getTileSource  # noqa
```
brianhelba: Keep the extra newline here.

requirements.txt
```diff
@@ -1,7 +1,9 @@
 enum34==1.1.6
 jsonschema==2.5.1
-repoze.lru==0.6
+psutil==4.2.0
 six==1.10.0
+cachetools==1.1.6
```
brianhelba: Put these requirements in alphabetical order.

salamb:
@brianhelba 
There are also areas where I left some of Dave's cache functions I will delete them when I address your comments.

server/cache_util/cachefactory.py
```diff
@@ -0,0 +1,98 @@
...
+    if psutil:
+        memory = psutil.virtual_memory().total
+    else:
+        memory = 1024 ** 3
+    numItems = max(int(math.floor(memory / portion / sizeEach)), 2)
+    return numItems
+
+
+class CacheFactory():
+
```
brianhelba: Remove whitespace.

server/cache_util/cachefactory.py
```diff
@@ -0,0 +1,98 @@
...
+    def getCache(self):
+
```
brianhelba: Remove whitespace.

server/cache_util/cachefactory.py
```diff
@@ -0,0 +1,98 @@
...
+                # use memcached
+                if cacheBackEnd.lower() == 'memcached':
+                    print 'using memcached'
+                    # lock needed because pylibmc(memcached client)
+                    # is not threadsafe
+                    tileCacheLock = threading.Lock()
+
+                    # check if credentials and location exist for girder
+                    # otherwise assume location is localhost with no password
+                    url = curConfig['large_image']['cache_memcached_url']
```
brianhelba: Use `curConfig['large_image'].get('cache_memcached_url')`, since the key may not even exist.

server/cache_util/cachefactory.py
```diff
@@ -0,0 +1,98 @@
...
+                    if url is '':
```
brianhelba: Just do `if not url`, since empty string or `None` means we default.

server/cache_util/cachefactory.py
```diff
@@ -0,0 +1,98 @@
...
+
+
+class CacheFactory():
+
+    def getCache(self):
+
+        if config:
+            curConfig = config.getConfig()
+            if 'large_image' in curConfig:
+                cacheBackEnd = curConfig['large_image']['cache_backend']
```
brianhelba: Backend is 1 word, when used as jargon. The variable should be `cacheBackend`.

server/cache_util/cachefactory.py
```diff
@@ -0,0 +1,98 @@
...
+                    # is not threadsafe
+                    tileCacheLock = threading.Lock()
+
+                    # check if credentials and location exist for girder
+                    # otherwise assume location is localhost with no password
+                    url = curConfig['large_image']['cache_memcached_url']
+                    if url is '':
+                        url = 'localhost'
+
+                    memcachedUsername = curConfig['large_image']['cache'
```
brianhelba: This line split is too confusing, you can just do:
```
memcachedUsername = \
    curConfig['large_image']['cache_memcached_'username']
```

server/cache_util/cachefactory.py
```diff
@@ -0,0 +1,98 @@
...
+                                                                 '_memcached_'
+                                                                 'username']
+
+                    memcachedPassword = curConfig['large_'
```
brianhelba: Same here.

server/cache_util/cachefactory.py
```diff
@@ -0,0 +1,98 @@
...
+                                                  'image']['cache_memcached_'
+                                                           'password']
+                    if memcachedUsername == '':
```
brianhelba: Do `if not`...

server/cache_util/cachefactory.py
brianhelba: Also use `.get`.

server/cache_util/cachefactory.py
```diff
@@ -0,0 +1,98 @@
...
+
+                    if memcachedPassword == '':
+                        memcachedPassword = None
+
+                    tileCache = MemCache(url, memcachedUsername,
+                                         memcachedPassword)
+
+                    return tileCache, tileCacheLock
+
+        tileCache = LRUCache(pickAvailableCache(256**2*4))
```
brianhelba: Put this in an `else` block, and let's add a `'cache_backend'` identifier for it as well. Maybe `python`?

server/cache_util/cachefactory.py
```diff
@@ -0,0 +1,98 @@
...
+    Given an estimated size of an item, return how many of those items would
+    fit in a fixed portion of the available virtual memory.
+
+    :param sizeEach: the expected size of an item that could be cached.
+    :return: the number of items that should be cached.  Always at least two.
+    """
+    # Estimate usage based on (1 / portion) of the total virtual memory.  Each
+    # class has its own cache, and many methods have their own class, so make
+    # this conservative.
+    portion = 16
```
brianhelba: Portion should be a configuration option, and passed in as a function argument.

server/cache_util/cachefactory.py
```diff
@@ -0,0 +1,98 @@
...
+def pickAvailableCache(sizeEach):
+    """
+    Given an estimated size of an item, return how many of those items would
+    fit in a fixed portion of the available virtual memory.
+
+    :param sizeEach: the expected size of an item that could be cached.
+    :return: the number of items that should be cached.  Always at least two.
+    """
+    # Estimate usage based on (1 / portion) of the total virtual memory.  Each
+    # class has its own cache, and many methods have their own class, so make
```
brianhelba: This comment should no longer be true, if this PR works as intended.

server/cache_util/memcache.py
```diff
@@ -0,0 +1,80 @@
...
+import pylibmc
+import hashlib
+
+
+def strhash(*args, **kwargs):
+    return str(hashkey(*args, **kwargs))
+
+
+class MemCache(Cache):
+    """subclass of cache that uses a memcached store"""
```
brianhelba: All docstrings should be English sentences.

manthey:
@salamb I made a few more improvements on the caching, added tests, plus cleaned up all of the issues @brianhelba commented on.  If the part I changed looks good, let me know so we can merge this in.

manthey:
This has been a long standing PR that is fully tested.  Since travis is happy, I'll merge it in.  If we discover any problems, lets create new issues and work from that.

PR #62, opened 2016-06-01
Static annotation rendering

jbeezley:
This provides static rendering of the annotation spec in SVG (currently lacking text support).   There are currently stubs for displaying annotations inside girder's item view, but for the moment, I am implementing geojs integration in a branch inside HistomicsTK.  I think eventually some of that code should move here to create a reusable "large image viewer" component.

The rest of my work for next week's demo video will go inside a branch in [HistomicsTK](https://github.com/DigitalSlideArchive/HistomicsTK/tree/annotation-rendering).


brianhelba:
LGTM.

My only question is about the use of the `ext` directory, instead of perhaps using `extra` for a similar purpose. Let's discuss that in-person, @jbeezley .

Issue #67, opened 2016-08-04
Cache loadmodel used with tile fetches


manthey:

manthey:
@zachmullen Currently, we call `@loadmodel(model='item', map={'itemId': 'item'}, level=AccessType.READ)` to load the item record and validate that the user has access to the item before serving a tile.  Do you have some ideas on the best way to cache that?

We could get the token string (as done in `getCurrentToken` and the item ID, and if it has been recent enough (1 minute, perhaps), return the item model value.    We could listen to token, user, and item save events and invalidate the cache on those.  This means that an expired token could continue to be served tiles for that cache duration, but that we would skip three database hits per request (one for token, one for user, and one for item).  I think if we invalidate cache on appropriate save events (or even on all of them), then that would be the only security ramification.

zachmullen:
Is it 1 tile per item, or does a single item contain all of the tiles?

manthey:
One item will serve hundreds or thousands of tiles.

zachmullen:
In that case, your plan sounds like the right one. 

> This means that an expired token could continue to be served tiles for that cache duration

I think this is totally fine if we set the cache timeout to a few minutes; token expiration typically happens on the order of days, so going an extra couple minutes won't offend anyone. If the token is explicitly deleted, we can listen to that event to invalidate the cache immediately.

manthey:
Thanks.  I'll work on it some time soon.

manthey:
Added in PR #68.

PR #68, opened 2016-08-11
Cache loadmodel for token/itemId combinations of getTile.

manthey:
This reduces fetches by around 6 ms per tile (including cached tiles) in my test environment by eliminating 3 database accesses.  Additionally, an Expires header has been added to encourage short-term browser caching.

### DIFF
--- a/plugin_tests/tiles_test.py
+++ b/plugin_tests/tiles_test.py
@@ -142,7 +142,7 @@ class LargeImageTilesTest(base.TestCase):
         return infoDict
 
     def _testTilesZXY(self, itemId, metadata, tileParams={},
-                      imgHeader=JPEGHeader):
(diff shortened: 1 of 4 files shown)


manthey:
This resolves issue #67.

zachmullen:
This is nice and simple. Does our deployment already depend on memcached? If so we may want to consider using that instead of caching in the server process memory.

manthey:
memcached is an optional dependency.  We may decide to make it required at some point.

zachmullen:
Ok. LGTM

zachmullen:
(The reason I mention it is that this sort of caching scheme breaks if we have to scale to multiple app servers.)

manthey:
True -- each server would cache the model independently.  In the future we can make it use memcached if available.

Issue #66, opened 2016-08-04
Reducing logging of tile fetches


manthey:

manthey:
Added in PR #72.

PR #72, opened 2016-08-17
Filter cherrypy access logs.

manthey:
Only log 1 out of 250 tile messages.

Also, switch to the new logprint function for print statements.

Resolves issue #66.

### DIFF
--- a/server/rest/tiles.py
+++ b/server/rest/tiles.py
@@ -52,7 +52,8 @@ class TilesItemResource(Item):
         apiRoot.item.route('GET', ('test', 'tiles', 'zxy', ':z', ':x', ':y'),
                            self.getTestTile)
         filter_logging.addLoggingFilter(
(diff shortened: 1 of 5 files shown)


server/rest/tiles.py
```diff
@@ -51,6 +51,8 @@ def __init__(self, apiRoot):
         apiRoot.item.route('GET', ('test', 'tiles'), self.getTestTilesInfo)
         apiRoot.item.route('GET', ('test', 'tiles', 'zxy', ':z', ':x', ':y'),
                            self.getTestTile)
+        filter_logging.addLoggingFilter(
+            'GET (/[^/ ?#]+)*/item/[^/ ?#]+/tiles/zxy(/[^/ ?#]+){3}', 250)
```
zachmullen: Do these regexes get compiled on each request, or just once?

manthey: Just once.

zachmullen: Great. My only suggestion would be to add a kwarg for 250 so it's obvious what it means.

zachmullen:
LGTM

Issue #49, opened 2016-04-06
Report when a conversion job fails


manthey:
When a job tries to convert a file to a ptif and fails, the job's status is properly marked as ERROR, but this doesn't clear the largeImage data, nor get reported in any clear manner to the UI.

We listen for data.process to see when a new ptif file appears, but we also need to listen on model.job.save for error or cancel messages.

manthey:
Added in PR #74.

PR #74, opened 2016-08-23
When a job fails or is canceled, update the large image item to reflect that.

manthey:
Send a file to the processing job rather than an item.  This allows processing an image from an item that has more than one file.

This change was prompted by trying to convert a file with a name that differed from its containing item in a non-filesystem assetstore.

### DIFF
--- a/server/base.py
+++ b/server/base.py
@@ -20,6 +20,7 @@
 from girder import events, plugin
 from girder.constants import AccessType
 from girder.utility.model_importer import ModelImporter
+from girder.plugins.jobs.constants import JobStatus
 
 from . import constants
 from .loadmodelcache import invalidateLoadModelCache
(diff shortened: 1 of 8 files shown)


manthey:
This addresses issue #49.

manthey:
@zachmullen (or anyone else), I'd love to see this reviewed and merged.

The switch in the tests to use rabbitmq as the broker should allow all tests to pass in the vagrant development environment.

server/base.py
```diff
@@ -50,6 +51,44 @@ def _postUpload(event):
...
+    job = event.info['job']
+    meta = job.get('meta', {})
+    if (meta.get('creator') != 'large_image' or not meta.get('itemId') or
+            meta.get('task') != 'createImageItem'):
+        return
+    status = job['status']
+    if status not in (JobStatus.ERROR, JobStatus.CANCELED, JobStatus.SUCCESS):
+        return
+    item = ModelImporter.model('item').load(meta['itemId'], force=True,
+                                            exc=False)
```
zachmullen: `exc=False` is the default

zachmullen:
LGTM other than the one nit

Issue #75, opened 2016-08-24
Enhance TileSource API to support analysis of all tiles of a whole-slide image


cdeepakroy:
This issue involves enhancing the interface of the [TileSource](https://github.com/DigitalSlideArchive/large_image/blob/master/server/tilesource/base.py#L62) class to support the analysis of all tiles in a whole-slide image with a simplified interface.

To start a discussion. I have summarize below the requirements that are on my mind:
* **an iterator to iterate through all tiles of a whole-slide image at any desired magnification**: The iterator should probably return a tuple with (top-left-x, top-left-y, width, height, tile-image). @cooperlab Do we need any other information. This iterator is probably the most important among other requirements listed here.
* **A member function to add/delete layers of the image pyramid**
* We need **a member function grab corresponding tiles at each resolution**. Lee can you please elaborate on the need for this. There is a function called [getTile(x, y, z)](https://github.com/DigitalSlideArchive/large_image/blob/master/server/tilesource/svs.py#L143) which probably does this. @manthey Can you please confirm or explain what this does? 
* **A member function to get an arbitrary region in the whole slide image**: The function could take  (top-left-x, top-left-y, width, height) as arguments.  Looks like [TileSource.getRegion(width. height, **kwargs)](https://github.com/DigitalSlideArchive/large_image/blob/master/server/tilesource/base.py#L271) already does this?
* Good docstrings that clearly explain what all the member functions do.

cdeepakroy:
Below is code-snippet that will help get an idea what we do currently to do such analysis using openslide:

```python
from histomicstk.preprocessing import color_deconvolution as cd
from histomicstk.preprocessing import color_normalization as cn
from histomicstk import features as ft
from histomicstk.segmentation import label as lb
from histomicstk import segmentation as sg
import histomicstk.utils.TilingSchedule as ts
import histomicstk.utils.ConvertSchedule as cs
import histomicstk.utils as ut

import nucleicuts as nc
import numpy as np
import openslide as os
import os.path as pth
import scipy as sp
import skimage.io as io
import sys
import time

def process_slide(Input, Output, Magnification=20, T=4096, Sigma=25,
                  SigmaMin=4*(2**0.5), SigmaMax=7*(2**0.5), r=10,
                  MinArea=20, MinWidth=5, Background=1e-4,
                  Smoothness=1e-4):
    """
    """

    #define color normalization and deconvolution parameters
    W = np.array([[0.650, 0.072, 0],
[...]

PR #77, opened 2016-09-05
Tile iterator

manthey:
Added a tile iterator.  This doesn't depend on Girder.  To use, get a tile source, then, for example,
```
for tile in tilesource.tileIterator(magnification=20, format=TILE_FORMAT_NUMPY):
  Tile = tile['tile'][:, :, :3]  # three components only
  # process the Tile
```
You can use many criteria for determining which tiles you get from the image (magnification, level, size of pixels, etc.).  See the method documentation of `tileIterator` in tilesource/base.py.  Currently, the tiles that you get from the iterator can optionally be scaled.

This should be fully functional (it is used by getRegion), but needs more tests and probably more documentation.

Report magnification and pixel size of images.

Changed __init__ to allow flake8 to process it (mostly).


manthey:
This addresses issue #75.

server/tilesource/base.py
```diff
@@ -352,29 +576,218 @@ def getRegion(self, width=None, height=None, **kwargs):
...
+            gwidth, gheight: size of of the current tile in maximum resolution
+                pixels.
+        If a region that includes partial tiles is requested, those tiles are
+        cropped appropriately.  Most images will have tiles that get cropped
+        along the right and bottom egdes in any case.  If an exact
+        magnification or scale is requested, no tiles will be returned unless
+        allowInterpolation is true.
+
+        :param format: the desired format or a tuple of allowed formats.
+            Formats are members of TILE_FORMAT_*.
```
cdeepakroy: Will probably be helpful if you list all formats here with a brief description or at least put a pointer to the place where they are described. Because this is what the developers will most-likely look at when they want to create an iterator.

server/tilesource/base.py
```diff
@@ -200,12 +457,16 @@ def canRead(cls, *args, **kwargs):
...
     def getMetadata(self):
+        mag = self.getMagnification()
         return {
             'levels': self.levels,
             'sizeX': self.sizeX,
             'sizeY': self.sizeY,
             'tileWidth': self.tileWidth,
             'tileHeight': self.tileHeight,
+            'magnification': mag['magnification'],
+            'pixelWidthInMillimeters': mag['mm_x'],
```
cdeepakroy: Maybe just use `mm_x` and `mm_y`. That way people there is only way to access these properties in the outputs of `getMetadata()` and `getMagnification()`.

server/tilesource/base.py
```diff
@@ -189,6 +203,249 @@ def _getRegionBounds(self, metadata, **kwargs):
...
+            mode: either 'RGB' or 'RGBA'.  This determines the color space used
+                for tiles.
+            level: the tile level used for iteration.
+            metadata: tile source metadata (from getMetadata)
+            left, top: the coordinates within the image of the region returned
+                in the level pixel space.
+            width, height: the requested output resolution in pixels.  If this
+                is different that regionWidth and regionHeight, then the
+                original request was asking for a different scale that is being
+                delivered.
```
cdeepakroy: `regionWidth` and `regionHeight` are confusing. Can you explain what are they for or how they differ from `width` and `height`?

manthey: If you don't want the tiles rescaled and you ask for a size (width and height or magnification) that doesn't exist natively in the image, then the iterator will return tiles that are not the requested size.  `width` and `height` is the requested size, `regionWidth` and `regionHeight` is the size that is returned based on the actual tiles.

server/tilesource/base.py
```diff
@@ -352,29 +576,217 @@ def getRegion(self, width=None, height=None, **kwargs):
...
+        if level is not None and self.levels and level != self.levels - 1:
+            mag['scale'] = 2.0 ** (self.levels - 1 - level)
+            if mag['magnification']:
+                mag['magnification'] /= mag['scale']
+            if mag['mm_x'] and mag['mm_y']:
+                mag['mm_x'] *= mag['scale']
+                mag['mm_y'] *= mag['scale']
+        if self.levels:
+            mag['level'] = level if level is not None else self.levels - 1
+        return mag
```
cdeepakroy: I feel `getMagnification` and `getMagnificationForLevel` are redundant. 

Maybe, we should just have one function `getMagnification` with an optional level parameter.

cdeepakroy: Or maybe remove `getMagnification`. 

That way we have a nice pair of functions `getMagnificationForLevel` and `getLevelForMagnification` which let us go back and forth between level and magnification

server/tilesource/base.py
```diff
@@ -352,29 +576,217 @@ def getRegion(self, width=None, height=None, **kwargs):
...
+
+    def getLevelForMagnification(self, magnification=None, exact=False,
+                                 mm_x=None, mm_y=None, rounding='round',
+                                 **kwargs):
+        """
+        Get the level for a specific magnifcation or pixel size.  If the
+        magnification is unknown or no level is sufficient resolution, and an
+        exact match is not requested, the highest level will be returned.
+          At least one of magnification, mm_x, and mm_y must be specified.  If
```
cdeepakroy: Looks like you return the maximum resolution level if none of magnification, mm_x, or mm_y are provided. Update documentation accordingly,

server/tilesource/base.py
manthey: `getMagnification` always returns the maximum level's magnification, and gets subclassed by the different tile sources (svs and ptif, for instance, have different techniques for getting it).  I could make this a more private method (just rename it to `_getMagnification`), but don't want to get rid of it, as `getMagnificationForLevel` shouldn't need to be subclassed in general.

plugin_tests/sources_test.py
```diff
@@ -0,0 +1,182 @@
...
+        self.assertEqual(mag['mm_x'], 0.000252)
+        self.assertEqual(mag['mm_y'], 0.000252)
+        self.assertEqual(mag['level'], 7)
+        mag = source.getMagnificationForLevel(0)
+        self.assertEqual(mag['magnification'], 0.3125)
+        self.assertEqual(mag['mm_x'], 0.032256)
+        self.assertEqual(mag['mm_y'], 0.032256)
+        self.assertEqual(mag['level'], 0)
+        self.assertEqual(source.getLevelForMagnification(), 7)
+        self.assertEqual(source.getLevelForMagnification(exact=True), None)
```
cdeepakroy: The above two usages are weird. Maybe `source.getLevelForMagnification(exact=True)` should also return the maximum resolution level.

Alternatively, you can throw an exception if none of magnification, mm_x, and mm_y are given. People can always get the maximum resolution level via `getMetadata`. 

Also is levels 0-based index or a 1-based index?

server/tilesource/base.py
cdeepakroy: Then may be keep it as an internal method then. 

I feel it is nice to have only one way to do or get anything through the API.

server/tilesource/base.py
```diff
@@ -352,29 +576,218 @@ def getRegion(self, width=None, height=None, **kwargs):
...
+                                 mm_x=None, mm_y=None, rounding='round',
+                                 **kwargs):
+        """
+        Get the level for a specific magnifcation or pixel size.  If the
+        magnification is unknown or no level is sufficient resolution, and an
+        exact match is not requested, the highest level will be returned.
+          If none of magnification, mm_x, and mm_y are specified, the maximum
+        level is returned is exact is False.  If more than one of these values
+        is given, an average of those given will be used (exact will require
+        all of them to match).
```
cdeepakroy: I wonder if we have a use case for this average strategy. @cooperlab Can you comment on this? I fell it will be mostly mutually exclusive.

server/tilesource/base.py
```diff
@@ -287,58 +548,21 @@ def getRegion(self, width=None, height=None, **kwargs):
         :param height: maximum height in pixels.
```
cdeepakroy: Is `width` and `height` always in maximum resolution pixels (irrespective of the units parameters) ?

server/tilesource/base.py
```diff
@@ -287,58 +548,21 @@ def getRegion(self, width=None, height=None, **kwargs):
...
         :param **kwargs: optional arguments.  Some options are encoding,
             jpegQuality, jpegSubsampling, top, left, right, bottom,
-            regionWidth, regionHeight, units ('pixels' or 'fraction').
+            regionWidth, regionHeight, units ('pixels' or 'fraction'),
+            magnification, mm_x, mm_y, exact.  See _tileIteratorInfo.
```
cdeepakroy: to which all parameters does `units` apply to among - `width, height, regionWidth, regionHeight, top, left, right, bottom`?

plugin_tests/sources_test.py
manthey: I can change the `exact=True` case.  I guess my thinking was that if you asked for an exact magnification but then didn't specify anything, that would be an ill-defined request (hence the None return), but if you didn't care if it were exact, then in some sense, any level would do (so return the default, which is the highest level).

levels runs from from 0 (the single tile instance) to <number of levels> - 1 (the maximum resolution).

plugin_tests/sources_test.py
```diff
@@ -0,0 +1,183 @@
...
+def setUpModule():
+    base.enabledPlugins.append('large_image')
+    base.startServer(False)
+
+
+def tearDownModule():
+    base.stopServer()
+
+
+class LargeImageSourcesTest(common.LargeImageCommonTest):
```
cdeepakroy: Need unit tests for `getRegion`? It is probably one of the functions that will be used quite a lot apart from `tileIterator`

server/tilesource/base.py
manthey: `regionWidth, regionHeight, top, left, right, bottom`.  It is documented in `tileIterator`, and `_tileIteratorInfo`.  I can duplicate the documentation in `getRegion`.

server/tilesource/base.py
manthey: `width` and `height` are the output resolution.  I'll improve the documentation.

plugin_tests/sources_test.py
cdeepakroy: Maybe write in the docstring that, if none of the 3 magnification parameters are given, then the maximum natively available magnification is assumed. That would make things less confusing.

plugin_tests/sources_test.py
manthey: The unit tests for `getRegion` are in the `tiles_test.py` file, as they are passed through the REST region endpoint.

server/tilesource/base.py
```diff
@@ -352,29 +576,218 @@ def getRegion(self, width=None, height=None, **kwargs):
...
+            BILINEAR, or BICUBIC to resample tiles that are not the target
+            output size.  Tiles that are resampled may have non-integer x, y,
+            width, and height values, and will have an additional dictionary
+            entries of:
+                scaled: the scaling factor that was applied (less than 1 is
+                    downsampled).
+                tile_x, tile_y: (left, top) coordinates before scaling
+                tile_width, tile_height: size of the current tile before
+                    scaling.
+            Note that scipy.misc.imresize uses PIL internally.
```
cdeepakroy: Since you are returning this, i feel we should avoid exposing the `regionWidth` and `regionHeight` parameters. There are too many width and height parameters now that make things confusing.


*

manthey: Do you mean we shouldn't accept `regionWidth` and `regionHeight` as optional parameters for `getRegion` and `tileIterator`?  They can be used along with or instead of `left, top, right, bottom` to select an ROI (so not having them available would not remove functionality, but does change some of the burden on calculating an ROI).

I don't know if it exists, but I could see a utility of support millimeters as units for ROI selection, too.

server/tilesource/base.py
cdeepakroy: If magnification is given and `units='pixels'` then are (top, left, right, bottom, width, height) in pixels at the requested magnification?

Also, what use case do you have in mind for input parameters -- regionWidth, regionHeight -- when there is already width and height.

manthey: Units of `pixels` is always in the maximum-resolution layer pixels.  If we want pixels at a specified magnification, then I think we need to have more (and clearer) unit names.  `basepixels` and `magnificationpixels` (seems too long)?

A test I was doing with `getRegion` was to select an ROI via `top`, `left`, `regionWidth`, and `regionHeight`, then scale it to a "nice" size via `width` and `height`.  If, for instance, you had something that annotated the image with an ROI, then this would be useful to make thumbnails of those ROIs.

server/tilesource/base.py
cdeepakroy: Maybe you can explain this more elaborately in the meeting, but it needs to made clear what the role of regionWidth and regionHeight as input and output parameters for both getRegion and tileIterator that are part of the public API.

plugin.cmake
```diff
@@ -68,6 +69,15 @@ set_property(TEST server_large_image.cached_tiles.PythonCache APPEND PROPERTY
   "LARGE_IMAGE_CACHE_BACKEND=python"
   "LARGE_IMAGE_DATA=${PROJECT_BINARY_DIR}/data/plugins/large_image")
 
+add_python_test(sources PLUGIN large_image BIND_SERVER EXTERNAL_DATA
+  # I don't know why adding this data file thows a cmake error.
+  # "plugins/large_image/sample_image.ptif"
```
manthey: I don't understand why uncommenting this external data file results in a cmake error.  The error is
```
CMake Error: Attempt to add a custom rule to output "/home/ubuntu/misc/large_image/_build/data/plugins/large_image/sample_image.ptif.sha512-stamp.rule" which already has a custom rule.
```
I can use other external data files in multiple tests, but this throws an error.

@zachmullen Do you know what this is happening, or who I should ask?

zachmullen: Hm, I'm not super familiar with the internals of ExternalData. Brad King is the expert on that subject.

server/tilesource/base.py
cdeepakroy: Change `See _tileIteratorInfo` to `See tileIterator` since _tileIteratroInfo is a private function.

Also, i would suggest renaming `width, height` to something like `maxOutWidth, maxOutHeight` and `regionWidth,  regionHeight` to `width, height`. This would weed away most of the confusion.

Also, maybe make an alias function called `getROI` without the `maxOutWidth` and `maxOutHeight` parameters. This intention of this function should be solely to get an ROI at any magnification.

Also, i would suggest using `units=pixels` to mean specified resolution pixels and add `units=base_pixels` to mean maximum resolution pixels .

cdeepakroy:
@manthey Few more comments.

@cooperlab Can you please take a look at this PR and comment on the API of the TileSource class?

server/tilesource/base.py
cdeepakroy: `A test I was doing with getRegion was to select an ROI via top, left, regionWidth, and regionHeight, then scale it to a "nice" size via width and height. If, for instance, you had something that annotated the image with an ROI, then this would be useful to make thumbnails 
of those ROIs.` 

This use case speaks to me that the functionality needed to support this should go into the getThumbnail function.

Just an idea: How about making the super flexible function getRegion private and expose only getThumbnail and getROI that are more focused on the use cases?

server/tilesource/base.py
cooperlab: @manthey @cdeepakroy many people refer to the max level magnification as "scanning magnification" or "native magnification" - perhaps we can change `getMagnification` to `getScanMagnification` or `getNativeMagnification`.

cooperlab:
@cdeepakroy @manthey I looked but probably need to walk through this with someone - or to implement the use cases we described. Will have more time to look at this later this week.

manthey:
I still have more changes coming based on @cdeepakroy's comments (renaming parameters, at least).

manthey:
One of the reasons width and height refer to the output dimensions is because getRegion functions as a superset of getThumbnail, and I'd expect callers of getThumbnail are only interested in width, height, and possibly encoding. There is no requirement that getRegion have the same parameter names as getThumbnail, so we can diverge the two calls.

One method to improve clarity is to rename `width` and `height` to `maxOutWidth` and `maxOutHeight`, and rename `left, top, right, bottom` to `regionLeft, regionTop, regionRight, regionBottom` (or change the `region` prefix to be `source` on these and on `regionWidth` and `regionHeight`). The thumbnail function and REST endpoint could still use `width` and `height`, but the region function and REST endpoint would not (and could even throw an exception if `width` or `height` were in the kwargs).

Another option is that the getRegion function could take nested parameters: `region: {left, top, right, bottom, width, height, units}, output: {maxWidth, maxHeight, magnification, mm_x, mm_y, exact, resample}, format, encoding, jpegQuality, jpegSubsampling`.

@cdeepakroy, opinions?

cdeepakroy:
@manthey 

> One method to improve clarity is to rename width and height to maxOutWidth and maxOutHeight, and rename left, top, right, bottom to regionLeft, regionTop, regionRight, regionBottom. The thumbnail function and REST endpoint could still use width and height, but the region function and REST endpoint would not (and could even throw an exception if width or height were in the kwargs).

I like this.

> region: {left, top, right, bottom, width, height, units}, output: {maxWidth, maxHeight, magnification, mm_x, mm_y, exact, resample}, format, encoding, jpegQuality, jpegSubsampling

I also like this because of the way it groups parameters but it will require a bit of extra work in creating the region and output dictionaries before calling the getRegion function. 

One issue you will face with both is - what will the function do if all of {left, top, right, bottom, width, height, units} are passed.

Maybe you could define two functions:

```python
def region_ltrb(left, top, right, bottom, units='base_pixels'):
    return {
        'units': units,
        'left': left,
        'top': top,
        'bottom': bottom,
        'right':  right,
        'width': right - left + 1,
        'height': bottom - top + 1
    }

def region_ltwh(left, top, width, height, units='base_pixels'):
    return {
        'units': units,
        'left': left,
        'top': top,
        'width': width,
        'height': height,
        'bottom': top + height,
        'right':  left + width,
    }
```

You could even create an abstract `region` class and define two concrete subclasses `region_ltrb` and `region_ltwh` and pass a region object in.

manthey:
@cdeepakroy I've refactored tileIterator (and getRegion) to use the dictionaries of region and output parameters.  Note that the code already handled specifying any two of `left, right, width` and any two of `top, bottom, height` for the region.

I think that documentation of using this should be in the next PR (possibly adding a setup.py file at that time, too).

cdeepakroy:
@manthey Looks good! This should be enough to enable HistomicsTK to start processing large images.

The only suggestion i have is to not include `magnification, mm_x, mm_y, exact, resample` in the output dictionary as they dont seem to fit well under it that well.  These parameters are for describing at which magnification to region spec belongs to, and what to do if that magnification is not available natively.

cdeepakroy:
Rest LGTM!

manthey:
@cdeepakroy Do you think it would be better to have `magnification, mm_x, mm_y, exact` in another parameter set (`scale` perhaps?), or just move them back to the base-level parameters?

cdeepakroy:
Another parameter set for `magnification, mm_x, mm_y, exact` is a good idea!

manthey:
@cdeepakroy I've added this as the `scale` parameter set.  If you want to look it over one more time before merge, I'd appreciate it.

cdeepakroy:
LGTM! Merge!

PR #81, opened 2016-09-13
Don't require memcache. 

manthey:
This should allow easier deployment to Windows.

This doesn't install pylibmc on Windows (if it happens to be installed, it will be used).

This changes requirements to >= instead of ==, which reduces install thrashing.

Assuming that you've installed OpenSlide and libtiff on Windows, examples/average_color.py now works on Windows.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -130,7 +130,8 @@ before_install:
 
 install:
   - cd $girder_path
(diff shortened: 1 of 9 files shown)


jbeezley:
I can't test the Windows improvements, but LGTM nonetheless.

manthey:
I'll leave writing an ansible script to test Windows as a distant future project.

PR #86, opened 2016-09-15
Add guards against memcached failures.

manthey:
If memcached had the wrong server address, it would fail.  Now, log the error (throttled) and don't use the cache.  Improve the retry options for the memcached client.  Add a test for the failure conditions.

Allow specifying a list of memcached servers in the config file.

### DIFF
--- a/server/cache_util/memcache.py
+++ b/server/cache_util/memcache.py
@@ -17,9 +17,11 @@
 #  limitations under the License.
 #############################################################################
 
-from cachetools import Cache
-import pylibmc
+import cachetools
 import hashlib
+import pylibmc
+import six
+import time
 
 try:
     from girder import logprint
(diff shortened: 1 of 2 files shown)


manthey:
@zachmullen You might be the best one to review this change.  @dgutman ran into this problem earlier today.

server/cache_util/memcache.py
```diff
@@ -17,27 +17,45 @@
...
+        # See http://sendapatch.se/projects/pylibmc/behaviors.html
+        behaviors = {
+            'tcp_nodelay': True,
+            'ketama': True,
+            'no_block': True,
+            'retry_timeout': 1,
+            'dead_timeout': 10,
+        }
+        # Adding remove_failed prevents recovering in a single memcached server
+        # instance, so onlydo it if there are multiple servers
```
zachmullen: only do*

zachmullen:
@manthey other than the small typo LGTM. I'd love to pull out the generic caching stuff from this plugin into girder/girder soon.

PR #87, opened 2016-09-20
data.process events are async, so adjust how they are handled

manthey:
data.process events are async, so adjust how they are handled, as they may occur after the job finishes.


manthey:
@zachmullen When you get a change, can you review this PR?  Any others you'd like to handle would be appreciated, too.

manthey:
Thank you.

PR #88, opened 2016-09-20
Shift an info message to a debug message so the logs don't contain needless warnings.

manthey:
Before, some file that could be read as svs files would log a warning when the tif reader reported that it wouldn't handle them.


cdeepakroy:
LGTM!

Issue #90, opened 2016-09-21
Using as girder plugin causes girder crash


Purg:
Trying to use as a girder plugin (ubuntu 14.04) and am installing `libmemcached-dev`, `libopenslide-dev` to the system and `numpy` with pip to a virtual env.

When trying to run `girder-server`, I get the following output:
```
Running in mode: development
Connected to MongoDB: mongodb://localhost:27017/girder
INFO:girder:Using memcached for large_image caching
Using memcached for large_image caching
Traceback (most recent call last):
  File "/home/vagrant/.virtualenvs/test/bin/girder-server", line 11, in <module>
    load_entry_point('girder', 'console_scripts', 'girder-server')()                                                           
  File "/home/vagrant/girder/girder/__main__.py", line 50, in main                                                             
    server.setup(args.testing)                                                                                                 
  File "/home/vagrant/girder/girder/utility/server.py", line 145, in setup                                                     
    root, appconf = configureServer(test, plugins, curConfig)                                                                  
  File "/home/vagrant/girder/girder/utility/server.py", line 115, in configureServer                                           
    plugins, curConfig, ignoreMissing=True))                                                                                   
  File "/home/vagrant/girder/girder/utility/plugin_utilities.py", line 114, in getToposortedPlugins                            
    allPlugins = findAllPlugins(curConfig)                                                                                     
  File "/home/vagrant/girder/girder/utility/plugin_utilities.py", line 362, in findAllPlugins                                  
    findEntryPointPlugins(allPlugins)                                                                                          
  File "/home/vagrant/girder/girder/utility/plugin_utilities.py", line 349, in findEntryPointPlugins                           
    data = getattr(entry_point.load(), 'config', {})                                                                           
  File "/home/vagrant/.virtualenvs/test/local/lib/python2.7/site-packages/pkg_resources/__init__.py", line 2258, in load   
    return self.resolve()                                                                                                      
  File "/home/vagrant/.virtualenvs/test/local/lib/python2.7/site-packages/pkg_resources/__init__.py", line 2268, in resolve
    raise ImportError(str(exc))                                                                                                
ImportError: 'module' object has no attribute 'load'                                                                           
```
I think this is occurring when trying to start the large image plugin. I'm not sure if this is a memcached issue or if the module isn't revealing the assumed load function. I tried deciphering the `setup.py` entry point and it looks like `server/__init__.py` should be exposing the `load` function defined in `server/base.py`, unless the error above is in reference to a different load function.

[...]

PR #91, opened 2016-09-22
When importing via an entrypoint for use in Girder, defer importing fellow plugin values.

manthey:
This fixes issue #90.


manthey:
@purg I believe you confirmed that this fixed your issues.  Can I go ahead and merge it?

Purg:
Yes it does!

Issue #89, opened 2016-09-20
Chrome limits number of connections


manthey:
Chrome limits the number of connections, which makes fetching thumbnails a blocking process.  Implement a ring buffer or some other workaround.

manthey:
Resolved by PR #93.

PR #93, opened 2016-09-27
Limit the number of simultaneous thumbnails that are requested in the itemList.

manthey:
Chrome limits the maximum connectors to a domain to 6.  Girder uses one for getting events, leaving five.  This uses no more than 3 for thumbnails, which leaves some overhead for navigation.  If you navigate between item lists quickly, the connections can still be saturated, as Chrome doesn't cancel the image requests.  We would have to load the image to a hidden element first, and then load it to the visible element to work around that, so that the hidden elements don't vanish on page transitions.

This resolves issue #89.

### DIFF
--- a/web_client/js/itemList.js
+++ b/web_client/js/itemList.js
@@ -15,14 +15,14 @@ girder.wrap(girder.views.ItemListWidget, 'render', function (render) {
      *      thumbnails are located.
      */
     function _loadMoreImages(parent) {
-        var loading = $('.large_image_thumbnail img.loading', parent).length;
-        if (maxSimultaneous > loading) {
-          $('.large_image_thumbnail img.waiting', parent).slice(0, maxSimultaneous - loading).each(function () {
-            var img = $(this);
-            img.removeClass('waiting').addClass('loading');
-            img.attr('src', img.attr('deferred-src'));
-          });
-        }
+      var loading = $('.large_image_thumbnail img.loading', parent).length;
+      if (maxSimultaneous > loading) {
+        $('.large_image_thumbnail img.waiting', parent).slice(0, maxSimultaneous - loading).each(function () {
+          var img = $(this);
+          img.removeClass('waiting').addClass('loading');
+          img.attr('src', img.attr('deferred-src'));
+        });
+      }
     }
 
     render.call(this);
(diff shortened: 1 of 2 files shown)


manthey:
@dgutman This will substantially improve the browser getting stuck when navigating between folders.  If you find through rapid navigation that this still happens, we can fetch images in a hidden element before showing them so that we track what is happening between navigation.  I don't want to implement that unless it is necessary, as it adds complexity.

manthey:
@zachmullen You are probably the best person to review this, when you get a chance.

zachmullen:
I would expect that deleting the `img` element would automatically cause the request for its data to be canceled, is that not the case?

manthey:
The request is still being processed by the server.  I can't tell if they are still counted by Chrome.

web_client/js/itemList.js
```diff
@@ -1,4 +1,30 @@
...
+     *      thumbnails are located.
+     */
+    function _loadMoreImages(parent) {
+        var loading = $('.large_image_thumbnail img.loading', parent).length;
+        if (maxSimultaneous > loading) {
+          $('.large_image_thumbnail img.waiting', parent).slice(0, maxSimultaneous - loading).each(function () {
+            var img = $(this);
+            img.removeClass('waiting').addClass('loading');
+            img.attr('src', img.attr('deferred-src'));
+          });
```
zachmullen: Are we running eslint on this? There are a few style issues here.

zachmullen:
> The request is still being processed by the server. I can't tell if they are still counted by Chrome.

I guess the server wouldn't know until it actually tried to send the bytes back down. This may be something the network tab in chrome can show, in particular if we introduce an artificial delay on the server side.

web_client/js/itemList.js
manthey: We are running eslint, but oddly, the plugin was given its own .eslintrc file, rather than using Girder's.

zachmullen: That's fine. It might be good to enforce a uniform indentation size, but that's out of scope of this issue.

manthey: I've fixed this indentation.  I'll make an issue to move to Girder's eslint style, as I don't see any reason not to.

manthey:
I just placed a 30 second delay on all large image thumbnails.   When I navigate between two folders that have large images, the previous pending images remain in Chrome's network list, and DO block things until they resolve.

I can either work around that as an extension to this PR, or only make a separate PR when we decide that it is a real-world problem and not just when I artificially introduce it.

zachmullen:
I think it's fine, mostly just curiosity.

Issue #83, opened 2016-09-13
Add job to create thumbnails


manthey:
This should take a list of thumbnail sizes, and be cancellable.

manthey:
Added in PR #96.

PR #96, opened 2016-09-27
Add rest endpoints for managing thumbnail files.

manthey:
Add the ability to delete thumbnail files (this facilitates testing and frees disk space).

Add an endpoint that starts a local job that creates thumbnail files.

This resolves issue #83, though we could want a setting to run a cron-like job at some point.


manthey:
This is ready for review.

plugin_tests/common.py
```diff
@@ -224,6 +235,11 @@ def _postTileViaHttp(self, itemId, fileId):
         self.assertTrue('Item already has' in req.json()['message'] or
                         'Item is scheduled' in req.json()['message'])
 
+        if jobAction == 'delete':
+            jobModel = self.model('job', 'jobs')
+            jobModel.remove(jobModel.find(
+                {}, sort=[('_id', SortDir.DESCENDING)])[0])
```
zachmullen: This is an odd construct, what's happening here?

server/models/image_item.py
```diff
@@ -256,6 +258,8 @@ def getThumbnail(self, item, width=None, height=None, **kwargs):
         """
         # check if a thumbnail file exists with a particular key
         keydict = dict(kwargs, width=width, height=height)
+        for key in [key for key in keydict if keydict[key] is None]:
```
zachmullen: A dict comprehension would probably be more readable here.

    keydict = {k: v for k, v in six.viewitems(keydict) if v is not None}

plugin_tests/common.py
manthey: This deletes the most recently created job.  Mongo's _ids increase monotonically, so the sort returns the highest (most recent) first.

Issue #94, opened 2016-09-27
return image as numpy array from tile_source.getRegion


cdeepakroy:
This issues involves adding an option to [tile_source.getRegion](https://github.com/DigitalSlideArchive/large_image/blob/master/server/tilesource/base.py#L620) to return the image as a numoy array.

manthey:
Added in PR #97.

PR #97, opened 2016-09-27
Allow getRegion to return different formats.

manthey:
Resolves issue #94.


manthey:
@cdeepakroy `getRegion` returns the image (if the desired format) and either the format OR the mimetype (if the format is an image).  This keeps the return the same as before this was added.  If desired, I could shift things so the return value of `getRegion` is a  dictionary of {image: <data>, format: <format>[, mimeType: <mimeType>}, but that would potentially break anything you are currently using.

cdeepakroy:
LGTM! The tuple `image, format` as return value is good with me.

Issue #95, opened 2016-09-27
Add getRegionAtAnotherScale() - converts region from one scale to another


cdeepakroy:
Considering the region spec is the following dictionary:

* `left`: the left edge (inclusive) of the region to process.
* `top`: the top edge (inclusive) of the region to process.
* `right`: the right edge (exclusive) of the region to process.
* `bottom`: the bottom edge (exclusive) of the region to process.
* `width`: the width of the region to process.
* `height`: the height of the region to process.
* `units`: either 'base_pixels' (default), 'pixels', 'mm', or 'fraction'.

Add a function `getRegionAtAnotherScale(source_region, target_scale, source_scale=None)` that takes a region at `source_scale` and returns region at the `target_scale`. If units of `source_region` is `pixels`, then `source_scale` must not be None.

cdeepakroy:
Also, later the usecase of converting a region spec from one type of units to another might come up.

manthey:
In the event that the `units` of the source region is anything other than `pixels`, then this function ends up being just `getRegion(region=source_region, output=target_scale, **kwargs)`.  In the `units = pixels` case, more ends up being necessary.

I'll add a `convertRegionScale(source_region, target_scale, source_scale=None)` function that will output a region that can be passed to `getRegion` or `tileIterator`.  Do we need `tileIteratorAtAnotherScale` besides `getRegionAtAnotherScale`?

cdeepakroy:
I don't have a use case for `tileIteratorAtAnotherScale` at the moment. But you may want to create an issue for it and we can address it when a use case arises. I can see some uses for it but not at the moment.

manthey:
Added in PR #98.

PR #98, opened 2016-09-28
Added several region scale utility functions.

manthey:
`convertRegionScale` converts a region from its current units to a different set of units.

`getRegionAtAnotherScale` and `tileIteratorAtAnotherScale` use that function to provide their core features at an alternate scale.

This resolves issue #95.


cdeepakroy:
LGTM!

Issue #84, opened 2016-09-13
Support small images


manthey:
Optionally support small images without having to convert them to a tiled image.  There should be an upper size limit beyond which this refuses to work so as not to use too much memory.

The UI should only show these if they have been marked in some way, which could happen automatically (off by default).

dgutman:
So this works automatically with open sea dragon using legacy tile
sources... I can send the syntax.

manthey:
Added in PR #99.

PR #99, opened 2016-09-30
Small image support

manthey:
Allow small images to be used as single tiles.

When used as a Girder plugin, the maximum size that is allowed is stored in a setting (a value of 0 disables direct use of small images).  When used without Girder, the maximum size is specified in server/tilesource/pil.py.  In either case, it can be overridden when asking for a tile source with the maxSize parameter.

This also includes an update to the geojs viewer to better handle images with massive tiles.  It will also properly crop edges when we update the geojs version.


manthey:
This resolves issue #84.

manthey:
@cdeepakroy This still needs some tests, but otherwise is fully functional.  Currently, the limit for using a single image is set to 4096x4096 pixels.

manthey:
@cdeepakroy This PR should be ready for testing.

cdeepakroy:
LGTM! I just tested it and it works!

PR #102, opened 2016-10-10
Add edge options for tiles.

manthey:
Tiles at the right and bottom edges of images can now be cropped or have unused areas converted all to a specific color.  With appropriate query strings, this allows all viewers to show the edges of images in correct manner.

Some of the tile output code has been moved into the base class to avoid repetition.  As a benefit, the TIFF tilesource can output tiles as PNGs if so desired.

Removed some vestigial code related to class cache keys.

### DIFF
--- a/server/tilesource/base.py
+++ b/server/tilesource/base.py
@@ -47,6 +47,8 @@ try:
         logger.warning('Error: Pillow v3.0 or later is required')
         PIL = None
     import PIL.Image
+    import PIL.ImageColor
+    import PIL.ImageDraw
 except ImportError:
     logger.warning('Error: Could not import PIL')
     PIL = None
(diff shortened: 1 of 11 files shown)


manthey:
With these changes, OpenSeaDragon will render NDPI files with very wide and short rectangular tiles properly.  Other viewers will also do better (geojs already does the right thing).

server/tilesource/base.py
```diff
@@ -96,25 +98,45 @@ class TileSource(object):
...
             self.cache, key=self.wrapKey, lock=self.cache_lock)(self.getTile)
 
-    def wrapKey(self, *args, **kwargs):
-        return strhash(self.getState()) + strhash(*args, **kwargs)
-
     @abc.abstractmethod
     def getState(self):
-        return None
+        return str(self.encoding) + ',' + str(self.jpegQuality) + ',' + str(
+            self.jpegSubsampling) + ',' + str(self.edge)
```
zachmullen: Is there a reason to have an implementation of a declared `abstractmethod`? Also, doesn't using `abc.abstractmethod` require you to specify `__metaclass__ = abc.ABCMeta`?

web_client/js/imageViewerWidget/leaflet.js
```diff
@@ -38,7 +38,8 @@ girder.views.LeafletImageViewerWidget = girder.views.ImageViewerWidget.extend({
                 [90.0, 180.0]
             ],
             layers: [
-                L.tileLayer(this._getTileUrl('{z}', '{x}', '{y}'), {
+                L.tileLayer(
+                    this._getTileUrl('{z}', '{x}', '{y}') + '?edge=%23DDD', {
```
zachmullen: Does `this._getTileUrl` make the guarantee that it won't include a query string in its returned URL?

jbeezley:
I looked over this and didn't have any specific concerns.  I was attempting to do manual testing, but my python environment had other ideas.

Issue #100, opened 2016-09-30
Add a function to tile_source to get a specific tile at any level or magnification


cdeepakroy:

manthey:
Would you like this to return a dictionary of information like the tile iterator returns per tile, or would you like it to just return the tile?

We could treat this exactly like the tile iterator, excepting that a `tile = {x: <x>, y: <y>}` is specified as if within all of the other specifications of the iterator (that is, if you specify a region, `x=0, y=0` is the top-left portion of that region, not of the whole image).   If a tile `level` is specified, then the specific image level would be used, regardless of what would be selected by the `region` and `scale` specifications.

Thoughts?

cdeepakroy:
I feel it should take the same parameters that tileIterator takes for specifying `region`, `scale` etc

I feel the purpose of this function to get one specific tile directly instead of iterating to find it.

I think the dictionary like the one that tileIterator returns at each iteration will be useful.

manthey:
Added in PR #103.

PR #103, opened 2016-10-12
Add functions for returning a single tile

manthey:
This adds two functions: `getSingleTile` and `getSingleTileAtAnotherScale`.  These functions take exactly the same parameters as `tileIterator`, but will return the first tile from the iterator.

`tileIterator` now takes an optional parameter `tile_position` which may be a 0-based integer which is functionally the same as `iter(list(tileIterator(...))[tile_position])`.  `tile_position` may also be a dictionary of either {i, j} or {x, y}, where {i, j} are the 0-based tile positions within the iterator, and {x, y} are the level-based locations of the tile at the iterated level.

The tile information yielded by the iterator now includes `tile_position` and `iterator_range` which include information about the tile position in the iterator and the ranges of x, y, i, j, and position that the iterator produces.

This resolves issue #100.


manthey:
@cdeepakroy Let me know if you want any changes.  In the simplest case, call `getSingleTile` with the same parameters as `tileIterator`, and it will return the first tile.

server/tilesource/base.py
```diff
@@ -1108,6 +1169,24 @@ def tileIteratorAtAnotherScale(self, sourceRegion, sourceScale=None,
                                          targetScale, targetUnits)
         return self.tileIterator(region=region, scale=targetScale, **kwargs)
 
+    def getSingleTile(self, *args, **kwargs):
+        """
+        Return the first tile from an iterator.  This takes exactly the same
+        parameters as tileIterator.
```
cdeepakroy: Update the documentation to say it can be used to get any specific tile (not just the first tile).

server/tilesource/base.py
```diff
@@ -1108,6 +1169,24 @@ def tileIteratorAtAnotherScale(self, sourceRegion, sourceScale=None,
...
+
+        :return: a tile dictionary or None.
+        """
+        return next(self.tileIterator(*args, **kwargs), None)
+
+    def getSingleTileAtAnotherScale(self, *args, **kwargs):
+        """
+        Return the first tile from an rescaled iterator.  This takes exactly
+        the same parameters as tileIteratorAtAnotherScale.
```
cdeepakroy: Update the documentation to say it can be used get any specific tile at another scale.

server/tilesource/base.py
```diff
@@ -340,6 +340,12 @@ def _tileIteratorInfo(self, **kwargs):
             mm_y: the vertical size of a pixel in millimeters.
             exact: if True, only a level that matches exactly will be returned.
                 This is only applied if magnification, mm_x, or mm_y is used.
+        :param tile_position: if present, either a number to only yield the
+            (tile_position)th tile [0 to (xmax - min) * (ymax - ymin)) that the
+            iterator would yield, or a dictionary of {i, j} to yield that tile,
+            where 0, 0 is the first tile yielded, and xmax - xmin - 1,
+            ymax - ymin - 1 is the last tile yielded, or a dictionary of {x, y}
+            to yield that specific tile if it is in the region.
```
cdeepakroy: I would suggest using more meaningful names:

* {tx, ty} instead of {i, j} and 
* {lx, ly} instead of {x, y}

cdeepakroy:
@manthey Few comments. Rest LGTM!

manthey:
@cdeepakroy I think this addresses your comments.

cdeepakroy:
LGTM!

PR #106, opened 2016-10-17
Fix some caching issues

manthey:
The LRU cache was using the wrong key.

The metaclass cache was using some of the query parameters for the keys, resulting in few cache hits for certain functions (like `getRegion`).  This fixes the cache keys.

Fix some caching issues.

Because of how the `cached` function from cachetools was being applied to class methods, those instantiated classes would never be garbage collected.  This led to leaving files open and exhausting resources.  A new decorator was added called `methodcache` which behaves akin to the `cachetools.cached` decorator, except that it uses class caches, locks, and, optionally, key functions.

This still needs tests.

### DIFF
--- a/plugin_tests/cached_tiles_test.py
+++ b/plugin_tests/cached_tiles_test.py
@@ -47,33 +47,32 @@ def tearDownModule():
 
 class LargeImageCachedTilesTest(common.LargeImageCommonTest):
     def _monitorTileCounts(self):
-        if hasattr(self, 'tileCounter'):
-            return
-        from girder.plugins.large_image.tilesource.test import TestTileSource
-        originalGetTile = TestTileSource.getTile
-        originalWrapKey = TestTileSource.wrapKey
-        self.tileCounter = 0
-        self.keyPrefix = str(time.time())
+        from girder.plugins.large_image.tilesource import test
+        test.tileCounter = 0
 
-        def countGetTile(ttsself, *args, **kwargs):
-            # Increment the counter after the call, so that exceptions won't
-            # increment it.
-            result = originalGetTile(ttsself, *args, **kwargs)
-            self.tileCounter += 1
-            return result
+        self.originalWrapKey = test.TestTileSource.wrapKey
+        self.keyPrefix = str(time.time())
 
         def wrapKey(*args, **kwargs):
             # Ensure that this test has unique keys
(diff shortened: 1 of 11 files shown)


manthey:
@law12019 This will fix your problem with using regions.

manthey:
@zachmullen This is ready for review, if you have time for it.

server/cache_util/cachefactory.py
```diff
@@ -55,6 +55,8 @@ def pickAvailableCache(sizeEach, portion=8):
     else:
         memory = 1024 ** 3
     numItems = max(int(math.floor(memory / portion / sizeEach)), 2)
+    if maxItems:
+        numItems = max(numItems, maxItems)
```
zachmullen: Shouldn't this be `min(numItems, maxItems)`? Also we should add a docstring for the new param.

server/cache_util/cache.py
```diff
@@ -17,14 +17,52 @@
...
+                if lock:
+                    with self.cache_lock:
+                        self.cache[k] = v
+                else:
+                    self.cache[k] = v
+            except ValueError:
+                pass  # value too large
+            return v
+        return wrapper
+    return decorator
```
zachmullen: An alternative way to do this is to use a decorator class and implement `__init__` and `__call__`, but if you prefer this way that's fine.

manthey: Yes -- I copied the method from the `cachetools` module and then changed it for our use.  Since it is based on that code, I decided to leave the structure as it.

plugin_tests/cached_tiles_test.py
```diff
@@ -47,33 +48,32 @@ def tearDownModule():
...
-            return result
+        self.originalWrapKey = test.TestTileSource.wrapKey
+        self.keyPrefix = str(time.time())
 
         def wrapKey(*args, **kwargs):
             # Ensure that this test has unique keys
-            return self.keyPrefix + originalWrapKey(*args, **kwargs)
+            return self.keyPrefix + self.originalWrapKey(*args, **kwargs)
+
+        test.TestTileSource.wrapKey = wrapKey
```
manthey: I'm not sure how to use `mock` in this case.  I assume I'd pass my modified wrapKey function as a `new_callable` in some manner, but I'm not sure how to get the class to be correct and have the context last for each test, since there is a separate setUp and tearDown.

PR #110, opened 2016-10-21
Move the settings endpoint.

manthey:
Move GET system/setting/large_image to GET large_image/settings to keep the namespace cleaner.


server/rest/large_image.py
```diff
@@ -83,15 +83,31 @@ def createThumbnailsJob(job):
 
 class LargeImageResource(Resource):
 
-    def __init__(self):
+    def __init__(self, apiRoot):
```
zachmullen: It looks like we don't need to pass in this apiRoot.

Issue #108, opened 2016-10-20
Many large image conversions exhaust file handles


manthey:
File handles are still being cached somewhere, since converting a large number of images to large images leaves handles open.

manthey:
Fixed in PR #112.

PR #112, opened 2016-10-24
Better tile source caching.

manthey:
Prior to this, each tile source class used a separate cache.  This makes memory use difficult to manage intelligently.  The cache size needs to be set based on available file handles, as that is typically the limiting factor.  TIFF tile sources tend to have one open file handle per level of tiles, so the maximum number of tile sources that can be cached is now set based on that resource if it can be queried.  Moreover, ask to use more file handles if possible (it may not be).

Resolves issue #108.

### DIFF
--- a/plugin_tests/cached_tiles_test.py
+++ b/plugin_tests/cached_tiles_test.py
@@ -177,7 +177,7 @@ class LargeImageCachedTilesTest(common.LargeImageCommonTest):
         itemId = str(file['itemId'])
         item = self.model('item').load(itemId, user=self.admin)
         # Clear the cache to free references and force garbage collection
-        LruCacheMetaclass.caches[TiffGirderTileSource].clear()
+        LruCacheMetaclass.classCaches[TiffGirderTileSource].clear()
         gc.collect(2)
         self.initCount = 0
         self.delCount = 0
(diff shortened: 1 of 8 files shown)


zachmullen:
I'm not that familiar with this, could you give some background? How do file descriptors relate to caching? What exactly is being cached?

manthey:
When we access a tile source (usually a tif or svs file), we call a relatively expensive function to open it.  For tif files, each tile layer is opened separately.  For svs files, the file is opened with the openslide module.  SO as to not repeat this work if we use them again soon, the instance of the tilesource class that opens the file(s) hangs on to those open handles.  We cache the class instance.  Prior to this PR, each tilesource class had a separate LRU cache (from the cachetools module).  Until the class instance falls out of the cache, it maintains an open reference to its file(s).  As file handles are a scarce resource (defaults to 1024 in Ubuntu 14.04), after opening roughly that many svs files or a fraction of that many tif files, no file handles were available, and all sort of tasks failed (such as importing another python module).

With this PR, there is a single tile source cache (rather than one per tile source class).  Moreover, the maximum size of the cache is taken as the number of available file handles less a reserved quantity and divided by an arbitrary factor (since tif sources use multiple handles).  Crudely, most images we are dealing with use around 10 file handles or less, but I've used a factor of 20 for safety.  A further sophistication would be to determine the number of handles used by the source and weight it that way (cachetools supports such weights), but that hasn't been implemented.

zachmullen:
Hm, I'm a bit surprised that the process of opening the file would be less expensive than actually reading it from the drive. My expectation for a tile cache would be that the actual image data is being stored in memory, rather than handles to it.

That said, if we are convinced that the process of acquiring the descriptors is the most practical cost to avoid, then this LGTM.

manthey:
The tiff library and the openslide libraries both perform a fair amount of work when opening the files.  While reading it from the drive is certainly more time consuming, this definitely has practical benefits.  There are a few svs files which consistently take more than 500 ms to open, even if they were just opened (so that the file data should be in the read cache).

zachmullen:
> There are a few svs files which consistently take more than 500 ms to open, even if they were just opened

Yowch!

PR #114, opened 2016-11-02
Flexible ptif support

manthey:
Support ptifs with odd directory orders.

Before, we required that ptifs have their directories in a fixed order (highest res tiles to lowest res, then non-tiled image).  Now, directories can be in any order, and may be sparse.  Much like svs files, if a level doesn't exist, tiles will be constructed for it from higher-resolution levels.

This also allows non-square tiles but requires tiles on all levels be the same size.

Also, if a PTIF contains JPEG 2000 compressed data, extract each tile through PIL and allow it to be reencoded as desired (typically either to JPEG or PNG).  For the tile iterators, this may also return the PIL image.  This requires OpenJpeg 2.x to be installed.


manthey:
@zachmullen Would you mind reviewing this when you get a chance?

zachmullen:
The code LGTM, but I haven't actually tested it.

PR #115, opened 2016-11-07
Add JFIF to jpegs from ptifs

manthey:
iOS 10.1 won't load a JPEGs without a JFIF record, even though it should be optional and ignored.

### DIFF
--- a/server/tilesource/tiff_reader.py
+++ b/server/tilesource/tiff_reader.py
@@ -245,7 +245,7 @@ class TiledTiffDirectory(object):
         # Strip the Start / End Of Image markers
         tableData = tableBuffer[2:tableSize - 2]
         # Add JFIF information to the header to keep iOS 10 happy
-        tableData = b'\xff\xe0\x00\x10JFIF\x00\x01\x01\x01\x00\x48\x00\x48\x00\x00' + tableData
+        tableData = b'\xff\xe0\x00\x10JFIF\x00\x01\x01\x00\x00\x01\x00\x01\x00\x00' + tableData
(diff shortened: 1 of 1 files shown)


brianhelba:
@manthey Which marker segments is it expecting? If they're any decent size, perhaps we should consider including them only conditionally, depending on the requesting User-Agent.

manthey:
@brianhelba Just the JFIF APP0 marker segment, which is 18 bytes.

server/tilesource/tiff_reader.py
```diff
@@ -244,6 +244,8 @@ def _getJpegTables(self):
 
         # Strip the Start / End Of Image markers
         tableData = tableBuffer[2:tableSize - 2]
+        # Add JFIF information to the header to keep iOS 10 happy
+        tableData = b'\xff\xe0\x00\x10JFIF\x00\x01\x01\x01\x00\x48\x00\x48\x00\x00' + tableData
```
brianhelba: Why does this use `\x01\x01` (v1.01) for the version? 1.02 is the current revision.

Why does this use `\x01` ("units = 1: X and Y are dots per inch") for the units? When not otherwise known, I think we'd want to specify "units = 0: no units, X and Y specify the pixel aspect ratio".

Why does this use `\x00\x48` (48 DPI) for the Xdensity and Ydensity? Especially if we switch to making the units be dimensionless, I think these should both be 1.

law12019: LGTM.  The patch to SlideAtlas works.

PR #116, opened 2016-11-15
Support libtiff 4.0.6.

manthey:
If we use OpenJPEG 2.1.1 or later, it pulls in libtiff 4.0.6, in which case pylibtiff 0.4 no longer fully works.

Add support for pylibtiff's master branch.  This requires being case-agnostic for certain tiff file methods.


zachmullen:
LGTM

Issue #117, opened 2016-11-21
Bad files stop thumbnail creation job


manthey:
When trying to generate a thumbnail of the SVS files TCGA-A8-A06U-01A-01-TS1.63824040-373f-4c6c-a74e-881c127567a6.svs , an error occurs, which stops the thumbnail generation job.

This SVS file has only one level, and, when we attempt to read some parts of the image, the OpenSlide library throws an error.

At the very least, we should make the thumbnail job log and proceed after certain types of errors, skipping the problem files.

manthey:
Fixed in PR #118.

PR #118, opened 2016-11-21
If the thumbnail job fails to create a thumbnail, indicate it in the logs and keep going.

manthey:
Fixes issue #117.


manthey:
@dgutman I'd like to run this branch on neuro.  It will show which files we have trouble with, and we'll end up with thumbnails for most of the other files.  I'd feel better if someone (@zachmullen?) would look it over first.

zachmullen:
LGTM

PR #119, opened 2016-11-23
Be more flexible in the order of tiff directories.

manthey:
Some tif files have multiple images before the tiled data.

Guard against excessively large svs regions.

### DIFF
--- a/server/tilesource/tiff.py
+++ b/server/tilesource/tiff.py
@@ -56,6 +56,8 @@ class TiffFileTileSource(FileTileSource):
         largeImagePath = self._getLargeImagePath()
         lastException = None
 
+        # Query all know directories in the tif file.  Only keep track of
+        # directories that contain tiled images.
         alldir = []
         for directoryNum in itertools.count():
             try:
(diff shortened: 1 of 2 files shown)


manthey:
@law12019 This PR allows large_image to read the file you had this morning.

manthey:
@zachmullen or @law12019 This is ready for review.

The sample file I have is 1.6 Gb, so I don't want to add it to the testing suite.

server/tilesource/tiff.py
```diff
@@ -56,49 +56,46 @@ def __init__(self, item, **kwargs):
...
-            directories[level] = td
-
-        if not len(directories):
+            alldir.append((td.tileWidth * td.tileHeight, level, td))
+        if not len(alldir):
             msg = 'File %s didn\'t meet requirements for tile source: %s' % (
                 largeImagePath, lastException)
             logger.debug(msg)
             raise TileSourceException(msg)
+        alldir.sort()
```
zachmullen: A comment about what this loop is doing, and all the negative indexing, would be helpful for future developers.

PR #120, opened 2016-12-02
Better error handling on SVS images and on jobs

manthey:
On bad svs images, throw an exception on trying to open them rather then when tiles are read. This helps prevent out-of-memory issues.

Stop running a thumbnail job on errors.


manthey:
@zachmullen These are some small changes to be more robust.  I have some bad SVS files, the smallest of which is ~10 Mb.  If you think it is worth it, I can add that as an explicit test.

PR #121, opened 2016-12-02
Lazily load images from the tile iterator.

manthey:
When a tile is returned by the tile iterator, don't load the actual image until it is needed.

### DIFF
--- a/server/tilesource/base.py
+++ b/server/tilesource/base.py
@@ -91,6 +91,73 @@ class TileSourceAssetstoreException(TileSourceException):
     pass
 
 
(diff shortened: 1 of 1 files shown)


manthey:
@cdeepakroy This should work without any changes to your code at all.  I still need to write a test that shows that tiles are lazily loaded.

manthey:
I realize that unless you have asked for the tile in PIL format, that it probably gets loaded before being returned any way.  I'll deal with that next.

cdeepakroy:
LGTM!

PR #122, opened 2016-12-05
Add tests for lazy tile loading.

manthey:
Refactored lazy tile loading so that it is lazy even when the target format is not PIL or the source image type.


manthey:
@cdeepakroy In the previous PR, if you had asked for a tile iterator returning a format other than PIL or the source image format, it still was loading the tile image as part of the iterator.  This now defers the conversion to a numpy array or other format, too. (and there is a test for the deferred load).

cdeepakroy:
Great! LGTM!

PR #123, opened 2016-12-06
TIF properties and greyscale files

manthey:
This fixes a minor bug introduced in In PR #119 where not all tif properties were read as expected.

Support for greyscale files has been added.

Don't collide file names when generating a tif file.


manthey:
@purg Perhaps you should be the one to review this PR, as it involves issues that you discovered.

server/models/image_item.py
```diff
@@ -90,6 +91,11 @@ def _createLargeImageJob(self, item, fileObj, user, token):
             user=user)
         jobToken = Job.createJobToken(job)
 
+        outputName = os.path.splitext(fileObj['name'])[0] + '.tiff'
+        if outputName == fileObj['name']:
+            outputName = (os.path.splitext(fileObj['name'])[0] + '.' +
```
Purg: We could possibly just include the timestamp all the time as that would be generally informative as to when the conversion occurred (is that represented elsewhere in File model metadata?).

manthey: Yes, Girder already has the time that the file was added to the system stored in the file metadata.

PR #125, opened 2016-12-13
Better unicode handling.

manthey:
If a path contained unicode, libtiff would fail to parse it.  If a girder worker job contained unicode, it would fail to exec.

More tests are probably necessary to exhaustively test this.


plugin_tests/tiles_test.py
```diff
@@ -276,6 +276,29 @@ def testTilesFromGreyscale(self):
         self.assertEqual(tileMetadata['mm_y'], None)
         self._testTilesZXY(itemId, tileMetadata)
 
+    def testTilesWithUnicodeName(self):
+        # Unicode file names shouldn't cause problems.
+        file = self._uploadFile(os.path.join(
+            os.path.dirname(__file__), 'test_files', 'yb10kx5k.png'))
+        # Our normal testing method doesn't pass through the unicode name
+        # proeprly, so just change it after upload.
```
Purg: proeprly -> properly

PR #127, opened 2016-12-19
Fix a JPEG vips issue

manthey:
JFIF headers are now conditional.

See if an iOS is requesting a tile.  If so and we don't have a JFIF header, convert the JPEG to a PIL image and then to a new JPEG rather than adding a JFIF header.

It turns out that the colorspace of a JPEG is complicated, and adding a JFIF header changes it in unexpected ways.  See the comments.  vips (especially latest stable) often generates RGB jpegs (not YCbCr).  If you add a JFIF header to this, the colorspace is implied to be YCbCr, which is wrong.  Therefore, if we want to add a JFIF header, we *must* convert the colorspace, too.

Increase output from the create_tif script to make it easier to debug.   This print the actual command used and always prints non-empty stdout and stderr.

### DIFF
--- a/plugin_tests/common.py
+++ b/plugin_tests/common.py
@@ -29,9 +29,9 @@ from tests import base
 from girder.constants import SortDir
 
 
-JFIFHeader = '\xff\xd8\xff\xe0\x00\x10JFIF'
-JPEGHeader = '\xff\xd8\xff'
-PNGHeader = '\x89PNG'
+JFIFHeader = b'\xff\xd8\xff\xe0\x00\x10JFIF'
+JPEGHeader = b'\xff\xd8\xff'
+PNGHeader = b'\x89PNG'
 
 
 class LargeImageCommonTest(base.TestCase):
--- a/plugin_tests/common.py
+++ b/plugin_tests/common.py
@@ -29,6 +29,7 @@ from tests import base
 from girder.constants import SortDir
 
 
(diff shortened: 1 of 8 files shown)


manthey:
@zachmullen Could you review this PR when you get a chance?

server/create_tiff.py
```diff
@@ -30,6 +30,8 @@
 
 convert_command = (
     'vips',
+    # '--vips-concurrency', '1',
+    # '--vips-progress',
```
zachmullen: Did you mean to leave these commented lines here?

plugin_tests/common.py
```diff
@@ -29,6 +29,7 @@
 from girder.constants import SortDir
 
 
+JFIFHeader = '\xff\xd8\xff\xe0\x00\x10JFIF'
```
zachmullen: Should this be `b\xff\xd8\xff\xe0\x00\x10JFIF`? (Same for the others below?) That would prevent certain undesired behavior between python2 and 3 when doing comparisons.

server/create_tiff.py
manthey: I did, as optionally enabling them has helped with tracking down vips issues on certain installations.  I should add a comment to that effect.

plugin_tests/common.py
manthey: Yes, it should.  Thanks.

Issue #126, opened 2016-12-15
Add a tile size parameter for TileSource.tileIterator


cdeepakroy:
This is to provide the ability to use a custom tile size when iterating through a whole slide image at a desired magnification.

Example usage:

```python
ts = large_image.getTileSource(wsi_path)
for tile in ts.tileIterator(scale={'magnification': 20}, 
                            tile_size=(tile_height, tile_width)):
    pass
```

manthey:
Further, this could support overlap:
```
tile_size={
  'width': width.
  'height': height,
  'overlap_x': overlap_x,
  'overlap_y': overlap_y
}
```
where `overlap_x` and `overlap_y` would be non-negative integers.  The overlap would be applied equally on the opposite sides.  Or, would overlap be better to be customizable in all four directions?

cdeepakroy:
I think equally on both sides would be good. I would suggest using a separate argument for overlap

```python
ts = large_image.getTileSource(wsi_path)
for tile in ts.tileIterator(scale={'magnification': 20}, 
                            tile_size={'width': width, 'height': height),
                            tile_overlap={'x': None, 'y': None}):
    pass
```

cooperlab:
Equal on both sides is good. I also like splitting out the arguments.

manthey:
Great.  Will do.

manthey:
So if we specify a tile size and overlap, is the size *excluding* the overlap or *including* the overlap?

cdeepakroy:
The size would exclude the overlap. Visualize a sliding window moved across the image from left-to-right and top-to-bottom -- 
* `tile_size` specifies the size of the window and 
* `tile_overlap` specifies the amount of overlap of adjacent windows (along x and y dirs). 

It would be nice if you could also make getSingleTile receive the same arguments -- `tile_size` and `tile_overlap`.

manthey:
One virtue of everything going through `_tileIteratorInfo` internally is that `getSingleTile` will get these options without additional work.

manthey:
@cdeepakroy I've got this working on my machine and am adding tests.

[...]

PR #129, opened 2016-12-20
Add options to retile an image.

manthey:
The tile iterator can now take tile_size and tile_overlap parameters.  If tile size is not the original tile size, the image is retiled, which involves loading individual tiles and compositing them together as needed to produce the new time.  tile_size can take a width, height, or both.  tile_overlap can take x and y parameters which will retile the image to include that much overlap in each direction.  Overlap never extends outside of the requested region.

Resolves issue #126.


manthey:
@cdeepakroy When you get a chance, can you review this and ask for changes as needed?  Thanks.

server/tilesource/base.py
```diff
@@ -1349,6 +1370,16 @@ def tileIterator(self, format=(TILE_FORMAT_PIL, ), resample=False,
                 only the stride of the tiles.  This is a dictionary containing:
             x: the overlap on the left and right in pixels.
             y: the overlap on the top and bottom in pixels.
+            edges: if True, then the edge tiles will exclude the overlap
+                distance.  If unset or False, the edge tiles are full size.
+                    As an example, suppose an image that is 8 pixels across
+                (01234567) and a tile size of 5 is requested with an overlap of
+                2.  If the edges option is False (the default), the following
+                tiles are returned: 01234, 12345, 23456, 34567.  Each tile
```
cdeepakroy: This is confusing. 01234, 12345, 23456, 34567 have an overlap of 4 elements.

Is the documentation wrong, or is that how it works?

manthey: The overlap is per side.  An overlap of 2 means it overlaps two pixels on the left and 2 pixels on the right, which totals 4 pixels.

cdeepakroy: CC: @cooperlab

cdeepakroy: When we think of overlap 2, i would expect neighboring tiles to overlap by or have that amount of pixels in common. In this case, i would expect -- (01, 01234, 34567, 67) or (01234, 34567)

cooperlab: I agree with @cdeepakroy here - specifying an overlap of 10 means that any two tiles would overlap by 10 pixels instead of 5. That's the way I think most image analysis people would conceptualize overlap.

manthey: Should we then reject non-even values with an exception?  Currently, the overlap is per-side, not total.  If it is total and an odd value, it wouldn't make sense.

cdeepakroy: you can still think of it as per-side. For example for a 1x10 image (0123456789) with a tile size of 4 and  odd overlap of 1 ... we would get (0123, 3456, 6789) or (0, 0123, 3456, 6789, 9) depending on how you want to handle edges. In both of these results, the overlap of each tile with the tile on left or right is equal to 1. 

This is how most computer vision or image analysis

manthey: I've been thinking about it from the mapping perspective, where an overlap is typically symmetric on each side, and tile is centered on a particular region.  I was also thinking that from an image analysis perspective, you'd have an overlap because the algorithm can't be run too close to an edge, so you'd need _n_ pixels extra on each side so as to process the core of the tile and output that core.

To be as general as possible and clear, allowing all four sides to be specified might be clearer.

cooperlab: Enabling users to specify a padding on each side will make the interface more difficult to understand, with 99% of use cases being uniform padding on all edges. Extra control is good but increases the complexity of interfaces. Just my personal preference.

manthey:
@cdeepakroy, @cooperlab  Overlap now is the total overlap, not the per-side overlap.

cdeepakroy:
LGTM!

PR #136, opened 2017-01-31
Remove unnecessary package.json file

jbeezley:
As noted in https://github.com/girder/girder/pull/1766#issuecomment-276383510, this file is no longer necessary and can cause module resolution problems.


zachmullen:
Test failure is a result of latest Girder changing its output format for error messages slightly

jbeezley:
Since the original issue has been fixed, I'll go ahead and close this.

PR #141, opened 2017-02-15
Add endpoints for associated images.

manthey:
Available associated images can be listed or fetched.  When getting an image, a maximum width and/or height may be specified, along with an output format (JPEG or PNG).  Getting an image requires ADMIN access to the item, as the image may contain PHI.


manthey:
@dgutman This adds some endpoints to get the macro and label images (and any other images that might be available).  You need to have admin-control over the item to get the images.  We could add a view of all associated images in the item view for users with appropriate authorization (though I'd rather that be in another PR).

server/rest/tiles.py
```diff
@@ -429,3 +433,56 @@ def getTilesRegion(self, item, params):
...
+        try:
+            result = self.imageItemModel.getAssociatedImage(item, image, **params)
+        except TileGeneralException as e:
+            raise RestException(e.message, code=400)
+        if not isinstance(result, tuple):
+            return result
+        imageData, imageMime = result
+        setResponseHeader('Content-Type', imageMime)
+        setRawResponse()
+        return imageData
```
jbeezley: Currently, this doesn't work on the swagger page.  I think setting the `content-type` header might fix it.

PR #143, opened 2017-02-17
Add guards to prevent duplicate CSS elements from being added to the DOM

brianhelba:


web_client/views/imageViewerWidget/leaflet.js
```diff
@@ -1,11 +1,19 @@
...
 import ImageViewerWidget from './base';
 
 var LeafletImageViewerWidget = ImageViewerWidget.extend({
     initialize: function (settings) {
         ImageViewerWidget.prototype.initialize.call(this, settings);
 
-        $('head').prepend(
-            $('<link rel="stylesheet" href="https://cdnjs.cloudflare.com/ajax/libs/leaflet/0.7.7/leaflet.css">'));
+        var cssLoaded = _.some(
+            document.styleSheets,
```
brianhelba: Is webpack typically cool with referencing `document` like this, without any import?

jbeezley: I don't see any problem with it.

web_client/views/imageViewerWidget/leaflet.js
```diff
@@ -4,8 +4,15 @@ var LeafletImageViewerWidget = ImageViewerWidget.extend({
     initialize: function (settings) {
         ImageViewerWidget.prototype.initialize.call(this, settings);
 
-        $('head').prepend(
-            $('<link rel="stylesheet" href="https://cdnjs.cloudflare.com/ajax/libs/leaflet/0.7.7/leaflet.css">'));
+        if (!$('#large_image-leaflet-css', 'head').length) {
```
manthey: Purely out of curiosity, is there a reason you prefer the jquery selector of `$('#large_image-leaflet-css', 'head')` over `$('head #large_image-leaflet-css')`?

brianhelba: Nope. That's an artifact of me thinking about doing the lookup of `<head>` only once, but then removing it. I'll simplify it further now.

PR #146, opened 2017-02-24
Trigger g:imageRendered on every render

jbeezley:
I moved the `g:imageRendered` event into the `render` method of each class.  This is necessary for applications to be able to bind to events on the viewer object itself.  This is used by this commit on HistomicsTK https://github.com/DigitalSlideArchive/HistomicsTK/pull/235/commits/23dee917f43fb98abb97f7e10c12d996c9164098.


jbeezley:
I seem to have lost write access on this repo.

PR #150, opened 2017-03-02
Add options to retile an image.

manthey:
Copy from version 1.7 to the version 2 master.


manthey:
Merging since the Girder 1.7 version was approved as PR #129.

PR #151, opened 2017-03-03
Fix requiring girder as a dependency, backported to the girder-1.7 branch

manthey:


cdeepakroy:
LGTM! I just tested it and I confirm that installing large_image without girder works.

PR #152, opened 2017-03-04
Add code to display annotations as geojson

jbeezley:
This adds an annotation model and collection.  The model contains a method to convert from our own annotation spec to geojson.  As noted in the code, this can't be done for all annotation element types.  For the moment, it **only** handles rectangle types to support the nuclei segmentation end-to-end demo.

A cleaner solution might be to do the conversion to geojson on the server, but ultimately, we will need to write custom code to render annotations directly as given.  This is just a stop gap solution to render basic types using existing functionality in geojs.

### DIFF
--- a/plugin.json
+++ b/plugin.json
@@ -9,7 +9,8 @@
     "npm": {
         "dependencies": {
             "geojs": "0.11.0",
-            "slideatlas-viewer": "4.0.3"
+            "slideatlas-viewer": "4.0.3",
+            "tinycolor2": "^1.4.1"
         }
     }
 }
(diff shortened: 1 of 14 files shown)


brianhelba:
@jbeezley Without commenting on the substance of this PR, I'd like to point out my proposal in https://github.com/girder/large_image/issues/155 . However, it probably makes more sense to still merge this first (assuming it's acceptable), before refactoring.

jbeezley:
I was considering doing some test cases for this, but it's another case of code that may be temporary.  As I mentioned before, I think doing the conversion on the server makes more sense going forward.

PR #154, opened 2017-03-06
Make sure tile ranges are integers.

manthey:
If tile_position is given as a float, the x ranges could also be floats, which doesn't work with range iterators.


manthey:
This is a forward-port of the xmax-cast-as-int branch (PR #153).

PR #156, opened 2017-03-10
Make sure tile width and height are integers.

manthey:
Reduce memcached warnings.


manthey:
@cdeepakroy There was a condition where we could still get float values for a width and height.  This guards against that and works in my testing.

Your latest HistomicsTK branch with can use large_image master (it doesn't need to use the Girder 1.7 branch), since it doesn't use Girder 1.7.

cdeepakroy:
LGTM!

Issue #145, opened 2017-02-21
Plugin fails to initialize when using an older memcached


jbeezley:
On travis with Ubuntu 14.04 with memcached 0.44, large_image fails to start with the following error:
```
ERROR: Failed to load plugin "large_image":
Traceback (most recent call last):
  File "girder/utility/plugin_utilities.py", line 87, in loadPlugins
    root, appconf, apiRoot = loadPlugin(plugin, root, appconf, apiRoot)
  File "girder/utility/plugin_utilities.py", line 212, in loadPlugin
    pluginLoadMethod(info)
  File "girder/utility/plugin_utilities.py", line 416, in wrapped
    return func(*arg, **kw)
  File "/home/travis/build/DigitalSlideArchive/digital_slide_archive/build/girder/plugins/large_image/server/base.py", line 198, in load
    from .rest import TilesItemResource, LargeImageResource, AnnotationResource
  File "/home/travis/build/DigitalSlideArchive/digital_slide_archive/build/girder/plugins/large_image/server/rest/__init__.py", line 20, in <module>
    from .tiles import TilesItemResource
  File "/home/travis/build/DigitalSlideArchive/digital_slide_archive/build/girder/plugins/large_image/server/rest/tiles.py", line 29, in <module>
    from ..models import TileGeneralException
  File "/home/travis/build/DigitalSlideArchive/digital_slide_archive/build/girder/plugins/large_image/server/models/__init__.py", line 23, in <module>
    from .image_item import ImageItem
  File "/home/travis/build/DigitalSlideArchive/digital_slide_archive/build/girder/plugins/large_image/server/models/image_item.py", line 33, in <module>
    from ..tilesource import AvailableTileSources, TileSourceException
  File "/home/travis/build/DigitalSlideArchive/digital_slide_archive/build/girder/plugins/large_image/server/tilesource/__init__.py", line 22, in <module>
    from .base import TileSource, getTileSourceFromDict, TileSourceException, \
  File "/home/travis/build/DigitalSlideArchive/digital_slide_archive/build/girder/plugins/large_image/server/tilesource/base.py", line 23, in <module>
    from ..cache_util import tileCache, tileLock, strhash, methodcache
  File "/home/travis/build/DigitalSlideArchive/digital_slide_archive/build/girder/plugins/large_image/server/cache_util/__init__.py", line 20, in <module>
    from .cache import LruCacheMetaclass, tileCache, tileLock, strhash, methodcache
[...]

PR #157, opened 2017-03-10
Check if memcached is available.

manthey:
If not, fall back gracefully.  This also fixes a memcached version mismatch.


manthey:
@cdeepakroy This quiets the output from memcached except for a line saying that it isn't available.

This also fixes issue #145.

cdeepakroy:
LGTM!

PR #159, opened 2017-03-16
Make rotation optional in the json annotation rectangle schema.

manthey:


manthey:
@cdeepakroy This is a fairly simple change to no longer require a rotation value on rectangles.  It also fixes the travis build as some of the Ubuntu packages have been updated which caused problems.

cdeepakroy:
LGTM!

PR #160, opened 2017-03-17
Compute the hashkey for tile caches ourselves.

manthey:
cachetools.hashkey changes between subsequent imports of cachetools.  This means that within a single run of the program we were caching, but between runs the cache was functionally useless.

Also, rename the internal variable for the memcached client to `_client`.  I found `_Cache__data` peculiar with the capitalization and underscores.


manthey:
@jbeezley Do you mind reviewing this?

PR #164, opened 2017-03-24
Fix an issue when not using girder.

manthey:


manthey:
@cdeepakroy I'm making a few changes to make sure the HistomicsTK cli docker work on all images that large_image works with.  This is a pretty simple change that is needed for some images.

PR #173, opened 2017-04-11
[WIP] Show label images

manthey:
Show label images in the item list.

This is configurable in the plugin.

This is currently very rough and could use css improvements.  It is expected that we'd also expose metadata in some manner, but that hasn't been done yet.


manthey:
@jbeezley, could you review this?  I feel it is still a work-in-progress, but there is a desire to have it merged in and then have PRs to change the behavior later.  To have complete tests, we'll need to expose some of the files which have associated images to the client tests, but I'll leave that for a separate PR.

web_client/views/itemList.js
```diff
@@ -24,20 +25,76 @@ wrap(ItemListWidget, 'render', function (render) {
...
+        container.append(elem);
+        /* We store the desired src attribute in deferred-src until we actually
+         * load the image. */
+        elem.append($('<img class="waiting"/>').attr(
+                'deferred-src', apiRoot + '/item/' +
+                item.id + '/tiles/thumbnail?width=160&height=100'));
+        var access = item.getAccessLevel();
+        var extra = extraInfo[access] || extraInfo[AccessType.READ] || {};
+
+        $.each(extra.images || [], function (imageName, idx) {
```
jbeezley: I think the arguments are reversed here.

jbeezley:
The item list looks a little odd with extra images enabled.  I think either we should remove all of the width's on those elements (so there is no padding) or make all of the item names line up.

test

web_client/views/itemList.js
```diff
@@ -24,20 +25,76 @@ wrap(ItemListWidget, 'render', function (render) {
...
+            elem.attr('extra-image', imageName);
+        });
+
+        $('.large_image_thumbnail', container).each(function () {
+            var elem = $(this);
+            /* Handle images loading or failing. */
+            $('img', elem).one('error', function () {
+                $('img', elem).addClass('failed-to-load');
+                $('img', elem).removeClass('loading waiting');
+                elem.addClass('failed-to-load');
```
jbeezley: How about adding a visual indicator (like a grey background) for this class?

jbeezley:
With the last change, I think it looks okay.  I'm okay with merging as is.

PR #174, opened 2017-04-11
Speed up annotation validation.

manthey:
Use the jsonschema.validate for large annotations is prohibitively slow.  However, since large annotations are largely similar, use that similarity to do some of the validation.  One a test of 338,695 rectangles this changed the validation time from 460 seconds to 5 seconds.

Add some tests for the annotation model.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -183,6 +183,8 @@ install:
   - cd $girder_path
   - pip install --no-cache-dir -U -r requirements.txt -r requirements-dev.txt -e .
   - pip install -r $main_path/requirements.txt -e .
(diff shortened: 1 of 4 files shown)


manthey:
@cdeepakroy I think this is necessary before we start serving region requests with annotations -- ingesting them was very slow.  It would be nice to have a large set of polylines to test the speed on, too.

cdeepakroy:
I will try to get you a large annotation file with nuclear polygons soon. I am optimizing one other function to speedup nuclei detection. After that, i will run it on a large slide.

PR #176, opened 2017-04-13
Change how the large_image config settings are accessed.

manthey:
This also delays when the cache is selected, so that the configuration can be altered after large_image is imported.  The average_color example demonstrates how this can be done.


cdeepakroy:
LGTM!

PR #177, opened 2017-04-13
Support drawing annotations with geojs

jbeezley:
This supports an upcoming PR in HistomicsTK to draw annotation on top of large_image viewers.  This adds a components:

1. New models and collections representing both "annotations" and "elements".  The annotation types are real models that can be sync'ed to the server, but elements are ephemeral objects for storing the individual primitive elements.
2. Modules in `web_client/annotations/geojs` that generate large_image annotations from native geojs annotation objects.
3. A couple of methods and events added to the viewer class itself to trigger "drawing mode".

The CI is current failing at `server_large_image.tiles`.  I assume the failure is also in master because I didn't make any relevant server side changes.

### DIFF
--- a/web_client/views/imageViewerWidget/geojs.js
+++ b/web_client/views/imageViewerWidget/geojs.js
@@ -181,6 +181,7 @@ var GeojsImageViewerWidget = ImageViewerWidget.extend({
                     }
 
                     layer.removeAllAnnotations();
+                    layer.geoOff(window.geo.event.annotation.state);
                     resolve(elements, annotations);
                 }
             );
--- a/web_client/views/imageViewerWidget/geojs.js
+++ b/web_client/views/imageViewerWidget/geojs.js
@@ -7,6 +7,9 @@ import events from 'girder/events';
 import ImageViewerWidget from './base';
 import convertAnnotation from '../../annotations/geojs/convert';
 
+/**
+ * Generate a new "random" element id (24 random 16 digits).
+ */
 function guid() {
     function s4() {
         return Math.floor((1 + Math.random()) * 0x10000)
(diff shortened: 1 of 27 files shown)


manthey:
The CI is failing because of a recent change to girder_worker and in travis we are sharing a python environment between girder and girder worker.  PR #176 contains a fix as well as its relevant code.

web_client/views/imageViewerWidget/geojs.js
```diff
@@ -52,47 +70,129 @@ var GeojsImageViewerWidget = ImageViewerWidget.extend({
...
+        layer.removeAllAnnotations();
+        return new Promise((resolve) => {
+            var element;
+
+            options = _.defaults(options || {}, {trigger: true});
+            layer.geoOn(
+                window.geo.event.annotation.state,
+                (evt) => {
+                    element = convertAnnotation(evt.annotation);
+                    if (!element.id) {
```
manthey: This shouldn't be necessary -- we assign ids to elements when they are added to the mongo database.

jbeezley: The reason this is done here is because annotation element models are generated for the draw widget before they are submitted to the server.  This "id" is really only used to uniquely identify the elements in the [DrawWidget](https://github.com/DigitalSlideArchive/HistomicsTK/pull/296/files#diff-94336ac7f0b9fbbea193e4bf777a6c91R98) as a "view model".  Because there aren't any endpoints for the annotation element model, it isn't useful for communication with the server (right now anyway).

I took the approach that drawn elements are temporary until submitted as a group into an annotation.  If we had endpoints such as `POST /annotation/:id/element` and `PUT /annotation/:id/element/:id` to set the individual elements, it would make more sense to make the `ElementModel` a server-backed backbone model and POST after creation to set the id.  I could do so now, but it would take some reorganization of how the annotation models are created in the UI.

manthey: That's fine, then.

jbeezley:
Yeah, the missing coverage is mostly in `imageViewerWidget/geojs.js`; particularly the drawing part.  I struggled a bit to figure out how to test much of that inside phantom.  I managed to do it in my integration tests in HistomicsTK using vgl mocking and `interactor.simulateEvent`.  I can do the same here in a future PR.

PR #178, opened 2017-04-14
Fix JPEG color-space issues in Safari

manthey:
The JPEG color-space issue in Mobile Safari now is also present in Safari.  We having nothing better than user-agent scraping to determine if we need to convert to a limited set of color spaces with JFIF headers.  Test against recent Safari and Chrome user agent strings.


manthey:
For the record, this issue was introduced into the non-mobile version of Safari 10.1, which came out in March.

manthey:
@jbeezley Would you mind reviewing this?  I have field confirmation that it fixes the problem.

PR #179, opened 2017-04-17
Fix reporting to codecov.

manthey:


manthey:
@jbeezley I think that this is working.

Issue #170, opened 2017-04-05
Query annotations by area and zoom level


manthey:
The GET annotation/{id} should either be extended or a new endpoint be created that will return annotations that are within a specified rectangle and some minimum size for a particular zoom level.

This will require a spatial index.  We will probably want to precompute extents of annotations.  Annotations smaller than some size (1 pixel, for instance) at a zoom level should be able to be excluded -- I think the 'minor radius' of the annotation is the relevant metric for this.  We will want any annotation that overlaps the specified region, and it will be better to err on giving some annotations that are not strictly necessary than on being too exclusive (for instance, a polyline that surrounds the region but doesn't enter it can be returned).

manthey:
A limit for the total number of annotations returned would be useful, too. 

Querying by a list of types of annotations may also be desirable.

manthey:
Simplistically, each annotation has a bounding box and a minimum size.  The minimum size can crudely be the minimum of the width and height of the bounding box.  It would be used as the threshold of whether to show the annotation at a particular zoom level.

Points, however, have zero-area bounding box, , so we may wish to artificially include them at some other resolution.  If we have a line (not a polygon), then pedantically, it has the same problem as a point, in that it has no breadth.  If we are using the bounding box for the size threshold, then this is only a problem if it is vertical or horizontal.  

I can see it to be useful to also include:
- limiting the number of annotations returned by either a fixed number or by a number relative to their complexity (so fewer complex polygons would be returned than points).  A complexity based threshold would ensure that the viewer always has enough resources to show what we send it, even if it has to only show some of the annotations because of it
- filling the limit request with elements outside of the strict bounding box so that (slow) panning doesn't have perceptible lag in showing annotations.  This could just be accomplished by asking for a larger region in the client (which would be easier).

I think we'll need to precompute the bounding boxes as we save the annotation elements, as it will otherwise be prohibitively slow to make our queries.

zachmullen:
I think minimum zoom level at which an annotation appears should be an out-of-band piece of metadata -- I could see it being useful to show very small annotations even at low magnification (perhaps rendered as something other than a rectangle), just so a user knows on which section they should zoom in.

manthey:
I was thinking that if each feature had a size-like parameter, then it can be used as an optional query, so that a client can choose what threshold should be used for that.

[...]

PR #180, opened 2017-04-19
Allow querying annotations by region.

manthey:
Annotation elements can be restricted by location, details, size, or the standard paging controls.

### DIFF
--- a/server/rest/annotation.py
+++ b/server/rest/annotation.py
@@ -23,7 +23,7 @@ from girder import logger
 from girder.api import access
 from girder.api.describe import describeRoute, Description
 from girder.api.rest import Resource, loadmodel, filtermodel, RestException
-from girder.constants import AccessType
+from girder.constants import AccessType, SortDir
 from girder.models.model_base import ValidationException
 from ..models.annotation import AnnotationSchema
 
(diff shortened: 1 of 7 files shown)


manthey:
Closes issue #170.

kotfic:
@manthey if i'm reading this right,  it looks a lot like vector tiling? Are we sure there aren't any open source projects that we could pull modules out of that could potentially deliver more functionality with less maintenance overhead?

manthey:
I've looked, and the only vector tiling I've seen that uses Mongo as a back-end requires using the Mongo `2dsphere` index and the associated `$near` operator.  Unfortunately, Mongo's 2dsphere index *only* works with latitudes and longitudes (coordinates must be in the range of ([-180, 180], [-90, 90])).  The older 2d index where one could have arbitrary coordinates is deprecated (as of Mongo 2.6).  This means that we have to either do wacky things with stored coordinates, switch databases, or roll our own tiling of the vector features.

web_client/models/AnnotationModel.js
```diff
@@ -51,6 +110,57 @@ export default Model.extend({
     },
 
     /**
+     * Set the view.  If we are paging elements, possibly refetch the elements.
+     *
+     * @param {object} bounds: the corners of the visible region.  This is an
+     *      object with left, top, right, bottom in pixels.
+     * @param {number} zoom: the zoom factor.
+     * @param {number} maxZoom: the maximum zoom factor.
+     * @param {function} drawFunc: if present, call this function if a new
```
jbeezley: I'm not a big fan of callbacks, could we return a promise?

web_client/views/imageViewerWidget/geojs.js
```diff
@@ -93,6 +93,19 @@ var GeojsImageViewerWidget = ImageViewerWidget.extend({
                 features: ['point', 'line', 'polygon']
             });
             this._layers[annotation.id] = layer;
+
+            var setBounds = _.bind(function () {
+                var zoom = this.viewer.zoom(),
+                    bounds = this.viewer.bounds(),
+                    zoomRange = this.viewer.zoomRange();
+                if (annotation.setView) {
+                    annotation.setView(bounds, zoom, zoomRange.max, _.bind(function () { this.drawAnnotation(annotation); }, this));
```
jbeezley: I think this could be simplified by rerendering to the `g:fetched` event instead.

server/rest/annotation.py
```diff
@@ -92,14 +92,41 @@ def getAnnotationSchema(self, params):
...
+               required=False, dataType='float')
+        .param('bottom', 'The bottom row (exclusive) of the area to fetch.',
+               required=False, dataType='float')
+        .param('low', 'The lowest z value of the area to fetch.',
+               required=False, dataType='float')
+        .param('high', 'The highest z value (exclusive) of the area to fetch.',
+               required=False, dataType='float')
+        .param('minimumSize', 'Only annotations larger than or equal to this '
+               'size in pixels will be returned.  Size is determine by the '
+               'minimum orthogonal extent of the annotation.  This probably '
```
manthey: I changed what I was using for the size but failed to update this text.  Originally I was using the minimum orthogonal extent, but thin elements were undervalued.

web_client/models/AnnotationModel.js
```diff
@@ -39,6 +49,55 @@ export default Model.extend({
     },
 
     /**
+     * Fetch a single resource from the server. Triggers g:fetched on success,
+     * or g:error on error.
+     * To ignore the default error handler, pass
+     *     ignoreError: true
+     * in your opts object.
+     */
+    fetch: function (opts) {
```
manthey: Sadly, the super method doesn't pass through the data parameter.  Instead, it constructs a new set of options to pass to the rest call.

manthey: I could submit a PR to Girder to fix that, which would allow us to use the super here...

manthey:
Although this is working as is (and I think I addressed all of @jbeezley's comments), we should probably have a visual indicator when not all elements are visible (or even requery to have a lower overlap when this occurs).

web_client/views/imageViewerWidget/geojs.js
```diff
@@ -94,17 +94,20 @@ var GeojsImageViewerWidget = ImageViewerWidget.extend({
...
                 if (annotation.setView) {
-                    annotation.setView(bounds, zoom, zoomRange.max, _.bind(function () { this.drawAnnotation(annotation); }, this));
+                    annotation.setView(bounds, zoom, zoomRange.max);
                 }
-            }, this);
+            };
             if (layer.geoOn) {
                 layer.geoOn(window.geo.event.pan, setBounds);
             }
+            annotation.off('g:fetched').on('g:fetched', () => {
```
jbeezley: That's too general a condition, I would do `annotation.off('g:fetched', null, this)`.  You should also unbind the event in `removeAnnotation`.

PR #183, opened 2017-04-26
Add mouse events to geojs drawn annotations

jbeezley:
This propagates geojs mouse events to the large_image viewer widget as custom events
* `g:mouseOverAnnotation`
* `g:mouseOutAnnotation`
* `g:mouseClickAnnotation`

I also added the source annotation model id to the `ElementModel` to disambiguate between different annotations for listeners of the events.

This is missing tests because it coupling with geojs is too strong to permit reasonable mocking.  We need to add infrastructure for integration tests including mocking webgl to get them to run in phantomjs.  I think we should look into doing this in a separate PR.

### DIFF
--- a/web_client/views/imageViewerWidget/geojs.js
+++ b/web_client/views/imageViewerWidget/geojs.js
@@ -127,6 +127,12 @@ var GeojsImageViewerWidget = ImageViewerWidget.extend({
                 }
 
                 annotation.off('g:fetched', null, this).on('g:fetched', () => {
+                    // Trigger anevent indicating to the listener that
+                    // mouseover states should reset.
+                    this.trigger(
+                        'g:mouseResetAnnotation',
+                        annotation
+                    );
                     this.drawAnnotation(annotation);
                 }, this);
                 setBounds();
--- a/web_client/views/imageViewerWidget/geojs.js
+++ b/web_client/views/imageViewerWidget/geojs.js
@@ -206,15 +206,30 @@ var GeojsImageViewerWidget = ImageViewerWidget.extend({
     },
 
     _onMouseOverFeature: function (evt) {
-        this.trigger('g:mouseOverAnnotation', evt.data.properties.element);
+        var properties = evt.data.properties || {};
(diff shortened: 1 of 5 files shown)


web_client/views/imageViewerWidget/geojs.js
```diff
@@ -95,7 +97,16 @@ var GeojsImageViewerWidget = ImageViewerWidget.extend({
...
         }
         window.geo.createFileReader('jsonReader', {layer})
-            .read(geojson, () => this.viewer.draw());
+            .read(geojson, (features) => {
+                _.each(features || [], (feature) => {
+                    feature.selectionAPI(this._hoverEvents);
+
+                    feature.geoOn(geo.event.feature.mouseover, (evt) => this._onMouseOverFeature(evt));
+                    feature.geoOn(geo.event.feature.mouseout, (evt) => this._onMouseOutFeature(evt));
+                    feature.geoOn(geo.event.feature.mouseclick, (evt) => this._onMouseClickFeature(evt));
```
manthey: Would it be better to capture `geo.event.feature.mouseon` and `geo.event.feature.mouseoff`, which get fired when the topmost hovered element changes?  We could capture all five `mouse*` events, as they could all be useful at some point.

Further, since the functions all do the same thing except for the event name, perhaps there should only be one function like `(evt) => this._onMouseFeature('g:mouseOverAnnotation', evt)`, where the function would trigger the specified event.

manthey:
With paged annotations, when an annotation is removed that is being hovered over, no mouseout event gets triggered.  We either need to trigger such an event or the end user needs to listen for the `g:fetched` event and refresh hover information then.

jbeezley:
I made an attempt at fixing the issue with paging by adding a reset event.  My HistomicsTK implementation just resets the collection it uses to generate the popover when this is triggered.  It seems to function well enough.  At least it doesn't result in the popover content getting stuck.

I'll add a couple of tests for the new events now that we have the other testing branch in.

PR #189, opened 2017-05-09
Guard against LRU cache issues.

manthey:
cachetools LRU cache is not thread safe.  If two threads each cause an eviction, it would throw a KeyError and never recover.  Our cache would then fail, functionally breaking large_image until the server was restarted.

This does two things: (1) use a threading lock for the LRU cache, just as we do for memcached.  This should prevent the problem from ever occuring, and (2) guard against such a KeyError, and, if it occurs, return the cache miss value in any case.  If the LRU cache is used in a multithreaded environment without a lock, this guards against the catastrophic failure.

Tests have been added to ensure that multithreaded use works both with and without a cache lock.

### DIFF
--- a/plugin_tests/cache_test.py
+++ b/plugin_tests/cache_test.py
@@ -19,6 +19,7 @@
 
 import os
 import six
+import threading
 
 from girder import config
 from tests import base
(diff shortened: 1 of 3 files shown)


manthey:
ping @brianhelba

manthey:
Yes, it does.

PR #193, opened 2017-07-05
Fix JPEG output for PILLOW 4.2.0.

manthey:
As of PILLOW 4.2.0, JPEG output fails when given an RGBA image.  Before
it saved just the RGB channels.  This works around the issue.


manthey:
I've based this PR on the `eslint-updates` branch, so it would be nice if someone could review both of these.

manthey:
@jbeezley Thanks.  Since this is based on the `eslint-updates` branch, would you prefer for me to merge this into that now or wait until it has been reviewed separately?

jbeezley:
Those changes look fine as well.  I think we can merge this into other branch and get a passing build to merge into master.

PR #194, opened 2017-07-12
Add support for TIFF output.

manthey:
For the endpoints that generate images (tiles, thumbnails, regions, associated images), we had supported JPEG and PNG.  This adds TIFF support (only small tiffs as encoded by PIL) and generalizes some of the encoding format testing.  TIFFs can have a compression set, which is one of PIL's string constants.


manthey:
This is necessary for https://github.com/DigitalSlideArchive/HistomicsTK/issues/352

PR #195, opened 2017-07-13
Handle more associated images.

manthey:
This extracts macro and label images from tiff directories (in addition to those stored in image descriptions).  It also uses the pixel resolution if it is sensible.  More files now have a known magnification.


manthey:
This is based on the `small-tif-output` branch since it allows for easier testing.

manthey:
@law12019 This has full tests and works on the sample images I have.

server/tilesource/tiff_reader.py
```diff
@@ -170,6 +173,10 @@ def _validate(self):
 
         :raises: ValidationTiffException
         """
+        if not self._mustBeTiled:
+            if self._tiffInfo.get('istiled'):
+                raise ValidationTiffException('Expected a non-tiled TIFF file')
+            return
```
jbeezley: This doesn't handle the `None` or `True` cases described in the docstring.

server/tilesource/tiff.py
```diff
@@ -58,6 +58,7 @@ def __init__(self, item, **kwargs):
 
         largeImagePath = self._getLargeImagePath()
         lastException = None
+        self._associatedImages = {}
```
jbeezley: It isn't obvious to me the distinction between associated and embedded images.  Could you add a comment describing how they are different?

server/tilesource/tiff_reader.py
```diff
@@ -248,6 +254,22 @@ def _loadMetadata(self):
         self._imageWidth = info.get('imagewidth')
         self._imageHeight = info.get('imagelength')
         self.parse_image_description(info.get('imagedescription', ''))
+        units = {2: 25.4, 3: 10}
```
jbeezley: Where do these numbers come from?

server/tilesource/tiff.py
```diff
@@ -66,6 +67,19 @@ def __init__(self, item, **kwargs):
...
+                    id = associated._tiffInfo.get(
+                        'imagedescription').strip().split(None, 1)[0].lower()
+                    # Only use this as an associated image if the parsed id is
+                    # a reasonable length, alphanumeric characters, and the
+                    # image isn't too large.
+                    if (id.isalnum() and len(id) > 3 and len(id) <= 20 and
+                            associated._pixelInfo['width'] <= 8192 and
+                            associated._pixelInfo['height'] <= 8192):
+                        self._associatedImages[id] = associated._tiffFile.read_image()
+                except Exception:
```
jbeezley: Is it possible to use a more specific exception?  Also, the complexity of this function is getting a little high.  How about breaking this block into a new function?

server/tilesource/tiff_reader.py
manthey: If `_mustBeTiled` is truthy, it won't take this branch, and will check later (line 219).  I'll fix the `None` case.

server/tilesource/tiff_reader.py
manthey: I'll add a comment (it is from the Tiff specification).

server/tilesource/tiff.py
manthey: I'll add a comment.  Associated images are images stored the "usual" way in a TIFF file that aren't tiled.  Embedded images are vendor-specific extensions where an image is put into a normally non-image tag and stored as metadata on a usual TIFF image.

PR #197, opened 2017-07-19
Handle pyramidal files that open with OpenSlide and have non-power-of-two-levels.

manthey:
There was a file in the field with a non power-of-two level that caused erroneous tiles to be served.  This fixes the problem and adds a synthetically retiled test image for testing.


server/tilesource/svs.py
```diff
@@ -183,6 +183,13 @@ def _getAvailableLevels(self, path):
                 self._openslide = openslide.OpenSlide(path)
         # sort highest resolution first.
         levels = [entry[-1] for entry in sorted(levels, reverse=True)]
+        # Discard levels that are not a power-of-two compared to the highest
+        # resolution level.
+        levels = [entry for entry in levels if
+                  abs(((math.log(float(levels[0]['width']) / entry['width']) /
+                        math.log(2) - 0.5) % 1) - 0.5) < 0.02 and
+                  abs(((math.log(float(levels[0]['height']) / entry['height']) /
+                        math.log(2) - 0.5) % 1) - 0.5) < 0.02]
```
jbeezley: This is hard to understand.  Can you break it into a for loop?

PR #198, opened 2017-07-20
Test the image viewer widget.

manthey:
This also fixes occasionally rendering viewers twice and works around a recent change in OpenSeadragon where it could not reliably be destroyed by using `delete window.OpenSeadragon`.

This updates the version of geojs to version 0.12.2, which eliminates some console warnings when drawing rectangles.

This now refreshes an item view in Girder when a image conversion job finishes.  Before, when asking an item to be converted to a large image, if an asynchronous process was needed, the user had to refresh the item view or navigate away and back to see the large image.  Now the view refreshes and shows the large image.


manthey:
@dgutman This fixes the issue with OpenSeadragon, and, more generally, slightly speeds up showing large images in the browser for all but the first time.

@jbeezley Could you review this when you get a chance?

web_client/views/imageViewerSelectWidget.js
```diff
@@ -80,7 +84,15 @@ var ImageViewerSelectWidget = View.extend({
...
         });
+        this.currentViewer.name = viewerName;
     }
 });
 
+wrap(ItemView, 'initialize', function (initialize) {
+    this.listenTo(eventStream, 'g:event.large_image.finished_image_item', () => {
+        this.model.fetch();
+    });
+    initialize.apply(this, Array.prototype.slice.call(arguments, 1));
```
jbeezley: `_.rest(arguments)`

web_client/views/imageViewerWidget/geojs.js
```diff
@@ -25,29 +25,32 @@ function guid() {
...
-                this.render();
-            }
-        );
+        $.ajax({  // like $.getScript, but allow caching
+            url: staticRoot + '/built/plugins/large_image/extra/geojs.js',
+            dataType: 'script',
+            cache: true
+        }).done(() => {
+            this.trigger('g:beforeFirstRender', this);
+            ImageViewerWidget.prototype.initialize.call(this, settings);
```
jbeezley: Making `initialize` call async breaks backbone conventions.  Is there a way to refactor this to keep the synchronous behavior?

PR #199, opened 2017-07-20
Support content disposition on some endpoints.

manthey:
On the region, associated image, and thumbnail endpoints, support requesting a content disposition.  The contentDisposition query parameter can be left blank for the existing behavior, or set to 'inline' or 'attachment'.  The filename is absed on the item name with the word 'region', 'thumbnail', or the associated image label appended to it and the appropriate suffix for the mimetype, except that cached thumbnails yield a generate '_largeImageThumbnail' filename.  The filename is encoded in a safe ASCII manner and a UTF-8 manner such that it should work on most clients.

This is based on the small-tif-output branch to leverage some of the testing infrastructure there.

### DIFF
--- a/server/models/image_item.py
+++ b/server/models/image_item.py
@@ -282,7 +282,10 @@ class ImageItem(Item):
         })
         if existing:
             return self.model('file').download(
-                existing, contentDisposition='inline')
+                existing,
+                contentDisposition='inline'
+                if kwargs.get('contentDisposition') != 'attachment'
+                else kwargs['contentDisposition'])
         tileSource = self._loadTileSource(item, **kwargs)
         thumbData, thumbMime = tileSource.getThumbnail(
             width, height, **kwargs)
--- a/server/models/image_item.py
+++ b/server/models/image_item.py
@@ -281,11 +281,12 @@ class ImageItem(Item):
             'thumbnailKey': key
         })
         if existing:
+            if kwargs.get('contentDisposition') != 'attachment':
+                contentDisposition = 'inline'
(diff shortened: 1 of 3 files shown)


server/models/image_item.py
```diff
@@ -282,7 +282,10 @@ def getThumbnail(self, item, width=None, height=None, **kwargs):
         })
         if existing:
             return self.model('file').download(
-                existing, contentDisposition='inline')
+                existing,
+                contentDisposition='inline'
+                if kwargs.get('contentDisposition') != 'attachment'
+                else kwargs['contentDisposition'])
```
jbeezley: This is a little hard to read.  Could you use a variable instead?

server/rest/tiles.py
```diff
@@ -186,6 +195,36 @@ def _getTilesInfo(self, item, imageArgs):
...
+        If requested, set the content disposition and a suggested file name.
+
+        :param item: an item that includes a name.
+        :param contentDisposition: either 'inline' or 'attachemnt', otherwise
+            no header is added.
+        :param mime: the mimetype of the output image.  Used for the filename
+            suffix.
+        :param subname: a subname to append to the item name.
+        :param params: a dictionary of parameters.  Some of these can be used
+            to suggest a filename.
```
jbeezley: `params` doesn't seem to be used anywhere.

manthey: Thanks.  I had thought I was going to add width, height, etc. to the filename, but decided it was clearer not to do so.

PR #201, opened 2017-08-09
Python 3 support

manthey:
Make changes necessary to be Python 3 compatible and test on Python 3.

Simplistically, changes are to (a) how some exceptions are handled, (b) to some utility scripts that may not be used, (c) to the test tile source, (d) to some tests.  Prior to this, the plugin was manually tested with Python 3 where it worked in common use cases.  Since this adds CI testing, it is more certain to stay working.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -187,7 +187,8 @@ before_install:
 
   - pip install --no-cache-dir numpy==1.11.3  # needed because libtiff doesn't install correctly without it.  This ensures we have the same version for libtiff as for the project.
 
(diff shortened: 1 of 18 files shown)


jbeezley:
Could you make a note in the README (or maybe also in the `requirements.txt`) about the special version of pylibtiff necessary for python 3?

manthey:
I've changed `requirements.txt` (and `setup.py`).  There isn't any harm in using the github version of pylibtiff, so I've made that the default with a comment in `requirements.txt` of when it isn't necessary.

PR #202, opened 2017-08-23
Update tests to handle changes in job cancellation.

manthey:
Testing fails until this is fixed.  A number of other PRs have this change so that they can pass testing.

The `jobs.cancel` is event is consumed by the remote worker plugin, so we can't rely on it here.  Since we only use it in testing, it is easier to change the tests than to change the order that events are processed in Girder.

### DIFF
--- a/plugin_tests/large_image_test.py
+++ b/plugin_tests/large_image_test.py
@@ -294,10 +294,18 @@ class LargeImageLargeImageTest(common.LargeImageCommonTest):
         self.assertEqual(results['removed'], 1)
 
         def preventCancel(evt):
-            evt.preventDefault()
+            from girder.plugins.jobs.constants import JobStatus
+            from girder.plugins.worker import CustomJobStatus
+
+            job = evt.info['job']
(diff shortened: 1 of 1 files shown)


zachmullen:
Change seems fine, but I'm wondering if the behavior of eating the event in the worker is correct?

manthey:
This has to do with how Girder handles events.  If any process bound to an event calls `stopPropagation`, then the next handler won't get the event (of course).  Further, the test as I had originally written it was expecting the default event handler to fire, but remote worker now prevents the default and instead issues an update.

Since the test is just *preventing* the job from changing state, so that I can verify that certain conditions are handled properly, I am not concerned that remote worker changed what it was doing.

cjh1:
The use of stopPropagation() also means that plugins essentially compete for the event, the plugin loaded first wins. I think this is a feature of the Girder event system.

zachmullen:
Since it inspects the handler, I guess it does make sense for it to use stopPropagation.

PR #203, opened 2017-08-23
restRequest path option was deprecated, so it has to change to url.

manthey:


zachmullen:
I didn't realize `path` was deprecated, what's the rationale there? It was a girder restRequest-specific concept, why does it have to go away?

manthey:
It was deprecated it in Girder PR #2273.  Perhaps @brianhelba or @jbeezley can explain why they both thought it a good idea.

manthey:
With the deprecation and without this change, the javascript console is spammed with deprecation warnings (really, one would be enough).

zachmullen:
I'm still not convinced the deprecation makes sense, has anyone come to its defense?

PR #205, opened 2017-08-23
Always use the cache lock for the class cache.

manthey:
Without this, there is a still the possibility that two threads will attempt to evict the same cache entry, which results in the lru cache breaking.


zachmullen:
Would it be a good idea to make eviction on a nonexistent cache line OK/idempotent?

manthey:
We can do that for the `memcached` based cache, but not for the python `cachetools` LRU cache (at least not without monkey-patching).

manthey:
We could definitely subclass the LRUCache (and separately modify our use of memcached).  The locks added in this PR are still necessary, I think, as the cachetools module isn't thread-safe.

manthey:
@zachmullen I could make a separate PR with the idempotent eviction, unless you'd rather see it on this one.

zachmullen:
Separate is fine 👍

PR #213, opened 2017-09-28
Specify abstract requirements in setup.py, separating extras

brianhelba:


brianhelba:
@manthey Ready for review. PTAL.

brianhelba:
@manthey Updated: https://github.com/girder/large_image/blob/extras-requirements/README.rst

PR #217, opened 2017-10-10
Switch to 2.x-maintenance branch of girder for testing

jbeezley:


jbeezley:
The build error here is fixed by https://github.com/girder/girder/pull/2397

PR #229, opened 2017-11-07
Add new endpoint to copy an annotation from an item

Pierre-Assemat:
This endpoint will require : 
- `annotationId` : the ID of the annotation that has to be copied
- `itemDest`: the item to which will be copied the annotation

### DIFF
--- a/plugin_tests/annotations_test.py
+++ b/plugin_tests/annotations_test.py
@@ -25,6 +25,10 @@ from six.moves import range
 
 from girder import config
 from girder.constants import AccessType
+from girder.models.collection import Collection
+from girder.models.folder import Folder
+from girder.models.item import Item
+
 from tests import base
 
 from . import common
@@ -91,7 +95,7 @@ class LargeImageAnnotationTest(common.LargeImageCommonTest):
         self.assertIsNotNone(schema.annotationElementSchema)
 
     def testAnnotationCreate(self):
-        annotModel = self.model('annotation', 'large_image')
+        from girder.plugins.large_image.models.annotation import Annotation
         file = self._uploadFile(os.path.join(
             os.environ['LARGE_IMAGE_DATA'], 'sample_Easy1.png'))
         itemId = str(file['itemId'])
@@ -104,14 +108,15 @@ class LargeImageAnnotationTest(common.LargeImageCommonTest):
                 'height': 10,
             }]
         }
-        result = annotModel.createAnnotation({'_id': itemId}, self.admin, annotation)
+        result = Annotation().createAnnotation({'_id': itemId}, self.admin, annotation)
         self.assertIn('_id', result)
         annotId = result['_id']
-        result = annotModel.load(annotId)
+        result = Annotation().load(annotId)
         self.assertEqual(len(result['annotation']['elements']), 1)
 
     def testSimilarElementStructure(self):
-        ses = self.model('annotation', 'large_image')._similarElementStructure
+        from girder.plugins.large_image.models.annotation import Annotation
+        ses = Annotation()._similarElementStructure
         self.assertTrue(ses('a', 'a'))
         self.assertFalse(ses('a', 'b'))
         self.assertTrue(ses(10, 10))
@@ -150,11 +155,12 @@ class LargeImageAnnotationTest(common.LargeImageCommonTest):
         ]}))
 
     def testLoad(self):
-        annotModel = self.model('annotation', 'large_image')
+        from girder.plugins.large_image.models.annotation import Annotation
+
+        annotModel = Annotation()
         file = self._uploadFile(os.path.join(
             os.environ['LARGE_IMAGE_DATA'], 'sample_image.ptif'))
-        item = self.model('item').load(file['itemId'], level=AccessType.READ,
-                                       user=self.admin)
+        item = Item().load(file['itemId'], level=AccessType.READ, user=self.admin)
         with six.assertRaisesRegex(self, Exception, 'Invalid ObjectId'):
             annotModel.load('nosuchid')
         self.assertIsNone(annotModel.load('012345678901234567890123'))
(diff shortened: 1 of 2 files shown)


manthey:
For consistency with other parts of girder, I think the endpoint should be `POST /annotation/{id}/copy`

dgutman:
I thought we already had this endpoint?   I guess this just combines the
get and put into a single call?

server/rest/annotation.py
```diff
@@ -160,6 +161,33 @@ def createAnnotation(self, item, params):
...
+        .errorResponse('Validation Error: JSON doesn\'t follow schema.')
+    )
+    @access.user
+    @loadmodel(model='annotation', plugin='large_image')
+    @filtermodel(model='annotation', plugin='large_image')
+    def copyAnnotation(self, annotationId, itemDest, params):
+        user = self.getCurrentUser()
+        annotation = self.model('annotation', 'large_image').load(annotationId, region=params)
+        if annotation is not None:
+            itemSrc = self.model('item').load(annotation.get('itemId'), force=True)
```
jbeezley: The `@loadmodel` decorator will assert that the annotation exists.  The conditional here isn't necessary.

server/rest/annotation.py
```diff
@@ -160,6 +161,33 @@ def createAnnotation(self, item, params):
...
+        if itemSrc is not None:
+            self.model('item').requireAccess(
+                item, user=user, level=AccessType.READ)
```
jbeezley:
```python
itemSrc = self.model('item').load(annotation.get('itemId'), user=user, level=AccessType.READ)
```
The will load and validate permissions in one call.

manthey:
Yes -- currently you can GET annotation/{id} and then POST annotation; this would allow it to be done in a single call.

server/rest/annotation.py
```diff
@@ -160,6 +161,33 @@ def createAnnotation(self, item, params):
...
+        if itemDest is not None:
+            self.model('item').requireAccess(
+                item, user=user, level=AccessType.WRITE)
+
+        return self.model('annotation', 'large_image').createAnnotation(
+            itemDest, user, annotation)
```
jbeezley: This block won't work as is.  `itemDest` (or `itemId` when you rename it) is an ID string not a document.  What you want to do is more like this:
```python
itemDest = self.model('item').load(
    itemId, user=user, level=AccessType.WRITE, exc=True)

return self.model('annotation', 'large_image').createAnnotation(
    itemDest, user, annotation)
```

Also, you don't need to check if it is `None`.  The decorators above will assert that it exists.

server/rest/annotation.py
```diff
@@ -160,6 +161,33 @@ def createAnnotation(self, item, params):
                     exc.args, ))
 
     @describeRoute(
+        Description('Copy an annotation from an item to an other.')
+        .param('annotationId', 'The ID of the annotation.', paramType='path',
+               required=True)
+        .param('itemDest', 'Pass this to move the annotation to this item.',
```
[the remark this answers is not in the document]
Pierre-Assemat: I agree

server/rest/annotation.py
Pierre-Assemat: But when I think about it, do we really need to load the source item ? 
We will not use it, because we already load the annotation with its ID. 
Maybe to validate permission call. What do you think ?

server/rest/annotation.py
```diff
@@ -160,6 +161,30 @@ def createAnnotation(self, item, params):
...
+    @access.user
+    @loadmodel(model='annotation', plugin='large_image')
+    @filtermodel(model='annotation', plugin='large_image')
+    def copyAnnotation(self, annotationId, itemId, params):
+        user = self.getCurrentUser()
+        annotation = self.model('annotation', 'large_image').load(
+            annotationId,
+            level=AccessType.READ,
+            region=params
+        )
```
jbeezley: The annotation model is not access controlled, so the `level` argument here doesn't work.  You will need to validate read access to the source item, probably by loading it with the correct keyword arguments.

Pierre-Assemat: Ok I was not sure, thank you

server/rest/annotation.py
```diff
@@ -160,6 +162,32 @@ def createAnnotation(self, item, params):
...
+        itemId = params['itemId']
+        user = self.getCurrentUser()
+        Item().load(
+            annotation.get('itemId'),
+            user=user,
+            level=AccessType.READ
+        )
+        item = Item().load(itemId, user=user, level=AccessType.WRITE, exc=True)
+        self.model('annotation', 'large_image').createAnnotation(
+            item, user, annotation)
```
manthey: `annotation` is the Girder annotation resource.  `createAnnotation` takes the annotation schema object, which is `annotation['annotation']`.

server/rest/annotation.py
```diff
@@ -160,6 +162,32 @@ def createAnnotation(self, item, params):
...
+        return Item().save(item)
```
manthey: We aren't mutating `item` in any way, so saving it isn't necessary.  Further, the `@filtermodel` implies that we are returning the annotation, so I think we want
```
return self.model('annotation', 'large_image').createAnnotation(item, user, annotation['annoation'])
```

plugin_tests/annotations_test.py
```diff
@@ -439,6 +442,49 @@ def testGetAnnotation(self):
         self.assertLess(elements[0]['width'] * elements[0]['height'],
                         elements[-1]['width'] * elements[-1]['height'])
 
+    def testAnnotationCopy(self):
+        # create annotation on an item
+        annotModel = self.model('annotation', 'large_image')
+        file = self._uploadFile(os.path.join(
+            os.environ['LARGE_IMAGE_DATA'], 'sample_image.ptif'))
+        itemSrc = self.model('item').load(
```
manthey: This can be changed to `itemSrc = Item().load(`.  Since you are using the new model semantics, let's use them in all new and changed code.

Similarly, we can import the annotation model via `from girder.plugins.large_image.models.annotation import Annotation`, and then use `Annotation()` instead of `annotModel`.

plugin_tests/annotations_test.py
```diff
@@ -25,6 +25,11 @@
 
 from girder import config
 from girder.constants import AccessType
+from girder.models.collection import Collection
+from girder.models.folder import Folder
+from girder.models.item import Item
+from girder.plugins.large_image.models.annotation import Annotation
```
manthey: Within the plugin test, imports from plugins have to be done in the test functions, not at the top of the file -- the plugin isn't present when the plugin test is imported.  Move the `from girder.plugins.large_image.models.annotation import Annotation` line to be inside each function where `Annotation()` is invoked (I'd add it just below each `def` as appropriate).

Pierre-Assemat: Oh I didn't know that !! Thank you I will be able to fix the test and also other PR tests!!

Issue #223, opened 2017-10-19
Add utility function for converting point coordinates between magnifications


manthey:

manthey:
@slee172, @cdeepakroy What would the preferred API for this look like?  For instance,
```
point = (x, y)
scaledPoint = tilesource.getPointAtAnotherScale(point, sourceScale, sourceUnits, targetScale, targetUnits)
```
where `sourceScale`, `targetScale`, and `targetUnits` are identical to the parameters that can get passed to `convertRegionScale`.

As an example, if you have coordinates at magnification 5 and want them in the base units coordinates, this would be
```
scaledPoint = tilesource.getPointAtAnotherScale(point, sourceScale={'magnification': 5}, sourceUnits='mag_pixels', targetUnits='base_pixels')
```

slee172:
This looks good to me.

manthey:
Added in PR #230.

PR #230, opened 2017-11-13
Add a getPointAtAnotherScale method.

manthey:
This converts a coordinate tuple between magnifications or physical units.


manthey:
This addresses issue #223.

If you pass negative coordinates, these are interpreted as offsets from the right or bottom of the image.  If it is desired to *not* do that (to return negative coordinates in the scaled coordinate system), we'll need to add a flag to affect the behavior.

manthey:
@slee172, @cdeepakroy This is ready for use.

PR #233, opened 2017-12-01
Only import cherrypy if we can import girder.

manthey:


server/tilesource/pil.py
```diff
@@ -46,6 +45,8 @@
     from .. import constants
 except ImportError:
     girder = None
+if girder:
+    import cherrypy
```
zachmullen: We could save the `if girder` and just move `import cherrypy` to the bottom of the above `try` block.

manthey: My thought was that a cherrypy import error would still be reported this way, but that is probably wrong-headed, as girder has cherrypy as a hard dependency.

PR #234, opened 2017-12-05
Allow updating annotations without passing elements.

manthey:
For annotations with a large number of elements, it shouldn't be necessary to pass the elements just to change the annotation name or move it to another id.  This fixes the javascript model to allow updating the name of an annotation without requiring that all of the elements are loaded (also allowing changing the name of an annotation with paged elements).

Added tests for creating updating annotations.  Also, refactored the annotation tests to avoid uploading images in most cases; this speeds up the tests by 15 seconds or so on my local machine.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -194,7 +194,9 @@ before_install:
 
 
 install:
-  - pip install --no-cache-dir -U -r $girder_path/requirements-dev.txt -e $girder_path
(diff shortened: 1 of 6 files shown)


manthey:
@jbeezley Without this change, in HistomicsTK, if you edit an annotation that hasn't been loaded it will delete its elements.

PR #236, opened 2017-12-15
Add a method to the geojs viewer to highlight drawn annotations

jbeezley:
This is meant to be used for reactive applications to make a single annotation (or element) prominent on the display.  I don't know if "highlight" is the word, but I couldn't think of anything better.

There are performance issues when dealing with a large number of elements, so the opacity toggling behavior is disabled for large annotations.  I'm sure there are a number of performance enhancements we could make if it becomes an issue.

### DIFF
--- a/web_client/views/imageViewerWidget/geojs.js
+++ b/web_client/views/imageViewerWidget/geojs.js
@@ -26,6 +26,8 @@ function guid() {
 var GeojsImageViewerWidget = ImageViewerWidget.extend({
     initialize: function (settings) {
         this._annotations = {};
+        this._featureOpacity = {};
+        this._highlightFeatureSizeLimit = settings.highlightFeatureSizeLimit || 10000;
         this.listenTo(events, 's:widgetDrawRegion', this.drawRegion);
         this.listenTo(events, 'g:startDrawMode', this.startDrawMode);
         this._hoverEvents = settings.hoverEvents;
@@ -133,6 +135,7 @@ var GeojsImageViewerWidget = ImageViewerWidget.extend({
             this.setBounds({[annotation.id]: this._annotations[annotation.id]});
         }
         var featureList = this._annotations[annotation.id].features;
+        this._featureOpacity[annotation.id] = {};
         window.geo.createFileReader('jsonReader', {layer: this.featureLayer})
             .read(geojson, (features) => {
                 _.each(features || [], (feature) => {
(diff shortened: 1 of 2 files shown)


jbeezley:
I'm not sure what to make of this build error.  It only happens on one of the python 2.7 build environments.   https://travis-ci.org/girder/large_image/jobs/317032114#L2795

manthey:
Yes -- it is odd, as we have python-dev installed, so the header file *should* be there.  I'm looking into it.

jbeezley:
I wonder if it needs `python2.7-dev`.  I'm trying it in another branch.

jbeezley:
Nope.  Didn't work.

manthey:
I cleared the travis cache and the push build succeeded.  If the pr build also succeeds, that was probably all that was necessary.

web_client/views/imageViewerWidget/geojs.js
```diff
@@ -133,6 +135,7 @@ var GeojsImageViewerWidget = ImageViewerWidget.extend({
             this.setBounds({[annotation.id]: this._annotations[annotation.id]});
         }
         var featureList = this._annotations[annotation.id].features;
+        this._featureOpacity[annotation.id] = {};
```
manthey: Should we delete the entry in `_featureOpacity` when the annotation is removed in the `removeAnnotation` method?

web_client/views/imageViewerWidget/geojs.js
```diff
@@ -151,11 +154,89 @@ var GeojsImageViewerWidget = ImageViewerWidget.extend({
...
+    },
+
+    /**
+     * Use geojs's `updateStyleFromArray` to modify the opacities of all elements
+     * in a feature.  This method uses the private attributes `_highlightAnntotation`
+     * and `_highlightElement` to determine which element to modify.
+     */
+    _mutateFeaturePropertiesForHighlight: function (annotationId, features) {
+        _.each(features, (feature) => {
+            const data = this._featureOpacity[annotationId][feature.featureType];
```
manthey: Is there ever a case where this could be called before `drawAnnotation`?  If so, there should be a guard to ensure that `this._featureOpacity[annotationId]` exists.

jbeezley: No, not directly.  From the call [here](https://github.com/girder/large_image/blob/5273ae00bab6ab0cf84896dbef87893ae95a6b93/web_client/views/imageViewerWidget/geojs.js#L195), it is only called if `this._annotations[id]` exists which is created synchronously with the `this._featureOpacity[id]` object in `drawAnnotation`.

PR #237, opened 2017-12-18
Mapnik source

dorukozturk:


server/tilesource/geospatial.py
```diff
@@ -0,0 +1,128 @@
...
+        self.sizeX = self.dataset.RasterXSize
+        self.sizeY = self.dataset.RasterYSize
+        self.tileWidth = 256
+        self.tileHeight = 256
+        # # Fair assumption
+        self.levels = 20
+
+        # Earth circumference for web mercator
+        self.circumference = 20037508.342789244
+
```
manthey: Can you add a subclass method for `getMetadata` that returns the information the base tilesource has, plus 
- `geospatial`: a boolean set to True so consumers of this tile source know that it is geospatial.  Eventually this can be used to choose a better default viewer.
- `bounds`: the geospatial bounds as a dictionary which contains the max/min x/y coordinates and the coordinate system that they are in.

It might also be nice (if known) if the maximum resolution of the underlying image is known, so that the appropriate maximum `levels` could be more intelligently guessed at.

server/tilesource/geospatial.py
```diff
@@ -0,0 +1,128 @@
...
+    from girder import logger
+    from .base import GirderTileSource
+except ImportError:
+    girder = None
+    import logging as logger
+    logger.getLogger().setLevel(logger.INFO)
+
+
+class MapnikTileSource(FileTileSource):
+    name = 'mapnik'
```
manthey: In keeping with other tile sources, change the non-girder name to `'mapnikfile'`.

Should the entire tile source be called `mapnik` rather than `geospatial`?

Also, add `cacheName = 'tilesource'`, and wrap the class with `@six.add_metaclass(LruCacheMetaclass)` so that we cache the tile source.  This will avoid calling `gdal.Open` for each tile.

server/tilesource/geospatial.py
```diff
@@ -0,0 +1,128 @@
...
+
+        self.sizeX = self.dataset.RasterXSize
+        self.sizeY = self.dataset.RasterYSize
+        self.tileWidth = 256
+        self.tileHeight = 256
+        # # Fair assumption
+        self.levels = 20
+
+        # Earth circumference for web mercator
+        self.circumference = 20037508.342789244
```
manthey: Isn't this the half circumference?

server/tilesource/geospatial.py
```diff
@@ -0,0 +1,128 @@
...
+        proj.ImportFromWkt(wkt)
+
+        return proj.ExportToProj4()
+
+    def getPixelSize(self, zoom):
+        """Returns an approximate pixel size for a given zoom level"""
+
+        equatorLatitude = 0
+        # For simplification we assume everything is equator latitude
+        pixelSize = self.circumference * math.cos(equatorLatitude) / 2 ** (zoom + 7)
```
manthey: To be clear about why the division occurs, this should probably be `pixelSize = self.circumference * math.cos(equatorLatitude) / self.pixelWidth`.

server/tilesource/geospatial.py
```diff
@@ -0,0 +1,128 @@
...
+
+    @staticmethod
+    def getMercatorProjection():
+        """Returns proj4 string for mercator projection"""
+        srs = osr.SpatialReference()
+        srs.ImportFromEPSG(3857)
+
+        return srs.ExportToProj4()
+
+    def getTile(self, x, y, z, **kwargs):
```
manthey: Wrap with `@methodcache` to allow tile caching.  By default, this uses all of `kwargs` as part of the cache hashing, so this will do the right thing if you process style information.

web_client/views/configView.js
```diff
@@ -115,7 +115,11 @@ var ConfigView = View.extend({
             name: 'slideatlas',
             label: 'SlideAtlas',
             type: 'slideatlas'
-        }
+        }, {
+            name: 'geospatial',
```
manthey: Can you name this so that it is obvious that it is an openlayers based geospatial viewer?  Eventually, a geojs geospatial viewer would also be nice.

web_client/views/imageViewerWidget/geospatial.js
```diff
@@ -0,0 +1,83 @@
+import ImageViewerWidget from './base';
```
manthey: Can you change this file to `import OpenlayersImageViewerWidget from './openlayers.js', and then extend that (which should allow you to remove all but the `render` method)?

server/tilesource/geospatial.py
dorukozturk: Yes, you are right. I am fixing this.

server/tilesource/geospatial.py
dorukozturk: http://wiki.openstreetmap.org/wiki/Zoom_levels I used the formula from here. Does that make sense?

manthey: I mistyped my replacement, which should have been `pixelSize = self.circumference * math.cos(equatorLatitude) / 2 ** zoom / self.pixelWidth`.  The `+ 8` is just because the tile width happens to be `2 ** 8`, but, there is no intrinsic requirement that the tile size be a power of two, so making it explicit where this factor comes from is clearer.

server/tilesource/mapniksource.py
```diff
@@ -0,0 +1,195 @@
...
+        ymin = geotransform[3] + (self.dataset.RasterYSize * geotransform[5])
+        xmin = geotransform[0]
+        ymax = geotransform[3]
+        return xmin, ymin, xmax, ymax
+
+    def getMetadata(self):
+        geospatial = False
+        if self.dataset.GetProjection():
+            geospatial = True
+        xmin, xmax, ymin, ymax = self.getBounds()
```
wphicks: These values are being unpacked out of order. From [above](https://github.com/girder/large_image/pull/237/files#diff-8ab06242ffbb85f106cb32aab64e64e4R106), this should be `xmin, ymin, xmax, ymax`.

PR #240, opened 2018-01-03
Change to the new model singleton reference semantics.

manthey:
Also, use the canonical exception import locations.

This PR shouldn't change the functional code in any way.

### DIFF
--- a/server/base.py
+++ b/server/base.py
@@ -22,10 +22,17 @@ import json
 
 from girder import events, plugin, logger
 from girder.constants import AccessType, SettingDefault
-from girder.models.model_base import ModelImporter, ValidationException
+from girder.exceptions import ValidationException
+from girder.models.file import File
+from girder.models.item import Item
+from girder.models.notification import Notification
+from girder.models.setting import Setting
 from girder.utility import setting_utilities
+from girder.plugins.jobs.models.job import Job
 
 from . import constants
+from .models.annotation import Annotation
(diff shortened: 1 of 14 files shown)


server/base.py
```diff
@@ -22,10 +22,17 @@
...
 from girder import events, plugin, logger
 from girder.constants import AccessType, SettingDefault
-from girder.models.model_base import ModelImporter, ValidationException
+from girder.exceptions import ValidationException
+from girder.models.file import File
+from girder.models.item import Item
+from girder.models.notification import Notification
+from girder.models.setting import Setting
 from girder.utility import setting_utilities
+from girder.plugins.jobs.models.job import Job
```
jbeezley: Should this import be put inside the function that uses it as is done with `JobStatus`?

manthey: Yes.  Thanks for catching it.

server/base.py
```diff
@@ -71,9 +71,10 @@ def _updateJob(event):
     Called when a job is saved, updated, or removed.  If this is a large image
     job and it is ended, clean up after it.
     """
-    global JobStatus
+    global JobStatus, Job
     if not JobStatus:
         from girder.plugins.jobs.constants import JobStatus
+        from girder.plugins.jobs.models.job import Job
```
jbeezley: I don't think this conditional is doing anything useful.  I think you can just remove the `Job` and `JobStatus` globals.

PR #246, opened 2018-01-17
Better handle PIL 16 bit modes.

manthey:
This better parses PIL modes to handle more 16 bit files.

### DIFF
--- a/server/tilesource/pil.py
+++ b/server/tilesource/pil.py
@@ -113,7 +113,8 @@ class PILFileTileSource(FileTileSource):
         # to an 8-bit integer.  This expects the source value to either have a
         # maximum of 1, 2^8-1, 2^16-1, 2^24-1, or 2^32-1, and scales it to
         # [0, 255]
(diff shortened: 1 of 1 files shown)


server/tilesource/pil.py
```diff
@@ -113,7 +113,8 @@ def __init__(self, path, maxSize=None, **kwargs):
         # to an 8-bit integer.  This expects the source value to either have a
         # maximum of 1, 2^8-1, 2^16-1, 2^24-1, or 2^32-1, and scales it to
         # [0, 255]
-        if self._pilImage.mode in ('I', 'F') and numpy:
+        pilImageMode = self._pilImage.mode.split(';')[0]
```
jbeezley: Is it possible to have PIL and handle this operation with something like `self._pilImage.convert('L')`?

manthey: Sadly, no.  PIL just crops the values from 16 bit images to a range of [0,255], so almost all values end up being 255.

PR #248, opened 2018-01-18
Add access control lists to the annotation model

jbeezley:
This changes the super class of the annotation model from `Model` to `AccessControlledModel`.  It will allow us to use all of the access control functionality of Girder's permission model for annotations instead of coding ad hoc implementations as needed.  The main backwards incompatible change here is that the `load` method now does permission checking.

There is a schema migration that occurs on load.  It will generate an access control list identical to the folder containing the item the annotation is attached to plus it will give the annotation creator admin access.  I believe this will allow a smooth migration for annotations that currently exist (in HistomicsTK at least).

### DIFF
--- a/server/models/annotation.py
+++ b/server/models/annotation.py
@@ -419,6 +419,29 @@ class Annotation(AccessControlledModel):
         for annotation in annotations:
             Annotation().remove(annotation)
 
+    def _loadAndMigrateAnnotation(self, id, *args, **kwargs):
+        """
+        Load the annotation and add access control information if necessary.
+        """
+        # We first need to load the full annotation document to perform the
+        # migration.
+        annotation = self._migrateACL(super(Annotation, self).load(id, force=True))
+
+        # If force is not provided, we can do an access control check without
+        # reloading the model.
+        if annotation is not None and not kwargs.get('force'):
+            self.requireAccess(
+                annotation, kwargs.get('user'), kwargs.get('level', AccessType.ADMIN)
+            )
+
+        # If any other keyword arguments were provided, we reload the model to avoid
+        # having to duplicate code from super class methods.  Since these flags are
+        # rarely used the performance implications should be minimal.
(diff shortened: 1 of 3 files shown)


server/models/annotation.py
```diff
@@ -436,8 +491,21 @@ def load(self, id, region=None, getElements=True, *args, **kwargs):
             annotation.
         :returns: the matching annotation or none.
         """
+        force = kwargs.pop('force', False)
+        kwargs['force'] = True
```
manthey: I think this would be clearer if `force` was explicitly in the method parameter list, and the call to `annotation  = super(...).load()` included `force=True` instead of modifying `kwargs`.

server/models/annotation.py
```diff
@@ -436,8 +491,19 @@ def load(self, id, region=None, getElements=True, *args, **kwargs):
             annotation.
         :returns: the matching annotation or none.
         """
-        annotation = super(Annotation, self).load(id, *args, **kwargs)
-        if annotation is not None and getElements:
+        annotation = super(Annotation, self).load(id, force=True)
```
manthey: Is there any chance we want to pass through the `args` and `kwargs`, such as the `exc` or `fields` parameters?

jbeezley: It's possible... One issue is that the fields parameter could break the assumptions of the migration.  I was trying to avoid having to load the model twice, but maybe it's unavoidable if we want a truly transparent implementation here.

manthey: Perhaps we would load twice only if the `fields` parameter was specified?

A different approach would be to migrate all of the annotations when the model is instantiated, and then this can just use the superclass load.

jbeezley: I do like the idea of making the schema migration a more explicit operation either done on startup or via a command like `django-admin migrate`.  I think this is infrastructure that would be nice to add to girder core though.  Perhaps I'll bring it up as a discussion topic in our girder meeting.

Since this feature is needed sooner, I'll just go with the loading the model twice when necessary.

PR #250, opened 2018-01-23
Add annotation history.

manthey:
This exposes a setting on the plugin settings page to toggle whether history is saved.  It defaults to on.

If history is turned on, old versions of annotations are maintained in the annotation and annotationelement mongo collections.  There is a strictly monotonic version number for all annotations. 
 When an annotation is updated, the old elements are retained, and the old version of the annotation is saved with an _active=false marker.  These old versions have an _annotationId to reference the original annotation.

Historic versions can be listed via a new endpoint.  Historic versions can be fetched via the original annotation id and version number or via the historic record _id (all with appropriate user level authentication).  A revert endpoint allows switching to previous versions of an annotation.

### DIFF
--- a/server/models/annotation.py
+++ b/server/models/annotation.py
@@ -601,14 +601,10 @@ class Annotation(AccessControlledModel):
 
             with self._writeLock:
                 self.collection.delete_one = deleteElements
-                exc = None
                 try:
                     result = super(Annotation, self).remove(annotation, *args, **kwargs)
-                except Exception as e:
-                    exc = e
-                self.collection.delete_one = delete_one
-                if exc:
-                    raise exc
+                finally:
+                    self.collection.delete_one = delete_one
         return result
 
     def save(self, annotation, *args, **kwargs):
@@ -682,15 +678,11 @@ class Annotation(AccessControlledModel):
         with self._writeLock:
             self.collection.replace_one = replaceElements
             self.collection.insert_one = insertElements
-            exc = None
             try:
                 result = super(Annotation, self).save(annotation, *args, **kwargs)
-            except Exception as e:
-                exc = e
-            self.collection.replace_one = replace_one
-            self.collection.insert_one = insert_one
-            if exc:
-                raise exc
+            finally:
+                self.collection.replace_one = replace_one
+                self.collection.insert_one = insert_one
         if _elementQuery:
             result['_elementQuery'] = _elementQuery
         logger.debug('Saved annotation in %5.3fs' % (time.time() - starttime))
--- a/server/models/annotation.py
+++ b/server/models/annotation.py
@@ -23,13 +23,14 @@ import enum
 import jsonschema
 import six
 import re
+import threading
 import time
 from six.moves import range
 
 from girder import events
 from girder import logger
 from girder.constants import AccessType, SortDir
-from girder.exceptions import ValidationException, AccessException
+from girder.exceptions import ValidationException
 from girder.models.model_base import AccessControlledModel
 from girder.models.folder import Folder
 from girder.models.item import Item
@@ -398,6 +399,7 @@ class Annotation(AccessControlledModel):
     )
 
     def initialize(self):
+        self._writeLock = threading.Lock()
         self.name = 'annotation'
         self.ensureIndices([
             'itemId',
@@ -597,9 +599,16 @@ class Annotation(AccessControlledModel):
                 Annotationelement().removeElements(annotation)
                 return ret
 
-            self.collection.delete_one = deleteElements
-            result = super(Annotation, self).remove(annotation, *args, **kwargs)
-            self.collection.delete_one = delete_one
+            with self._writeLock:
+                self.collection.delete_one = deleteElements
+                exc = None
+                try:
+                    result = super(Annotation, self).remove(annotation, *args, **kwargs)
+                except Exception as e:
+                    exc = e
+                self.collection.delete_one = delete_one
+                if exc:
+                    raise exc
         return result
 
     def save(self, annotation, *args, **kwargs):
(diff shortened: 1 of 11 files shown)


server/models/annotation.py
```diff
@@ -555,16 +586,20 @@ def remove(self, annotation, *args, **kwargs):
...
-            return ret
+            def deleteElements(query, *args, **kwargs):
+                ret = delete_one(query, *args, **kwargs)
+                Annotationelement().removeElements(annotation)
+                return ret
 
-        self.collection.delete_one = deleteElements
-        result = super(Annotation, self).remove(annotation, *args, **kwargs)
-        self.collection.delete_one = delete_one
+            self.collection.delete_one = deleteElements
```
jbeezley: Overriding methods on the mongo collection in a rest request makes me very uncomfortable.  Doesn't this create some pretty ugly race conditions?

Given that this is done elsewhere in the Annotation model, maybe this PR isn't the correct place to solve it, but I would like to have us explore an alternative mechanism for dealing with the interaction between the Annotation and Element collections.

server/models/annotation.py
```diff
@@ -727,3 +776,109 @@ def validate(self, doc):
...
+
+        if annotationId and not isinstance(annotationId, ObjectId):
+            annotationId = ObjectId(annotationId)
+        # Don't apply limit and offset here, as they are subject to access
+        # control and other effects
+        entries = self.find(
+            {'$or': [{'_id': annotationId}, {'_annotationId': annotationId}]},
+            sort=sort)
+        isAdmin = force or (user is not None and user['admin'])
+        for entry in entries:
```
jbeezley: If I understand this loop correctly, it looks like the reason you aren't using the builtin `filterResultsByPermission` is because of the possibility of duplicate version numbers.  I'm not sure I fully understand the conditions under which this will happen, or why it would be more likely than inconsistencies resulting from overriding methods on `self.collection`.  Would it be possible to avoid the situation alltogether using locks or to use `$group` and `$first` queries to filter out duplicates?

server/models/annotation.py
```diff
@@ -727,3 +776,109 @@ def validate(self, doc):
...
+        :param user: the Girder user.  If the user is not an admin, they must
+            have read access on the item and the item must exist.
+        :param force: if True, don't get the user access.
+        """
+        if annotationId and not isinstance(annotationId, ObjectId):
+            annotationId = ObjectId(annotationId)
+        entry = self.findOne({
+            '$or': [{'_id': annotationId}, {'_annotationId': annotationId}],
+            '_version': int(version)
+        })
```
jbeezley: Would it be worth using `fields='_id'` to prevent loading the full document twice in this function?

server/models/annotation.py
```diff
@@ -727,3 +776,109 @@ def validate(self, doc):
...
+        annotation = Annotation().getVersion(id, version, user, force=force)
+        if annotation is None:
+            return
+        # If this is the most recent (active) annotation, don't do anything.
+        # Otherwise, revent it.
+        if not annotation.get('_active', True):
+            item = Item().load(annotation.get('itemId'), force=True)
+            if item is not None and not force:
+                Item().requireAccess(
+                    item, user=user, level=AccessType.WRITE)
```
jbeezley: The access control is a thorny issue here.  I don't really like relying on the item access control here and the annotation access control elsewhere.  Even though it isn't ideal, I think using the access list of the current version of the annotation would be more consistent.  Perhaps we should avoid considering the access list part of the versioning.

server/models/annotation.py
manthey: Mongo doesn't have the concept of a transaction lock, nor can I find any mechanism for placing a write lock on a collection.  As long as annotations and annotation elements are in two collections, we have to manually do something that is equivalent to a transaction.

When writing, we get a new version (which is guaranteed to be unique), save the new elements (which have a distinct version), then update the annotation with information that includes the new version.  When saving history, we copy the existing annotation record to a new record (which gets a new ID but keeps the old version.  It is possible if more than one process is updating an annotation, this copying can happen more than once.  There will only be one annotation record with the latest version, but there *could be* more than one copy of the annotation with the same old version number.

In reconsidering this, I do need to add a lock inside the annotation collection so that we can ensure that when we override the methods it is guarded.

I would love to find a better method for mongo transactions.

server/models/annotation.py
manthey: Yes.  I'll switch to using an aggregation by group and `filterResultsByPermission`

web_client/templates/largeImageConfig.pug
```diff
@@ -65,5 +65,16 @@ form#g-large-image-form(role="form")
       | Caching files speeds up thumbnail retrieval but takes some storage space.  Use 0 to not cache thumbnails as files.
     input.input-sm.form-control.g-large-image-max-thumbnail-files(
       type="text", value=settings['large_image.max_thumbnail_files'], placeholder="Use 0 to not cache thumbnails as files.")
+  .form-group
+    label Store annotation history
+    p.g-large-image-description
+      | Whenever annotations are saved, a record of the annotation can be stored in a separate database collection.
```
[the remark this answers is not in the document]
manthey: No -- I wrote this when I thought I was going to use a separate database.

server/models/annotation.py
manthey: It should be sufficient to just validate that we can write the annotation.  I'll change this to:
```
if not force:
    self.requireAccess(annotation, user=user, level=AccessType.WRITE)
```

server/models/annotation.py
jbeezley: I think we are hitting limitations of mongo here because we are trying to use it in a way that it isn't intended to be used.  Transactions are just not something we should try and implement.  Likely locks are the best immediate solution to this particular problem, but I wouldn't consider this a pattern to encourage elsewhere.  My hope is that in the future, we can do more to design our schemas in a way that these hacks aren't necessary.

server/models/annotation.py
```diff
@@ -555,16 +588,27 @@ def remove(self, annotation, *args, **kwargs):
...
+            with self._writeLock:
+                self.collection.delete_one = deleteElements
+                exc = None
+                try:
+                    result = super(Annotation, self).remove(annotation, *args, **kwargs)
+                except Exception as e:
+                    exc = e
+                self.collection.delete_one = delete_one
+                if exc:
+                    raise exc
```
jbeezley: Could you just use a finally block?

PR #251, opened 2018-01-24
Raster styling with palettable paletttes

dorukozturk:
Add styling functionality to single band rasters using palettable palettes


manthey:
Can you add tests for bad style json, and invalid min, max, and band values (not numbers)?

PR #252, opened 2018-01-31
Get more band information for mapnik gdal sources.

manthey:
This adds band information into the metadata.

### DIFF
--- a/server/tilesource/mapniksource.py
+++ b/server/tilesource/mapniksource.py
@@ -251,6 +251,50 @@ class MapnikTileSource(FileTileSource):
             self._bounds[srs] = bounds
         return self._bounds[srs]
 
+    def getBandInformation(self):
+        if not getattr(self, '_bandInfo', None):
+            infoSet = {}
+            for i in range(self.dataset.RasterCount):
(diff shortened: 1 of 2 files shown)


manthey:
@dorukozturk I ran this on the few geotiff files I had and it works for each.  We could add color palette information into the band information record, too, if that would be useful.

Locally, I tested it on a hyper-spectral image, an RGBA image, and a categorical image (palettized), and they all seem sensible.  If we need the computed statistics or the computed min/max, that should get either an optional parameter to fetch them or a new endpoint, as those could be slow to process.

server/tilesource/mapniksource.py
```diff
@@ -251,6 +251,50 @@ def getBounds(self, srs=None):
...
         return self._bounds[srs]
 
+    def getBandInformation(self):
+        if not getattr(self, '_bandInfo', None):
+            infoSet = {}
+            for i in range(self.dataset.RasterCount):
+                band = self.dataset.GetRasterBand(i + 1)
+                info = {}
+                stats = band.GetStatistics(True, True)
+                # The statistics provie a min and max, so we don't fetch those
```
dorukozturk: "provie" --> "provide"

PR #253, opened 2018-02-01
Add a getPixel method and rest endpoint.

manthey:
Internally, `getRegion` can now specify width and height in different units than the edges of a region.  This allows a 1-pixel square region to be specified with the left and top specified via non-pixel units. 
 `getPixel` gets a 1x1 pixel region and returns a tuple of the channel values as an object.  Subclasses of the tile source can augment this with additional information (such as band data).

For the mapnik source, this also returns band information at the pixel.

Also, the generated hash key for methods in cached classes now includes the class hash key.  Without this change, either the methodcache decorators need more parameters or a change in the source may use a cached tile (for instance, just changing the mapnik/gdal style).

### DIFF
--- a/plugin_tests/source_mapnik_test.py
+++ b/plugin_tests/source_mapnik_test.py
@@ -209,3 +209,75 @@ class LargeImageSourceMapnikTest(common.LargeImageCommonTest):
         image2 = self.getBody(resp, text=False)
         self.assertEqual(image2[:len(common.PNGHeader)], common.PNGHeader)
         self.assertNotEqual(image, image2)
+
+    def testPixel(self):
+        file = self._uploadFile(os.path.join(
+            os.path.dirname(__file__), 'test_files', 'rgb_geotiff.tiff'))
+        itemId = str(file['itemId'])
+
+        # Test in pixel coordinates
+        resp = self.request(
+            path='/item/%s/tiles/pixel' % itemId, user=self.admin,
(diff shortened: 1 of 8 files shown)


plugin_tests/source_mapnik_test.py
```diff
@@ -209,3 +209,75 @@ def testThumbnailFromGeotiffs(self):
...
+        resp = self.request(
+            path='/item/%s/tiles/pixel' % itemId, user=self.admin,
+            params={'left': 2120, 'top': 198})
+        self.assertStatusOk(resp)
+        self.assertEqual(resp.json, {})
+
+        # Test with a projection
+        resp = self.request(
+            path='/item/%s/tiles/pixel' % itemId, user=self.admin,
+            params={
```
dorukozturk: Can we add one more test with WGS 84 coordinates and see if we get the same results?
I did the conversion and it would be something like this:

```
params={
    'left': -117.97493778,
    'top': 33.8642302306,
    'projection': 'EPSG:4326',
    'units': 'projection'
}
```

dorukozturk: After checking this out second time I realized https://github.com/girder/large_image/blob/375c2c7e107ecc819d27313efd4bf6d000176c6f/server/tilesource/mapniksource.py#L97-L100. So I am happy.

PR #254, opened 2018-02-05
Mapnik python3

### LINKED ISSUE #243: Test mapnik source using Python 3 in CI

dorukozturk:
Fixes #243.
Add a docker-compose file to spin up the environment to run the mapnik source tests
against Python3.5


plugin_tests/source_mapnik_test.py
```diff
@@ -117,8 +128,18 @@ def testTileStyleFromGeotiffs(self):
...
         testImage = PIL.Image.open(testImage)
-        self.assertIsNone(PIL.ImageChops.difference(image, testImage).getbbox())
+        testImagePy3 = PIL.Image.open(testImagePy3)
+
+        diffs = [
+            PIL.ImageChops.difference(image, testImage).getbbox(),
+            PIL.ImageChops.difference(image, testImagePy3).getbbox()
+        ]
+
+        self.assertTrue(None in diffs)
```
manthey: This can be written as `self.assertIn(None, diffs)`

dorukozturk: Much better, I will change it.

server/tilesource/mapniksource.py
```diff
@@ -91,7 +91,11 @@ def __init__(self, path, projection=None, style=None, **kwargs):
             # -180 to +180 degrees horizontally.  We should accept this as a
             # parameter eventually.
             inProj = pyproj.Proj('+init=epsg:4326')
-            outProj = pyproj.Proj(self.projection)
+            if isinstance(self.projection, six.binary_type):
+                projection = self.projection.decode('utf8')
+            else:
+                projection = self.projection
```
manthey: We explicitly `encode` the projection to a binary string earlier, but here we are explicitly converting to a unicode string.  It looks like `pyproj.Proj` always handles unicode (in Python 2 and 3), but fails with bytes in Python 3.  So, safely, since we converted to bytes earlier, here we can safely say `outProj = pyproj.Proj(self.projection.decode('utf8'))`.

manthey: Though a comment might be warranted to say why that is safe...

server/tilesource/mapniksource.py
```diff
@@ -232,6 +236,8 @@ def getBounds(self, srs=None):
             }
             if srs and srs != nativeSrs:
                 inProj = pyproj.Proj(nativeSrs)
+                if isinstance(srs, six.binary_type):
+                    srs = srs.decode('utf8')
```
manthey: We cache the bounds based on `srs`, so this might confuse the issue to change it here.  Perhaps remove these lines and just change the following line to `outProj = pyproj.Proj(srs.decode('utf8') if isinstance(srs, six.binary_type) else srs)` (that way we don't alter `srs`).

manthey:
It doesn't harm anything, but do we want to spin up RabbitMQ as part of the docker compose step?  Also, we git clone girder_worker, but don't install or use it.  I think it would be better to do nothing with RabbitMQ or girder_worker OR to install and start girder_worker in a docker_compose service so it is available if we want it.

dorukozturk:
@manthey You are right. [This](https://github.com/girder/large_image/blob/master/plugin.json#L5) confused me. I fixed it.

PR #257, opened 2018-02-19
Allow units to specify any proj4 projection.

manthey:
The projection parameter is the projection of the tiles.  It can be useful to have a different projection for the units, so allow the units to include a few enumerated values ('wgs84', '4326') or any string that starts with 'epsg:' or 'proj4:' (all case insensitive).  In these cases, convert from the specified units to pixels or the specified tile projection.

This also changes `toNativePixelCoordinates` to round to the nearest coordinates by default.  This resolves some differences in pixel values that were showing in tests.

There is a change in computing the class cache hashing string.  Before `MapnikTileSource(file, projection)` got different results than `MapnikTileSource(file, projection=projection)`, whereas it should not.  The change affects the mapnik and pil tile sources.

### DIFF
--- a/plugin_tests/source_mapnik_test.py
+++ b/plugin_tests/source_mapnik_test.py
@@ -244,7 +244,7 @@ class LargeImageSourceMapnikTest(common.LargeImageCommonTest):
             })
         self.assertStatusOk(resp)
         self.assertEqual(resp.json, {
-            'r': 77, 'g': 82, 'b': 84, 'a': 255, 'bands': {'1': 74.0, '2': 80.0, '3': 84.0}})
+            'r': 77, 'g': 82, 'b': 84, 'a': 255, 'bands': {'1': 77.0, '2': 82.0, '3': 84.0}})
         # Check if the point is outside of the image
         resp = self.request(
             path='/item/%s/tiles/pixel' % itemId, user=self.admin,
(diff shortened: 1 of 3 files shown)


server/tilesource/mapniksource.py
```diff
@@ -409,14 +408,100 @@ def getTile(self, x, y, z, **kwargs):
...
+        :param proj: a binary or unicode projection string.
+        :returns: a proj4 projection object.  None if the specified projection
+            cannot be created.
+        """
+        if isinstance(proj, six.binary_type):
+            proj = proj.decode('utf8')
+        if not isinstance(proj, six.text_type):
+            return
+        if proj.lower().startswith('proj4:'):
+            proj = proj.split(':', 1)[1]
```
dorukozturk: Why do we have the "proj4:" prefix as an option? I think most geospatial software expects the proj4 string as is.

plugin_tests/source_mapnik_test.py
```diff
@@ -285,4 +285,111 @@ def testPixel(self):
...
+        self.assertStatusOk(resp)
+        self.assertEqual(resp.json, {
+            'r': 77, 'g': 82, 'b': 84, 'a': 255, 'bands': {'1': 77.0, '2': 82.0, '3': 84.0}})
+
+    def testSourceErrors(self):
+        from girder.plugins.large_image.tilesource import TileSourceException
+        from girder.plugins.large_image.tilesource.mapniksource import MapnikTileSource
+        filepath = os.path.join(
+            os.path.dirname(__file__), 'test_files', 'rgb_geotiff.tiff')
+        with six.assertRaisesRegex(self, TileSourceException, 'must not be geographic'):
```
dorukozturk: Why is this still getting raised?

manthey: Tiles have to be in a non-geographic projection.  Units can be in any projection, but the `projection` for a tile can't be, since it needs to map to a 2D conformal space.

dorukozturk: Oh ok, now I got it.

server/tilesource/mapniksource.py
manthey: You don't need the `proj4:` prefix if the string starts with `+proj=`.  Proj4 strings don't have to start in any consistent manner (they can start with `+proj=`, `+init=`, `+units=`, etc.).  By adding an extra optional prefix of `proj4:`, I can assume the rest is a proj4 string.

Instead, I could try to generate a `pyproj.Proj` from the string, and, if it throws an exception, assume that it is some other unit that is elsewhere handled.  This means that we would hide pyproj.Proj exceptions, since we wouldn't know if the error occurred because of an invalid proj4 string or a unit that was not otherwise handled.

dorukozturk: Ok, I think as long as we have the option to directly pass the proj4 string I am happy.

PR #258, opened 2018-02-19
Track girder cache changes.

manthey:
In Girder PR #2274, instead of using `getCurrentToken` to cache that cookies are being used, the cherrypy.request attr `girderAllowCookie` is used.  We cache loading models to avoid one or more database lookups per tile request, and this change broke that caching.

### DIFF
--- a/plugin_tests/common.py
+++ b/plugin_tests/common.py
@@ -63,6 +63,12 @@ class LargeImageCommonTest(base.TestCase):
                 self.publicFolder = folder
             if folder['name'] == 'Private':
                 self.privateFolder = folder
+        folders = Folder().childFolders(self.user, 'user', user=self.admin)
+        for folder in folders:
+            if folder['name'] == 'Public':
+                self.userPublicFolder = folder
+            if folder['name'] == 'Private':
+                self.userPrivateFolder = folder
         # Authorize our user for Girder Worker
         resp = self.request(
             '/system/setting', method='PUT', user=self.admin, params={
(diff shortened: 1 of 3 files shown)


manthey:
@opadron, @kotfic This fixes the 401 errors you were seeing.

plugin_tests/common.py
```diff
@@ -63,6 +63,12 @@ def setUp(self):
                 self.publicFolder = folder
             if folder['name'] == 'Private':
                 self.privateFolder = folder
+        folders = Folder().childFolders(self.user, 'user', user=self.admin)
+        for folder in folders:
+            if folder['name'] == 'Public':
+                self.userPublicFolder = folder
+            if folder['name'] == 'Private':
+                self.userPrivateFolder = folder
```
opadron: Let's say I make a new account on your girder instance, and decide that the default `Public` and `Private` folders are lame and I replace them with my own.  Have I just broken this part of the code?

manthey: This is the test where we create the folders in the test, so, no.

manthey:
Caching permissions generally across requests is a slippery business.  Here, the speed is worth the possible reduction in security -- if the security credentials have changed, we might continue serving tiles we shouldn't for the duration of the cache expiration time.  This isn't too big a risk -- individual tiles don't contain metadata or allow access to the full original item.  However, if this was done generally, then it is possible that an item that was accidentally made public and then is made private is exposed much longer than expected.  You either end up with short cache times which reduces the cache benefit, or long exposure times, or a long list of events that can invalidate the cache (for instance, we could listen for user, token, and group changes and invalidate the cache on all of those -- but I don't think Girder core emits model events on `update`, only on `save` and `remove`).

PR #260, opened 2018-02-21
Make annotated images search case insensitive

jbeezley:


manthey:
In the `re.split` call, can you add `flags=re.UNICODE`, please?  I have an image with annotations named `Образец Картина` (I know it's bad Russian), and if I search for `Картина`, it won't find it otherwise.

PR #265, opened 2018-03-08
Guard against bad WGS84 coordinates

dorukozturk:


server/tilesource/mapniksource.py
```diff
@@ -232,6 +232,11 @@ def getBounds(self, srs=None):
                 },
                 'srs': nativeSrs,
             }
+            # Make sure geographic coordinates do not exceed their limits
+            if pyproj.Proj(nativeSrs).is_latlong() and srs:
+                yBound = 85.0511 if '3857' in srs else 90.0
```
dorukozturk: @manthey Do you know a better way to check if srs is web mercator?
```python
if '3857' in srs
```
This is kind of ugly.

server/tilesource/mapniksource.py
```diff
@@ -232,6 +232,11 @@ def getBounds(self, srs=None):
                 },
                 'srs': nativeSrs,
             }
+            # Make sure geographic coordinates do not exceed their limits
+            if pyproj.Proj(nativeSrs).is_latlong() and srs:
+                yBound = 85.0511 if '3857' in str(srs) else 90.0
```
manthey: We could just test if 90.0 is valid for this projection:
```
  yBound = 89.999999 if pyproj.Proj(src)(0, 90)[1] == 1e30 else 90.0
```

The `Proj` object can be called with coordinates which will return `1e30` if they can't be converted.  If you pass `errcheck=True`, it would throw an exception, so another way would be:
```
  try:
    pyproj.Proj(src)(0, 90)
    yBound = 90.0
  except RuntimeException:
    yBound = 89.999999   
```

Otherwise, we have to know the maximum coordinates the projection can represent, which would involve probing the results of transforms for that projection.

PR #266, opened 2018-03-09
Add a unitsPerPixel option.

manthey:
If a stereographic projection is used, the method of calculating the level 0 size was flawed.  This prevents an exception from being thrown, but isn't particularly general.

Short of known the range and domain of each projection, there probably isn't a general solution.  When this proves to be insufficient, we'll need to allow passing in the valid range.

### DIFF
--- a/plugin_tests/source_mapnik_test.py
+++ b/plugin_tests/source_mapnik_test.py
@@ -338,6 +338,26 @@ class LargeImageSourceMapnikTest(common.LargeImageCommonTest):
         with six.assertRaisesRegex(self, TileSourceException, 'does not have a projected scale'):
             MapnikTileSource(filepath)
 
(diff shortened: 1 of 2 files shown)


manthey:
I have no particular use case for a stereographic projection at this time, so this PR is mainly because I noticed that they didn't work AND have a `unitsPerPixel` parameter allows for handling general tiles in most imaging libraries.  If we are only accessing tiles via getRegion (such as from a WMS query that specifies the resolution and bounds), then the `unitsPerPixel` is not particularly relevant -- instead of failing of there is no such value, we could have a default value, which would obviate the need to pass value in that case.

manthey:
This has no strict need, but I'd love to have it reviewed and merged.  @dorukozturk, can you take a look when you get a chance?

PR #267, opened 2018-03-09
Allow girder to show load errors in more cases.

manthey:
If girder is installed but requirements of large_image are not, the plugin would report as enabled but not be functional.  This checks if girder is available, and, if so, allows girder to report errors.

Eventually, it will be nice to separate large_image without girder from the necessary girder parts to make this easier to manage.  For now, this should make diagnosis easier.

As a simple test, when everything works under girder, uninstall a required module (cachetools, for instance), and restart girder.  Before, no error was reported.  With this change, the missing module is reported.

### DIFF
--- a/server/__init__.py
+++ b/server/__init__.py
@@ -18,8 +18,10 @@
 ###############################################################################
 
 try:
-    from .base import *  # noqa
-except ImportError as exc:
+    import girder
+    # This should include all of the explicitly required Girder plugins.
+    import girder.plugins.worker
+except ImportError:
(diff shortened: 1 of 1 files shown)


manthey:
This change will now report the error that @law12019 ran into instead of hiding it silently.

brianhelba:
Girder's pip-installable plugins change will also help to prevent this, right? Since the installation of `large_image` will force requirements to be pulled in.

manthey:
@brianhelba Yes.

PR #270, opened 2018-03-15
Packer based generation of girder + large_image docker image

kotfic:
This still requires more documentation,  but in principal is a minimum viable feature.

If you want to try it out install [Packer](https://www.packer.io/docs/install/index.html) and cd to ```devops/packer/``` and run ```packer build large_image.json```  this will require that ansible is installed in your environment.  It should build a docker image with girder & large_image with stock libtiff on an ubuntu-16.04 image.  Once build you should be able to test it out with:

```sh
docker run -it --net=host --rm large_image/large_image:latest
```

You will need a local instance of mongo running.  Once girder is up you will have to create a user,  assetstore,  enable the plugins and rebuild.


kotfic:
@brianhelba @danlamanna @manthey This is ready for review,  best place to start is the [README.md](https://github.com/girder/large_image/tree/example-docker/devops) PTAL!

manthey:
Do we want to be able support vagrant on Windows?  It seems like this would be desirable, but because this uses `ansible-galaxy` on the host machine, it doesn't work.  I think otherwise you need to switch provisioners from `ansible` to `ansible-local`.

kotfic:
Vagrant is still supported as it normally would be on windows.  Once you have the large_image.box file you can import that box,  and do a ```vagrant init <imported name>```.  Packer + Ansible is really just for building the box file (and eventually uploading it to some kind of registry like vagrant cloud). The assumption is that the build environment will be a gnu/linux based machine which is apart of the continuous delivery pipeline. 

We could switch to an ansible_local provisioner,  but that would blunt some of the utility of the provisioning scripts which we don't currently use,  but might like to use in the future. Eventually,  those ansible scripts should be run frequently, and in parallel, from a remote devops machine against deployed hosts.  This prevents configuration drift and dramatically reduces the effort to scale machines and/or recover from failure.

manthey:
Almost (all?) projects I've seen that have vagrant support basically allow you to do `vagrant up` without any other vagrant commands.  I recognize that if we public a vagrant box somewhere that is built via packer, we can then have a simple vagrant file that just is based on that box so that this works on Windows.  This necessitates having the published vagrant box.  Can we have a short-term plan to automate building such a box when master is updated?  Otherwise, avoiding compatibility for undefined future utility seems like a poor trade.

kotfic:
sorry this got put on the back-burner,  i'll resolve the conflict shortly. i'm happy to file/self-assign an issue to do CD as apart of another PR.  I assume we want a full build (e.g. mapnik & openslide tile sources?)

manthey:
Yes, we want all a full build with all of the tile sources.

kotfic:
@manthey  I've rebased on master to resolve conflicts and force pushed this branch. I've also opened https://github.com/girder/large_image/issues/285 and assigned to myself to integrate this work into CI/CD as a follow on PR

manthey:
Normally, running docker twice in a row does very little the second time because all of the build steps have been cached.  Should I expect this behavior with packer?  When I run `packer build -var "tile_sources=openslide,mapnik" docker.json` twice in a row, it reruns ansible scripts which take a while.

manthey:
I see some odd artifacts in a geotiff that is being rendered, possibly due to using GDAL version 1.  You can get GDAL version 2 via `add-apt-repository ppa:ubuntugis/ppa` and then installing the `gdal-bin` package.  I'm not sure what else is necessary.

kotfic:
@manthey happy to update to GDAL2,  may take a little while until i can get back around to this.    In terms of rebuilding,  the expectation is that you will have to run the ansible each time. Basically, the trade off for consistent builds across environments (e.g. docker, virtualbox, AWS, etc ) is that you can't use target environment specific features (docker layer caching in this instance). packer treats docker as "just another machine" and so all operations are performed in a single layer.  The upside is you will usually see smaller images

manthey:
Closing this.  Please reopen against the current code base if it is still desired.

PR #276, opened 2018-04-06
Return a list of element groups with annotation endpoints

jbeezley:
To support the use case of grouping annotations when listing them on the client, this maintains a denormalized list of "groups" on the annotation object inside the database.  This property is calculated on demand for annotations that existed before this change; otherwise, it is added/modified automatically on `annotation.save(...)`.

I would have preferred to keep this code entirely on the model layer, but the only place to inject the migration behavior for the listing endpoint is in the `find` method.  This however returns a mongo cursor rather than a normal iterator and would be difficult to override.

### DIFF
--- a/server/models/annotationelement.py
+++ b/server/models/annotationelement.py
@@ -20,6 +20,7 @@
 import datetime
 import math
 import time
+import six
 from six.moves import range
 
 from girder.constants import AccessType, SortDir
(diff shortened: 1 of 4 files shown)


manthey:
Should the existence of annotation elements that have no group value be recorded in the groups value (as a null, perhaps)?

jbeezley:
Sure, it makes some of the client logic that I wrote simpler.

manthey:
You'll need to merge or rebase on master for the circleci tests to pass.

server/models/annotationelement.py
```diff
@@ -296,3 +296,14 @@ def updateElements(self, annotation):
...
                 len(elements), time.time() - startTime))
+
+    def getElementGroupSet(self, annotation):
+        query = {
+            'annotationId': annotation.get('_annotationId', annotation['_id']),
+            '_version': annotation['_version']
+        }
+        groups = sorted(self.collection.distinct('element.group', filter=query))
+        query['element.group'] = None
+        if self.collection.find_one(query):
```
manthey: We should probably add an index to the annotationelement collection with `annotationId, _version, element.group`.

PR #282, opened 2018-04-25
Add an annotation list widget to the item page

jbeezley:
This adds a simple table displaying annotations attached to the current item.  The user can click on them to view them if using the GeoJS viewer (the only one capable of rendering annotations at the moment).  The delete and access buttons are only visible if the user has the necessary permissions.

test

### DIFF
--- a/web_client/templates/annotationListWidget.pug
+++ b/web_client/templates/annotationListWidget.pug
@@ -17,7 +17,7 @@ if annotations.length
         - var creator = creatorModel ? creatorModel.get('login') : annotation.get('creatorId');
         tr.g-annotation-row(data-annotation-id=annotation.id)
           td.g-annotation-toggle
-            input(type='checkbox', disabled=!canDraw, checked=drawn.has(annotation.id))
+            input(type='checkbox', disabled=!canDraw, checked=drawn.has(annotation.id), title='Show annotation')
           td.g-annotation-name(title=name)
             = name
 
--- a/web_client/templates/annotationListWidget.pug
+++ b/web_client/templates/annotationListWidget.pug
@@ -2,36 +2,39 @@
   i.icon-pencil
   | Annotations
 
-table.g-annotation-list.table.table-hover.table-condensed
-  thead
-    th.g-annotation-toggle
-    th.g-annotation-name Name
-    th.g-annotation-user Creator
-    th.g-annotation-date Date
-    th.g-annotation-actions
-  tbody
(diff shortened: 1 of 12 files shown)


aashish24:
A very useful and neat feature 👍

web_client/views/annotationListWidget.js
```diff
@@ -0,0 +1,127 @@
...
+        this.collection = this.collection || new AnnotationCollection();
+        this.users = new UserCollection();
+
+        this.listenTo(this.collection, 'all', this.render);
+        this.listenTo(this.users, 'all', this.render);
+
+        this.collection.fetch({
+            itemId: this.model.id,
+            sort: 'created',
+            sortdir: -1
```
manthey: There is something odd about the display order.  We request the annotations in reverse creation order, but they are displayed in a different order (I'm not sure what).  For instance, I get:
image
I was expecting those dates to be descending.

jbeezley: Is that the order the `/annotations` endpoint returns them in or is the client code changing it for some reason?

jbeezley:
Looks like the test I added is flaky... :disappointed:

web_client/views/annotationListWidget.js
msmolens: You might need to change the sorting attributes on the `AnnotationCollection`:
https://github.com/girder/girder/blob/bbab319395d3908f43ac9e0eefaa4fe880338f82/clients/web/src/collections/Collection.js#L17-L18

manthey: The endpoint returns them in the created date order.  The client is changing that (but not to name order, so I don't know how it is being changed).

jbeezley: Well, the default girder behavior is to sort by `model.get('name')`, which is undefined for these models.  If the comparator is stable, it should retain the original order.  I'll just try disabling backbone's sorting and see if that works.

manthey:
If you look at an large_image item without annotations, it shows one entry with no name and "invalid date", and shows a popup that a 400 error occurred.

web_client/templates/annotationListWidget.pug
```diff
@@ -0,0 +1,40 @@
+.g-annotation-list-header
+  i.icon-pencil
+  | Annotations
+
+if annotations.length
```
manthey: I think it would be better to not show the entire annotation list block when there aren't any annotations.

Issue #232, opened 2017-12-01
Refactor for easier development


manthey:
This is a general issue that can be removed once the main components of it have been accomplished, moved to their own issues, or removed.  As large_image is used in more places, more features are desired and it becomes harder to maintain.  To aid with future development, refactor the code so that:
- [x]  Make large_image (without girder) pip installable.  Use wheels for complex dependencies.  This should resolve issue #209.
- [x] Make individual tile sources pip installable.  When a tile source is needed, all installed tile source can be found and used.
- [x] Make the girder plugin pip installable (this will require the Girder 3 plugin refactor)
- [x] Move annotations so that they appear as a separate module (so that something can depend on large_image without pulling in annotation code).
- [x] Make annotations a separate Girder plugin.  This will resolve issue #155.
- [x] Improve devops on the image-conversion process (which uses vips) so it can be easily installed or run as a docker job
- [x] Ensure that we can pass complex parameters to tile and region endpoints to allow styling information to be passed through
- [x] Generalize unit/projection support in region requests (this would allow extension to WMS tile sources)

manthey:
Completed with #329.

PR #286, opened 2018-05-10
Use the Girder FUSE mount.

manthey:
The direct path will be used on the filesystem assetstore unless the Girder item has a file besides the large_item file and an originating file.

### DIFF
--- a/server/tilesource/base.py
+++ b/server/tilesource/base.py
@@ -25,11 +25,9 @@ from ..cache_util import getTileCache, strhash, methodcache
 try:
     import girder
     from girder import logger
-    from girder.exceptions import ValidationException
-    from girder.models.assetstore import Assetstore
+    from girder.exceptions import ValidationException, FilePathException
     from girder.models.file import File
     from girder.models.item import Item
-    from girder.utility import assetstore_utilities
     from ..models.base import TileGeneralException
     from girder.models.model_base import AccessType
 except ImportError:
@@ -1726,8 +1724,7 @@ class FileTileSource(TileSource):
 
 # Girder specific classes
 
-if girder:
-
+if girder:  # noqa - the whole class is allowed to exceed complexity rules
     class GirderTileSource(FileTileSource):
         girderSource = True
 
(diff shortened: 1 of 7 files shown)


server/tilesource/base.py
```diff
@@ -1726,8 +1724,7 @@ def canRead(cls, path, *args, **kwargs):
 
 # Girder specific classes
 
-if girder:
-
+if girder:  # noqa - the whole class is allowed to exceed complexity rules
```
jbeezley: This noqa disables all linting for the code block.  Can the logic in `_getLargeImagePath` be simplified to not exceed complexity rules?

manthey: The logic in `_getLargeImagePath` doesn't exceed complexity rules.  The linter is measuring the complexity of the total `if girder:` block, which is the total complexity of the `GirderTileSource` class and complaining about that.  I'm not sure how to get the linter to stop that short of moving the `GirderTileSource` to a secondary source file which, for conceptual reasons, I'd prefer not to do.

kotfic: It seems kind of confusing that depending on the runtime context this class is either defined,  or not defined.  The behavior i would expect is to see is a `RuntimeError` being raised either when importing or instantiating in a non-valid context.

manthey: Girder is optional.  Girder insists on having plugins have a server module.  Until we refactor Girder to allow pip installable plugins, there is a mashup of Girder and Girder-free code.  Somewhere in this mashup you need to have conditionally defined classes, as some can only be defined if Girder is present.  A conditional with an import would hide this more, but conceptually it is doing the same thing, and I think such a change would make it less obvious what is going on (but then I prefer fewer source files compared to what some others prefer).

kotfic: As implemented this will result in imports of `GirderTileSource` throwing `ImportError` depending on the run time context.  To me,  it makes more sense to throw a `RuntimeError` that explicitly notifies the developer what about the run time context is invalid.

manthey: I think this *should* be an `ImportError` and *not* a `RuntimeError`, as the failure in this condition is that you can't import girder (`RuntimeErrors` are for errors that don't fall into other categories, and this falls into a failure to import).  Admittedly, the error message would be better as `ImportError: no module named girder`, and not `ImportError: cannot import name GirderTileSource` (assumably from something like `from base import GirderTileSource`).

While this is a fine point to discuss, it doesn't actually have anything to do with the current PR, and will probably change (hopefully to @kotific's satisfaction) when issue #232 is resolved.

kotfic: Yeah,  not looking to bike shed this,  sorry to derail.

PR #297, opened 2018-08-10
Return annotations via yielding elements

manthey:
Instead of returning annotations as a monolithic response, yield from the database as the elements are read.  This reduces memory and speeds up the response.  For a large annotation set (~0.3 million polygons), this reduced output time by a factor of 3.

Use the ujson library for serializing annotation elements.  This takes 2/3 the time of the default json library for polygons.

Strip trailing .0 from floats that have integer values to reduce output size when it can be done trivially.  Javascript treats the results the same.

### DIFF
--- a/server/rest/annotation.py
+++ b/server/rest/annotation.py
@@ -24,12 +24,14 @@ import cherrypy
 from girder import logger
 from girder.api import access
 from girder.api.describe import describeRoute, autoDescribeRoute, Description
-from girder.api.rest import Resource, loadmodel, filtermodel
+from girder.api.rest import Resource, loadmodel, filtermodel, setResponseHeader
 from girder.constants import AccessType, SortDir
 from girder.exceptions import ValidationException, RestException
 from girder.models.item import Item
 from girder.models.user import User
+from girder.utility import JsonEncoder
 from ..models.annotation import AnnotationSchema, Annotation
+from ..models.annotationelement import Annotationelement
 
 
 class AnnotationResource(Resource):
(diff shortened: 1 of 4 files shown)


server/rest/annotation.py
```diff
@@ -141,16 +144,63 @@ def getAnnotationSchema(self, params):
     )
     @access.cookie
     @access.public
-    @filtermodel(model='annotation', plugin='large_image')
+    # @filtermodel(model='annotation', plugin='large_image')
```
jbeezley: Is this needed?

manthey: No.  I could remove the line or comment it further.  `@filtermodel` doesn't work with a function that returns a generator rather than returning a complete result.  The filtering is now done within the function.

jbeezley: Ah, I see.  I think removing it would be better.

server/rest/annotation.py
```diff
@@ -141,16 +144,62 @@ def getAnnotationSchema(self, params):
...
+                # Javascript parses the json identically with and without the
+                # trailing '.0'.  If we can trivially do so, string these
+                # trailing values.  This splits the string on ", but only if it
+                # has no possibility of escaped quotes, then strips the
+                # trailing zeroes from the non-quoted sections, then puts the
+                # string back together.
+                if '.0' in result and '\\"' not in result:
+                    result = '"'.join([
+                        d if '.0' in d and di % 2 else d.replace('.0,', ',').replace('.0]', ']')
+                        for di, d in enumerate(result.split('"'))])
```
jbeezley: This block makes me uncomfortable.  I would rather see the enhanced json serialization get moved to a function that gets unit tested independently.  To make it more readable, it would be better as a loop with better variable names.

I'm also curious about the performance implications of this.  Is the bandwidth saving worth the cost?  Would preprocessing the element dict to replace float's with int's be better than string manipulation?

manthey: I'm not wholly convinced the bandwidth savings are worth the cost.  On a local machine, it is faster to not do this and accept that we push lots of useless trailing zeroes.  As soon as the machine isn't local (even just on the intranet), however, the bandwidth starts becoming the dominant factor.

Replacing floats with ints was substantially slower (I agree it was conceptually cleaner).  The json serialization code gets done in C if it uses defaults.  It is faster to use the defaults and the string replacement than either preprocessing floats to ints or overriding the serialization method.

I can just remove this, and if we ever decide the bandwidth is vital we can reintroduce it or another method.

jbeezley: Sounds good to me.

PR #298, opened 2018-08-15
Change the default format for the tileIterator to NUMPY.

manthey:


manthey:
@cdeepakroy This changes the default for the tile iterator from PIL to numpy.  The tests have been updated, too.

PR #299, opened 2018-08-16
Update tile iterator docstring.

manthey:
Some information was only in the _tileIterator docstring.

### DIFF
--- a/server/tilesource/base.py
+++ b/server/tilesource/base.py
@@ -821,6 +821,8 @@ class TileSource(object):
                 image encoding.
             level: level of the current tile
             level_x, level_y: the tile reference number within the level.
+                Tiles are numbered (0, 0), (1, 0), (2, 0), etc.  The 0th tile
+                yielded may not be (0, 0) if a region is specified.
             tile_position: a dictionary of the tile position within the
                     iterator, containing:
                 level_x, level_y: the tile reference number within the level.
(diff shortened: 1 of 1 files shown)


server/tilesource/base.py
```diff
@@ -1460,16 +1460,36 @@ def tileIterator(self, format=(TILE_FORMAT_NUMPY, ), resample=False,
         Iterate on all tiles in the specifed region at the specified scale.
         Each tile is returned as part of a dictionary that includes
             x, y: (left, top) coordinates in current magnification pixels
-            width, height: size of current tile incurrent magnification pixels
+            width, height: size of current tile in current magnification pixels
             tile: cropped tile image
             format: format of the tile
             level: level of the current tile
             level_x, level_y: the tile reference number within the level.
```
cdeepakroy: It is not clear what `tile reference number` means

cooperlab: This is a larger issue - but somewhere we probably need to describe where the origin is and how the coordinate system works.

manthey: I think that there is a better place for a discussion of the image and tile coordinate systems.  Perhaps there should be a block comment elsewhere that the tile iterator refers to by reference.

Similarly, it would be nice if the internal `_tileIterator` method could only document how its parameters differ from the main `tileIterator` method.

I'll suggest that such things could be in another PR.

cdeepakroy: @manthey this looks good to me!

PR #308, opened 2018-10-03
Add Python 3.6 to the test matrix.

manthey:
Compress some tests files and add minor variations for Python 3.6.

Suppress some mapnik warnings when asking for pixels outside of the image.


manthey:
The most recent merge to master failed in Circle-CI's mapnik test because the base docker had moved to Python 3.6.  The image comparison tests we do as part of the mapnik source test are subtly different; more likely this is a GDal version and not a Python version issue, but I've added those variations to the tests so that they will pass.  Visually this has to do with subpixel antialiasing within the generated images.

manthey:
Also, I'll switch the required tests to the include the new circleci variants once this all passes and is approved.

PR #320, opened 2018-11-27
Harden the tiff reader.

manthey:
When trying to parse a file with the tiff reader, if it parses somewhat but reports no samples per pixel, the reader was raising a `TypeError`.  This caused Girder to fail to move to the next possible tile source.


law12019:
Looks good to me.

PR #322, opened 2018-11-29
Improve support for link files.

manthey:
For the mapnik tilesource, link files can be read directly.  For other tilesources, they can be converted via vips.


manthey:
The new code is covered, but only the version of gdal needed for it only tuns in the Python 3 Circle variants, so the coverage isn't reported.

manthey:
This is in the girder-3 branch.  If a backport is needed, open an issue.

PR #323, opened 2018-11-30
Handle deleting a file from a largeImage item.

manthey:
If you deleted a largeImage file from the item, the large image would still be displayed.  When the file is deleted, this appropriately removes the large image information.  On the web client, it will rerender the page to remove the viewer, too.


dgutman:
Thanks! I've seen some squirrelly behavior from that
.

PR #324, opened 2018-12-06
Add support for netcdf files.

manthey:
Some netcdf files are very slow to open, and the time penalty is incurred twice, once for gdal and once for mapnik.

Some data isn't rendered correctly at map boundaries.  This seems especially egregious in the region -180 to -160 longitude or so.

There are no tests yet.

You can access different time slices and subdata sets using the style parameter.  For instance, style={'band': 1} will get the first time slice of the default data set, band = 2 will get the second time slice, etc.  You can get non-default datasets by using the appropriate dataset name with the time slice, e.g., band = "precipitation_flux:1".

### DIFF
--- a/plugin_tests/source_mapnik_test.py
+++ b/plugin_tests/source_mapnik_test.py
@@ -191,15 +191,15 @@ class LargeImageSourceMapnikTest(common.LargeImageCommonTest):
 
         self._assertStyleResponse(itemId, {
             'band': 1,
-            'min': 'min',
+            'min': 'bad',
             'max': 100,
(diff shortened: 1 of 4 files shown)


manthey:
@aashish24 This should work IF you have appropriate versions of GDAL and mapnik installed.  The wheels at https://github.com/manthey/large_image_wheels work.

Note that I've based this off of PR #322, but it could probably be based off of master.

manthey:
This is in the girder-3 branch.  If a backport is needed, open an issue.

PR #328, opened 2019-01-27
Girder 3 web client tests

zachmullen:
This depends on https://github.com/girder/girder/pull/2914

The annotationSpec requires a fixture that I haven't added yet, but we can do it as a simple pytest fixture.


zachmullen:
In order to test this here, we'll need to bump the dependency on `pytest_girder`.

zachmullen:
@manthey we'll need to install node.js in the CircleCI environment. Do you have a preference on how we do that?

zachmullen:
LGTM!

PR #334, opened 2019-02-07
Thread parallelize thumbnail generation.

manthey:
This also allows caching associated images as part of the thumbnail generation job.

Before, when more than one thumbnail specification was given, each specification was processed in turn.  Now, all specifications for each image are processed to avoid having to load the image multiple times.

### DIFF
--- a/server/models/image_item.py
+++ b/server/models/image_item.py
@@ -256,7 +256,7 @@ class ImageItem(Item):
         self.removeThumbnailFiles(item)
         return deleted
 
-    def getThumbnail(self, item, width=None, height=None, **kwargs):
+    def getThumbnail(self, item, checkAndCreate=False, width=None, height=None, **kwargs):
         """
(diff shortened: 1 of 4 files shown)


manthey:
On a 4 physical (8 logical) core machine, I get about a 3x speed up in generating thumbnails, at which point the machine became mostly I/O bound, yet still seemed responsive for ordinary activity.

This uses threading for parallelization.  This means there will still be times that the python GIL prevents all cores from being used.  In principal, using processes instead of threads (via the `concurrent.futures.ProcessPoolExector`) would get around this, but that is known to be somewhat buggy in Python 2.  We could do this conditionally on Python version, or accept the risks for Python 2.  If, in practice, we are always I/O bound, then switching wouldn't matter.

Also, you can now precache associated images via a thumbnail spec like:
`[{"width": 160, "height": 100}, {"width": 160, "height": 100, "imageKey": "macro"}, {"width": 160, "height": 100, "imageKey": "label"}]`

PR #335, opened 2019-02-12
More robust access optimization for tile download route

zachmullen:
The history here is that we deploy large_image into some environments
in which the access.public decorator has its behavior changed to be
equivalent to access.token, and the optimization used to facilitate
cookie-based access to the tile endpoint does not work in that condition.
By also removing access.public and explicitly disabling the access
level checking, we can make this optimization work in that environment.

### DIFF
--- a/server/rest/tiles.py
+++ b/server/rest/tiles.py
@@ -300,13 +300,15 @@ class TilesItemResource(ItemResource):
     )
     # Without caching, this checks for permissions every time.  By using the
     # LoadModelCache, three database lookups are avoided, which saves around
-    # 6 ms in tests.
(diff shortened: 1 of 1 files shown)


brianhelba:
@zachmullen This seems like a significant usability problem with Girder's API. Can we fix this upstream so this sort of workaround is not necessary?

zachmullen:
@brianhelba I agree, I think this indicates that having `cookie` as a separate decorator is not a great idea, and it should simply be an argument to the other access decorators.

zachmullen:
https://github.com/girder/girder/issues/2936

PR #339, opened 2019-02-20
Memoize annotation style color.

manthey:
A significant fraction of the time preparing annotation with many elements is spent calling tinycolor.  Since most elements will share colors, memoizing this speeds it up substantially.

### DIFF
--- a/web_client/annotations/style.js
+++ b/web_client/annotations/style.js
@@ -1,37 +1,41 @@
-import _ from 'underscore';
 import tc from 'tinycolor2';
 
-const props = [
-    'label'
-];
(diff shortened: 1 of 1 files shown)


manthey:
When displaying a large number of rectangular annotations, this change plus https://github.com/OpenGeoscience/geojs/pull/979 and https://github.com/OpenGeoscience/geojs/pull/980 substantially speed up the initial render.

Specifically, when showing the maximum number of rectangles (currently 25000 / 4), after the json call for the data finished, a Chrome browser was spending around 2.4 seconds preparing to display the annotations.  With these three changes, it takes only around 0.6 seconds.

PR #342, opened 2019-03-15
Add the ability to copy annotations with items.

manthey:
This adjusts the `POST resource/copy`, `POST item/{id}/copy`, and `POST folder/{id}/copy` endpoints.  Each of these endpoints take an additional boolean parameter of `copyAnnotations`.

Note that the item is copied and then the annotations are copied, so, if the annotations are large, you will see the item in the browser and the annotations will show up a brief time later.  The endpoint only returns when the entire copy is done.

Copied annotations will have the same permissions as the copied item.

This resolves https://github.com/DigitalSlideArchive/HistomicsTK/issues/536.


jbeezley:
I can see it being confusing having the "Copy picked resources with annotations" option always there for Girder instances doing more than just serving annotated images.  Have we identified a scenario where we *don't* want to copy the annotations?  As a recursive copy, it makes sense that all child resources would be included.  I think it would be best to remove the extra menu option and make this the default behavior of the endpoint.

manthey:
Originally, we intentionally didn't copy the annotations with the items, but I don't recall why.  It can be slow if the annotations are large.  @dgutman and @cdeepakroy, do either of you remember why we didn't want to always copy annotations?

PR #347, opened 2019-04-17
Support some uncompressed tiled tiffs via the tiff reader.

manthey:
This allows us to read some additional tiffs that are not read by the openslider reader.  For instance, an uncompressed OME Tiff's first frame can be read with this change.


dgutman:
Been using the endpoint and it's working seccuessfuly

PR #348, opened 2019-04-17
Add support for OME Tiffs.

manthey:
This also adds generalized support for multi-frame large images.


manthey:
This still needs tests.

cooperlab:
Does this address the .scn format? I was surprised to see that OpenSlide claims to support this but we had issues reading these files.

manthey:
No, this change probably won't directly help scn files.  The openslide reader can read *some* scn files, but not all.  There will be another PR to add better support for some of those files.

manthey:
I've added python tests with a synthetic 3-frame OME tiff file I made.  The UI for choosing a frame doesn't have tests -- these are awkward to add in the current framework, so I'd prefer to defer adding those only to the girder-3 branch where it is easier.

If this works for others beside me, I'm content with the amount of testing for now.

PR #349, opened 2019-04-18
Show the largest image from SCN files.

manthey:
Ultimately, we will want to better parse associated images and handle more than one image from SCN files, but this at least allows the highest-res image to be used.


manthey:
This needs tests.

manthey:
@dgutman This branch is built on the OME Tiff branch, so pulling it will get that functionality as well.

dgutman:
this is necessary until we have a cleaner way to build a composite scene from the multiple scn regions

PR #352, opened 2019-04-30
Add two endpoints to act as a deepzoom server.

manthey:
This defaults to a tile size of 256 with no overlap.  It will only serve jpeg or png tiles.  Setting the tile size and overlap may not work for some readers, as those values have to be passed via query parameters.


manthey:
@dgutman We can add extra parameters to inform the viewer about pixel size, etc. once I have an example.

PR #357, opened 2019-06-28
Ensure threads do not access same GDAL datasets

wphicks:
Since GDAL datasets are [not thread-safe](https://trac.osgeo.org/gdal/wiki/FAQMiscellaneous#IstheGDALlibrarythread-safe), do not re-use them across threads in the Mapnik tile source. Instead, open a new Dataset when the MapnikTileSource dataset property is accessed. This may resolve intermittent segfaults recently observed on several Girder instances.

### DIFF
--- a/server/tilesource/mapniksource.py
+++ b/server/tilesource/mapniksource.py
@@ -125,7 +125,6 @@ class MapnikTileSource(FileTileSource):
         super(MapnikTileSource, self).__init__(path, **kwargs)
         self._bounds = {}
         self._path = self._getLargeImagePath()
-        self.dataset = gdal.Open(self._path)
         self.tileSize = 256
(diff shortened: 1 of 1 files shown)


manthey:
@wphicks In the girder-3 branch, a lock is used to handle this.  Opening the dataset multiple times (in each thread) seems to cause a lot of memory fragmentation in Python, causing memory to grow faster than using a lock and one open dataset.  In principal, using multiple datasets could be faster, but in a simple trial it didn't appear so.  I think, if you need this in Girder 2, I'd rather see the lock code backported.  (See https://github.com/girder/large_image/blob/girder-3/sources/mapnik/large_image_source_mapnik/__init__.py#L150)

wphicks:
Understood. I'll close this and open a backport PR.

PR #369, opened 2019-09-04
Add an option to get centroids.

manthey:
When there are too many elements to load all of them due to complexity, this loads all element centroids (actually the center of their bounding boxes) along with the element size (the diagonal of the bounding box).  If an element has not been loaded, a circle is shown at the centroid with an adjusted diameter that is related to the bounding box size.

This doesn't actually load *all* element centroids, just the first 2,000,000.  This number should be empirically determined based on GPU memory (something like half the memory divided by (16 or 17) * 4 * vpf would be appropriate).

Other improvements would be to (a) add hover events for centroid elements -- we know the ID and the center position, but not much else, (b) determine if any point's radius would be bigger than the webgl maximum point size, and, if so, render using a different primitive shape; otherwise, very large points render incorrectly, (c) if zoomed-out fully there seems to be an extra fetch of data on initial load, (d) compute a better centroid and size that the bounding box center and diagonal; this could be done on ingest.

This PR will benefit from a GeoJS release -- small points are rendered poorly in master.

When annotation centroids are retrieved, the response is a json-like object with a binary payload in the center.  If there are a large number of distinct styles in the annotation, this can become inefficient (for instance, if every element has a distinct fill and stroke color).

This has been developed on the 2.x-maintenance branch.  Tests will be added on the master branch version.

### DIFF
--- a/server/models/annotationelement.py
+++ b/server/models/annotationelement.py
@@ -128,7 +128,7 @@ class Annotationelement(Model):
             element for element in self.yieldElements(
                 annotation, region, annotation['_elementQuery'])]
 
-    def yieldElements(self, annotation, region=None, info=None):  # noqa
+    def yieldElements(self, annotation, region=None, info=None):
         """
         Given an annotation, fetch the elements from the database.
             When a region is used to request specific element, the following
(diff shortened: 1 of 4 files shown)


manthey:
Centroids

manthey:
@kheffah Would you care to review this?  The tests are only added on the PR that is for the master branch (#371).

kheffah:
@manthey Yes, I'd be happy to. I'll take a look.

kheffah:
**@TODO (?)** -> Probably not practical at the moment given the existing annotation primitives to work with, but using an ellipse that is axis-aligned will be better visually.

PR #374, opened 2019-10-01
Sgratzl/bioformats

sgratzl:
start with integrating https://www.openmicroscopy.org/bio-formats/ using https://github.com/CellProfiler/python-bioformats/ as a tile source. 

notes: 
 * since BioFormats is Java based a JVM is needed
 * and Python >= 3.5


manthey:
There are enough issues with this that I'll close this PR.  A future one will be opened with refactored code.

PR #380, opened 2019-10-02
Add an openjpeg source using the glymur library.

manthey:


alasla:
I gave you test images for jp2 files right?  Let me know if I didn't and I'll resend a small one you can use in testing.

manthey:
Yes.  I've added one of these files to the tests.

PR #388, opened 2019-10-18
Increase the response time limit.

manthey:
When the database is slow, getting and putting annotations can time out.  This increase the time limit.  It would be better to speed up the database, but until that is done, this at least allows it to function.


manthey:
@kheffah, @dgutman This is a partial solution to the failure to delete some annotations when the database is slow -- it doesn't speed anything up (and therefore doesn't solve the real problem), but should prevent it from failing.

Reference https://github.com/DigitalSlideArchive/HistomicsTK/issues/719.

kheffah:
@manthey Looks like a good temporary fix, Thanks!

PR #390, opened 2019-10-30
Handle different tiff orientations

manthey:
For a Tiled TIFF that is not in the standard top-left orientation, this uses a less-efficient code path that may composite up to four tiles to get a conceptual tile at the location desired.


manthey:
Confirmed in external use.

PR #397, opened 2019-11-14
Add a style option

manthey:
This better handles high bit-depth images.

The style option is an optional json-encoded parameter.  It is either an object with "bands" followed by a list or a single entry as from that list.  Each entry is an object with all optional parameters.  These are:

- band: either a number or a string.  If -1 or None, unspecified, the same as "gray".  If a number, a 1-based numerical index into the channels of the image.  If a string, one of ('red', 'green', 'blue', 'gray', 'alpha').  Note that 'gray' on an RGB or RGBA image will use the green band, and all colors on a greyscale image will use the luminance band.

- min: the value to map to the first palette value.  Defaults to 0.  'auto' to use 0 if the reported minimum and maximum of the band are between [0, 255] or use the reported minimum otherwise.  'min' or 'max' to always uses the reported minimum or maximum.

- max: the value to map to the last palette value.  Defaults to 255.  'auto' to use 0 if the reported minimum and maximum of the band are between [0, 255] or use the reported maximum otherwise.  'min' or 'max' to always uses the reported minimum or maximum.

- palette: a list of two or more color strings, where color strings are of the form #RRGGBB, #RRGGBBAA, #RGB, #RGBA (or anything else PIL can parse).

- nodata: the value to use for missing data.  null or unset to not use a nodata value.

- composite: either 'lighten' or 'multiply'.  Defaults to 'lighten' for all except the alpha band.  Bands are composited in the order listed.

- clamp: either True to clamp values outside of the [min, max] to the ends of the palette or False to make outside values transparent.

As part of this PR, this adds support for small OME TIFF images that aren't tiled.  Strips are aggregated together to make a minimum size, as serving individual strips is inefficient.

Add a histogram method and endpoint.


manthey:
This has been field-tested, and new features will need to build on this.  Merging.

PR #404, opened 2019-12-29
migrated slideatlas viewer to girder3, (+annotation user field)

law12019:
Update the slide atlas viewer to the latest annotation tools.


manthey:
@law12019 Aside from some linting issues, this fails in some of the tests because canvas context `resetTransform` isn't supported in the PhantomJS browser where the tests are run (and, if we care, also isn't supported in Edge).  Changing to `resetTransform()` to `setTransform(1, 0, 0, 1, 0, 0);` in SlideAtlas would fix this (otherwise we'll either need to mock in the tests or do something different).

manthey:
Merging this more for bookkeeping than any demand.

PR #418, opened 2020-01-21
Add a GDAL source.

manthey:
This handles more transforms than the mapnik source.  It doesn't currently handle NetCDF files.

It has been set to be preferred over mapnik.


manthey:
@wphicks Let me know if there are any issues.

PR #423, opened 2020-02-20
Use the web root public path where appropriate.

manthey:
On the Girder plugin, we need to use the webroot public path to reference packages stored alongside the girder plugin javascript.


manthey:
@dgutman If you pull this branch and rebuild Girder it should work in the proxy.

PR #433, opened 2020-03-31
Standardize frame metadata.

manthey:
Any tile source that exposes frames will have a frames list.  This list now will always contain Frame and Index values, and, where appropriate IndexC, IndexT, IndexZ, IndexXY values.  The top-level metadata will always have IndexRange and IndexStride entries.  When known, the top-level metadata will have channels and channelmap entries with channel names.


manthey:
@waxlamp This has tests and should work as desired.

PR #450, opened 2020-06-03
Read more OME Tiff files.

manthey:
Some OME Tiff files have one main record any everything else as a Plane.

Also, for TIFFs with mostly missing directories (such as OME Tiffs with only a tiled maximal resolution layer), synthesize intermediate empty directories to avoid excessive memory use.


manthey:
@dgutman This works, but is super slow on the OME Tiff examples you have.

PR #463, opened 2020-07-07
Add a bioformats tilesource.

manthey:
This is based off of an original branch by sgratzl, but has been extensively refactored to allow non-girder usage and for consistency with the rest of the code base.

There are some remaining issues:

- Some CZI files have multiple images that have different resolutions.

- Associated images don't have sensible labels (like "macro" and   "label").  We need to find the appropriate metadata for these.

- Tests on non-CZI files are needed.

- We may want to use a newer version of bioformats.  This would probably   involve forking python-bioformats.


manthey:
Externally confirmed.

PR #465, opened 2020-07-13
Work around a file descriptor issue in cheroot.

manthey:
There is an issue that was introduced in 8.1.0 that throws an error when file descriptor numbers are large (see https://github.com/cherrypy/cheroot/issues/249).  We can't pin cheroot to an older version, as that breaks support for modern python versions.


dgutman:
system stable since patch

kavehtp:
Looks like the issue is resolved on cheroot: https://github.com/cherrypy/cheroot/issues/249#issuecomment-661071261

manthey:
Once https://github.com/cherrypy/cheroot/pull/301 is merged and in a version on pypi, I'll revert this commit.

kavehtp:
Hi @manthey. They have released the new version: https://pypi.org/project/cheroot/

PR #469, opened 2020-07-22
Support reading OME Tiff files from sub-ifds.

manthey:
I anticipate generalizing this to all tiff files that are so constructed eventually.

We should add an explicit test case for this.


manthey:
@dgutman This will work on any OME Tiff that uses subifds assuming that the subifds are ordered largest to smallest and are every power of two below the main image (e.g., no skipped levels).  This matches the samples I've seen.  Let me know if any of the files you have are still super slow.

manthey:
I'm merging this into master as I have reports that it is working.  Any issues found can be new PRs.

PR #590, opened 2021-04-13
Add option to turn off BIGTIFF

banesullivan:
Adds a way to control the COG conversion utility's output and turn off the default BIGTIFF option

### DIFF
--- a/utilities/converter/large_image_converter/__init__.py
+++ b/utilities/converter/large_image_converter/__init__.py
@@ -117,7 +117,7 @@ def _generate_geotiff(inputPath, outputPath, **kwargs):
     }.update(yes_no)
     options.update({k: v for k, v in kwargs.items() if v not in (None, '')})
     cmdopt = ['-of', 'COG']
-    cmdopt += ['-co', 'BIGTIFF=%d' % predictor[yes_no['bigTiff']]]
+    cmdopt += ['-co', 'BIGTIFF=%d' % options[yes_no['bigTiff']]]
     cmdopt += ['-co', 'BLOCKSIZE=%d' % options['tileSize']]
     cmdopt += ['-co', 'COMPRESS=%s' % options['compression'].upper()]
     cmdopt += ['-co', 'QUALITY=%s' % options['quality']]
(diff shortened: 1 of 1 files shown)


utilities/converter/large_image_converter/__init__.py
```diff
@@ -103,15 +104,20 @@ def _generate_geotiff(inputPath, outputPath, **kwargs):
...
         'yes': 'YES',
     }
+    predictor = {
+        'horizontal': 'STANDARD',
+        'float': 'FLOATING_POINT',
+    }.update(yes_no)
     options.update({k: v for k, v in kwargs.items() if v not in (None, '')})
-    cmdopt = ['-of', 'COG', '-co', 'BIGTIFF=YES']
+    cmdopt = ['-of', 'COG']
+    cmdopt += ['-co', 'BIGTIFF=%d' % predictor[yes_no['bigTiff']]]
```
manthey, line 120: I think you want `predictor[yes_no['bigTiff']]` to be `yes_no[options['bigTiff']]`.

manthey:
There is a `IF_NEEDED` value for the `BIGTIFF` option that we should use instead.

manthey:
Actually, the GDAL docs say:
```
  -  IF_NEEDED will only create a BigTIFF if it is clearly needed (in
      the uncompressed case, and image larger than 4GB. So no effect
      when using a compression).
   -  IF_SAFER will create BigTIFF if the resulting file \*might\*
      exceed 4GB. Note: this is only a heuristics that might not always
      work depending on compression ratios.
```
Neither are guaranteed.  Looking at the code, `IF_SAFER` checks if the uncompressed image is > 2GB regardless of compression.  `IF_NEEDED` only if the uncompressed image is > 4GB AND uncompressed.  So `IF_SAFER` will probably mostly be sufficient as the default, and then we can allow the values to be overridden as in this PR.

utilities/converter/large_image_converter/__init__.py
banesullivan: Whoops. This is what happens when I try to multitask and not run the code

manthey:
#594 is going to cause this PR to need to be refactored.  It also will switch the default to `IF_SAFER`, which might be sufficient for most use cases.

manthey:
I'm going to close this, as I think the changes in #594 obviated the need for it.  Please reopen if you think it is still necessary.

Issue #567, opened 2021-03-12
Allow getRegion to output larger tiffs


manthey:
When getRegion is asked for a tiff file larger than what PIL can write, use vips.  We probably should tile output tiffs over a certain size, too.

manthey:
For geospatial tilesources, also include geospatial information.

manthey:
Added in #594.

PR #594, opened 2021-04-20
Provide an option to output large regions as tiled tiffs

manthey:
This also will geotag geospatial outputs.

Specifically, when calling the getRegion function or endpoint, if an encoding of `TILED` is used (previously this could have been one of `JPEG`, `PNG`, or `TIFF`), a tiled tiff is output.  For non-geospatial sources, this is tiled using vips.  For geospatial sources, this is tiled using gdal.

When using the `TILED` option, a variety of parameters that can be used in conversions can be used: `compression`, `quality`.  For compatibility, these can also be `tiffCompression` and `jpegQuality`.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -5,6 +5,7 @@
 ### Features
 - Allow setting the cache memory portion and maximum for tilesources (#601)
 - Add heatmap and griddata to the annotation schema (#589)
+- Regions can be output as tiled tiffs with scale or geospatial metadata (#594)
 
 ### Improvements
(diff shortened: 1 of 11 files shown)


manthey:
This resolves #567.  It handles metadata for geospatial sources, but doesn't include scale for non-geospatial sources yet.

manthey:
When getting a tiled region of a geospatial source without any styling, this uses gdal.Warp to create the region.  This code path preserves semantic information in the data, such as categorical mappings.

PR #635, opened 2021-08-12
Use tile-frames for better frame browsing.

manthey:
This requires a version of GeoJS newer than v1.2.0 to have any effect.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -5,11 +5,12 @@
 ### Features
 - Provide band information on all tile sources (#622, #623)
 - Add a tileFrames method to tile sources and appropriate endpoints to composite regions from multiple frames to a single output image (#629)
-- The test tile source now support frames (#631)
+- The test tile source now support frames (#631, #632, #634, #634)
 
 ### Improvements
 - Better handle TIFFs with missing levels and tiles (#624, #627)
 - Better report inefficient TIFFs (#626)
+- Smoother cross-frame navigation (#635)
 
 ## Version 1.6.2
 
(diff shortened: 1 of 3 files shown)


manthey:
As an example of the difference, this uses the exact same 1000-frame source file.  Before, we switched frames only when loaded:

https://user-images.githubusercontent.com/8781639/129386249-ce7902cc-afcf-40da-a53b-2fbdd0b944e8.mp4

After this update, we switch to the preview image immediately and then load the frames.  You can see jpeg artifacts in the preview frames; these vanish as the actual tiles are loaded when the scroll pauses.

https://user-images.githubusercontent.com/8781639/129386350-cd609aec-e2cc-435d-9853-56fe6779781a.mp4

aashish24:
this is very nice. hard to read the code with many codecov warnings (not sure why that is happening). what is a tile frame?

manthey:
I don't know why codecov is commenting PRs, since it is set to `comment: false`.

Many tile sources have multiple "frames" -- different bands on geospatial data, different z slices on histology images, time series, XY pan series, multiple channels, etc.  large_image aggregates all of this into a concept of "frames" so that there is a 0-based index of these separate images.  It requires that all of the images be the same resolution and have the same tile size and a similar tile structure (e.g., all frames are expected to have the same maximum depth of pyramidally arranged tiles, but not necessarily have the same levels of the pyramid filled out).

The synthetic image shown above simulates a moderate sized image with 1000 z slices -- I have some histology examples, but they are not necessarily public or, in some cases, particularly clear, so the synthetic image makes a more obvious demo.

PR #710, opened 2021-12-06
tifftools is a hard requirement of the tiff source.

manthey:


banesullivan:
Thanks, I caught this in the conda recipe and meant to add this back here

manthey:
> Thanks, I caught this in the conda recipe and meant to add this back here

I saw that, which is what led to this PR.  Thanks.

PR #714, opened 2021-12-07
Add support for named MPL colormaps

banesullivan:
These changes are adapted from what I implemented here in `localtileserver` to support using MPL colormaps by name:

https://github.com/banesullivan/localtileserver/blob/0c444034f63165857c57f4043a05cb73b5df5b71/localtileserver/palettes.py#L56-L63

This should make it so that any Matplotlib named colormap can be used as a `'palette'` when MPL is installed. https://matplotlib.org/stable/tutorials/colors/colormaps.html

(I haven't tested this locally yet)

### DIFF
--- a/sources/gdal/large_image_source_gdal/__init__.py
+++ b/sources/gdal/large_image_source_gdal/__init__.py
@@ -358,6 +358,15 @@ class GDALFileTileSource(FileTileSource, metaclass=LruCacheMetaclass):
         try:
             return attrgetter(palette)(palettable).hex_colors
         except AttributeError:
+            pass
+        try:
+            # Attempt getting palette by MPL cmap name, e.g. 'viridis' or 'jet'
+            import matplotlib
+            import matplotlib.colors as mcolors
+
+            cmap = matplotlib.cm.get_cmap(palette, 255)  # 255 might be overkill
+            return [mcolors.rgb2hex(cmap(i)) for i in range(cmap.N)]
+        except (ImportError, ValueError):
             raise TileSourceError('Palette is not a valid palettable path.')
 
     def getProj4String(self):
--- a/test/test_source_gdal.py
+++ b/test/test_source_gdal.py
@@ -108,7 +108,7 @@ def testTileStyleMatplotlibColormap():
     testDir = os.path.dirname(os.path.realpath(__file__))
     imagePath = os.path.join(testDir, 'test_files', 'rgb_geotiff.tiff')
     style = json.dumps({'band': 1, 'min': 0, 'max': 100,
-                        'palette': 'jet',  # use a named MPL colormap
+                        'palette': 'viridis',  # use a named MPL colormap
                         'scheme': 'linear'})
     source = large_image_source_gdal.open(
         imagePath, projection='EPSG:3857', style=style, encoding='PNG')
--- a/test/test_source_gdal.py
+++ b/test/test_source_gdal.py
@@ -104,6 +104,17 @@ def testTileLinearStyleFromGeotiffs():
     _assertImageMatches(image, 'geotiff_style_linear_7_22_51')
 
 
+def testTileStyleMatplotlibColormap():
+    testDir = os.path.dirname(os.path.realpath(__file__))
+    imagePath = os.path.join(testDir, 'test_files', 'rgb_geotiff.tiff')
+    style = json.dumps({'band': 1, 'min': 0, 'max': 100,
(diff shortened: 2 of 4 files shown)


sources/gdal/large_image_source_gdal/__init__.py
```diff
@@ -358,6 +358,15 @@ def getHexColors(palette):
         try:
             return attrgetter(palette)(palettable).hex_colors
         except AttributeError:
+            pass
+        try:
+            # Attempt getting palette by MPL cmap name, e.g. 'viridis' or 'jet'
+            import matplotlib
+            import matplotlib.colors as mcolors
+
+            cmap = matplotlib.cm.get_cmap(palette, 255)  # 255 might be overkill
```
banesullivan, line 367: This is probably major overkill. @manthey, any idea what might be a more reasonable value here?

manthey:
I think the colormaps (both from palettable and what you are adding here) should ideally be moved up to the base tilesource class, rather than exist just for the GDAL source (and mapnik, since it inherits from GDAL).  This would add palettable as a dependency (optional?) to large_image.  I'll make an issue for this, as it is beyond what this PR should be trying to do.  It would also be nice if you could specify a palette as a single color (e.g., "palette": "green", which would expand to "palette": ["black", "green"]).

sources/gdal/large_image_source_gdal/__init__.py
manthey: You don't need to specify the length.  That is,
`cmap = matplotlib.cm.get_cmap(palette)` will yield the same as `cmap = matplotlib.cm.get_cmap(palette, len(maplotlib.cm.get_cmap(palette).colors))`

banesullivan:
> It would also be nice if you could specify a palette as a single color (e.g., "palette": "green", which would expand to "palette": ["black", "green"]).

This would be pretty easy to implement as well. We could borrow some code from PyVista to have big list of colors and their hex values: https://github.com/pyvista/pyvista/blob/main/pyvista/plotting/colors.py

Or perhaps MPL actually has this support somewhere?

banesullivan:
> This would add palettable as a dependency (optional?) to large_image

Palettable is pure Python with no dependencies so I think it's absolutely okay to make a hard dependency of large_image. MPL on the other hand should remain optional

manthey:
> Or perhaps MPL actually has this support somewhere?

`matplotlib.colors.to_hex` does what we'd want: e.g., `matplotlib.colors.to_hex('blue')` or `matplotlib.colors.to_hex('xkcd:blue')`.  Generally, `matplotlib.colors.get_named_colors_mapping().keys()` will get a list of all color names it handles.

banesullivan:
I think it's worth having general, named colors support without requiring MPL - to do that, should we make a big list of colors, effectively taking the code out of MPL and putting it in large_image?

banesullivan:
When deploying a tile serving app, every requirement has a cost and MPL isn't trivial

manthey:
> I think it's worth having general, named colors support without requiring MPL - to do that, should we make a big list of colors, effectively taking the code out of MPL and putting it in large_image?

There is a package that already does this (mpl-colors).

manthey:
> > I think it's worth having general, named colors support without requiring MPL - to do that, should we make a big list of colors, effectively taking the code out of MPL and putting it in large_image?
> 
> There is a package that already does this (mpl-colors).

Nevermind, mpl-colors imports matplotlib.

test/test_source_gdal.py
```diff
@@ -104,6 +104,17 @@ def testTileLinearStyleFromGeotiffs():
     _assertImageMatches(image, 'geotiff_style_linear_7_22_51')
 
 
+def testTileStyleMatplotlibColormap():
+    testDir = os.path.dirname(os.path.realpath(__file__))
+    imagePath = os.path.join(testDir, 'test_files', 'rgb_geotiff.tiff')
+    style = json.dumps({'band': 1, 'min': 0, 'max': 100,
+                        'palette': 'jet',  # use a named MPL colormap
```
manthey, line 111: I don't see `jet` in MPL.  Did it get added in your test project by another package?

banesullivan: Oh shoot, your right. I have `colorcet` installed which added jet to MPL. I will change this to viridis for the test

manthey:
I've made PR #724 that supercedes this.  It generalizes this behavior for all tile sources and expands its capabilities.

PR #725, opened 2021-12-13
Use a different format for specifying extras based on platform.

### LINKED ISSUE #723: Memcached depend condition not working on Windows
This condition should prevenr pylibmc from being installed on windows:

https://github.com/girder/large_image/blob/8e8deaae3b4165522a6894ea8d0cb5432d4839c2/setup.py#L13

However, it is not working:

```
pip install -r requirements.txt
  shell: C:\Program Files\PowerShell\7\pwsh.EXE -command ". '{0}'"
  env:
    CONDA_PKGS_DIR: C:\Users\runneradmin\conda_pkgs_dir
...
Collecting large-image[memcached]
  Downloading large_image-1.8.11-py3-none-any.whl (51 kB)
...
Collecting pylibmc>=1.5.1
  Downloading pylibmc-1.6.1.tar.gz (64 kB)
...
```

see Windows CI: https://github.com/banesullivan/localtileserver/runs/4500283500?check_suite_focus=true

I wonder if this is because the wheel is built on Linux and `pylibmc` is then listed as a dependency.

manthey:
See
https://setuptools.pypa.io/en/latest/userguide/dependency_management.html#platform-specific-dependencies

Closes #723.  See the discussion there.

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -1,6 +1,5 @@
 import itertools
 import os
-import platform
 
 from setuptools import setup
 
@@ -10,7 +9,7 @@ description = 'Python modules to work with large, multiresolution images.'
 long_description = readme
 
 extraReqs = {
-    'memcached': ['pylibmc>=1.5.1'] if platform.system() != 'Windows' else [],
+    'memcached': ['pylibmc>=1.5.1 ; platform_system != "Windows"'],
     'converter': ['large-image-converter'],
 }
 sources = {


banesullivan (approved):
As far as I can tell, this should work... 

I created #728 to attempt testing this... it's probably more trouble than it's worth to set up CI to install all dependencies and test on Windows

manthey:
I like where #728 is going and look forward to having it.

PR #728, opened 2021-12-13
[WIP] GitHub Action to test wheels on all platforms

banesullivan:
Attempt to test #725

### DIFF
--- a/.circleci/make_wheels.sh
+++ b/.circleci/make_wheels.sh
@@ -13,14 +13,14 @@ mkdir ~/wheels
 # If we need binary wheels, we would need to step through the various versions
 # of python and also run auditwheel on each output
 pip wheel . --no-deps -w ~/wheels && rm -rf build
-cd "$ROOTPATH/girder"
-pip wheel . --no-deps -w ~/wheels && rm -rf build
-cd "$ROOTPATH/girder_annotation"
-pip wheel . --no-deps -w ~/wheels && rm -rf build
+# cd "$ROOTPATH/girder"
+# pip wheel . --no-deps -w ~/wheels && rm -rf build
+# cd "$ROOTPATH/girder_annotation"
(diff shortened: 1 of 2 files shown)


banesullivan:
@manthey, any idea what's going on here:

```
ERROR: Cannot install -r requirements-dev-core.txt (line 13) and large-image-source-tiff 0.1.dev1+g79153a0 (from /home/runner/work/large_image/large_image/sources/tiff) because these package versions have conflicting dependencies.
ERROR: ResolutionImpossible: for help visit https://pip.pypa.io/en/latest/user_guide/#fixing-conflicting-dependencies
INFO: pip is looking at multiple versions of large-image-source-openjpeg to determine which version is compatible with other requirements. This could take a while.
INFO: pip is looking at multiple versions of large-image-source-nd2 to determine which version is compatible with other requirements. This could take a while.
INFO: pip is looking at multiple versions of large-image-source-gdal to determine which version is compatible with other requirements. This could take a while.
INFO: pip is looking at multiple versions of large-image-source-dummy to determine which version is compatible with other requirements. This could take a while.
INFO: pip is looking at multiple versions of large-image-source-deepzoom to determine which version is compatible with other requirements. This could take a while.
INFO: pip is looking at multiple versions of <Python from Requires-Python> to determine which version is compatible with other requirements. This could take a while.
INFO: pip is looking at multiple versions of large-image-source-bioformats to determine which version is compatible with other requirements. This could take a while.

The conflict is caused by:
    The user requested large-image-source-tiff 0.1.dev1+g79153a0 (from /home/runner/work/large_image/large_image/sources/tiff)
    large-image-source-ometiff 0.1.dev1+g79153a0 depends on large-image-source-tiff>=1.0.0

To fix this you could try to:
1. loosen the range of package versions you've specified
2. remove package versions to allow pip attempt to solve the dependency conflict
```

PR #735, opened 2021-12-15
Don't require girder-worker unless using tasks.

manthey:


manthey:
This was already added as part of #715.

PR #741, opened 2021-12-16
Add a getAvailableNamedPalettes function.

### LINKED ISSUE #731: Add getAvailableNamedPalettes() utility function
This would list all names that can be used as style palettes that are available, including color names.

Would be desirable to have an option to only list palettes that are not color names (e.g., "red" is functionally a palette of ["black", "red"], but seems like it is in a different category from "viridis")?

manthey:
Closes #731.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,9 +1,10 @@
 # Change Log
 
-## Unreleased
+## Version 1.9.0
 
 ### Features
 - Better palette support in styles ([#724](../../pull/724))
+- Enumerate available palettes ([#741](../../pull/741))
 
 ### Improvements
 - The openjpeg girder source handles uploaded files better ([#721](../../pull/721))
--- a/large_image/tilesource/utilities.py
+++ b/large_image/tilesource/utilities.py
@@ -1,5 +1,6 @@
 import io
 import math
+import types
 import xml.etree.ElementTree
 from collections import defaultdict
 from operator import attrgetter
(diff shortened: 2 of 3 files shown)


manthey:
There are a _lot_ of named palettes.  With palettable and matplotlib, I get 3380 named palettes excluding individual colors and 4541 including them.

banesullivan:
>  I get 3380 named palettes excluding individual colors and 4541 including them.

ooof. that's a lot. Too many if try to send that over to a drop down menu.

Could we separate them into a dictionary. E.g,

```py
{
'matplotlib': ['...', ...],
'palettable': ['...', ...],
'colors': ['...', ...],
...
}
```

manthey:
A large number of the variations come from (a) palettable having distinct names for each variant with a different number of points in the lookup table (e.g., `cartocolors.diverging.Earth_2`, `cartocolors.diverging.Earth_3`, `cartocolors.diverging.Earth_4`, `cartocolors.diverging.Earth_5`, `cartocolors.diverging.Earth_6`, `cartocolors.diverging.Earth_7`), and most palettes having an `_r` reversed variant (so `cartocolors.diverging.Earth_2` and `cartocolors.diverging.Earth_2_r`).  If we exclude the lower count palettes, the reversed palettes, and the colors, the count drops to a mere 212 palettes.

I think a better choice might be to add a `reduced` option to the function that would exclude the reversed palettes and the lower numbered palettes, then `getAvailableNamedPalettes(includeColors=False, reduced=True)` would return a much smaller set (212).

manthey:
I pushed a change adding the `reduced` option.  If you think this is sufficient, please approve the PR.  If you still think return a dictionary of results makes sense, we probably need to discuss the granularity of it.  For instance, should PIL colors and matplotlib colors be separate entries?  Should we subset the palettable entries in any manner?

manthey:
Lastly, in the reduced set of palettes, there are some function duplicates: `lightbartlein.diverging.BlueGray_8` is the same as  `lightbartlein.diverging.BlueGrey_8`; matplotlib shares 9 palettes with palettable such as `Pastel2` and `colorbrewer.qualitative.Pastel2_8`.  Short of instantiating the palettes, duplicates are somewhat hard to detect (and then you have to pick a preferred name).

banesullivan (approved):
I think shortening the way you did is a good implementation and this PR is good to merge.

Downstream, we may just create this list ourselves in RGD. I'm not huge on palettable and I think we'll just stick with MPL's full set (along with cmocean's as well) as the reversed maps are very useful

Issue #678, opened 2021-11-12
Support image overlay annotations


manthey:
Add image overlay annotations to the schema and renderer of the girder_annotation module.

Specifically, an image annotation needs to contain:
- an image (this could be a relative path to the main image, or a Girder image ID).  Images should themselves be large_image or directly readable via a common browser (e.g., png or jpeg)
- default opacity for the image (if unset, this defaults to 1)
- default location for the image
  I'm not sure the best format for this, as we could be very general (e.g., (a) upper-left, width, and height, (b) center, rotation, scale, (c) all four corners, (d) affine transform matrix, etc.)
  By default, this would place the image at the upper left in its native resolution

Check what others have done for this (I think that there is a Girder 1 implementation of this on GitHub).

Relative paths are probably conceptually in the Girder resource path space (since that will translate to physical filesystems directly and is clear what it means in remote assetstores).

manthey:
Since primary files are often copied, we need to make sure that they remain associated with the copied primary file.  This can be done as a separate task from the basic implementation.

naglepuff:
> Check what others have done for this (I think that there is a Girder 1 implementation of this on GitHub).

Do you know where I might be able to find this attempt? Would it be here in the `large_image` repo or an older branch of the `girder` repo?

manthey:
I'm guessing it is here: https://github.com/abcsFrederick/overlays, but it may be in a different repo within that organization.

PR #742, opened 2021-12-17
Add image overlays to annotation schema

naglepuff:
This PR addresses server-side changes for #678 

Changes include:

- Modification of existing annotation schema to allow for specification of an image overlay

An image overlay is specified on the annotation itself, and not as an element. At a minimum, an image overlay requires a girder large image item ID. Opacity and location can also be specified. Location is specified with properties for position of the top-left corner of the image to overlay, as well as ta height and width. By default the overlaid image has opacity 1 and is aligned in the upper left in its native resolution.

The reason I went with adding the overlay as a direct child to the annotation schema, and not as a type of annotation element, is that I am trying to avoid a major refactor of what an annotation element is. It seems that a lot of downstream code assumes that annotation elements are made up of points/lines/polygons, and image overlays will need to be handled differently by client code (rendering on different layers, ensuring the linked tiled image exists). 

I am open to discussing other possibilities regarding how best to model image overlays on the back end. Particularly I'm wondering if my reasoning above is indeed valid, or if there's another, better approach, like making a dedicated `girder Model` for image overlays.

### DIFF
--- a/girder_annotation/girder_large_image_annotation/models/annotation.py
+++ b/girder_annotation/girder_large_image_annotation/models/annotation.py
@@ -450,12 +450,26 @@ class AnnotationSchema:
         ]
     }
 
+    transformArray = {
+        'type': 'array',
+        'items': {
+            'type': 'array',
+            'minItems': 2,
+            'maxItems': 2
+        },
+        'minItems': 2,
+        'maxItems': 2,
+        'description': 'A 2D matrix representing the transform of an '
+                       'image overlay.'
+    }
+
     overlaySchema = {
         '$schema': 'http://json-schema.org/schema#',
         'type': 'object',
         'properties': {
             'girderId': {
                 'type': 'string',
+                'pattern': '^[0-9a-f]{24}$',
                 'description': 'Girder item ID containing the image to '
                                'overlay.'
             },
@@ -466,22 +480,22 @@ class AnnotationSchema:
                 'description': 'Default opacity for this image overlay. Must '
                                'be between 0 and 1. Defaults to 1.'
             },
-            'location': {
+            'transform': {
                 'type': 'object',
-                'description': 'Position (upper-left coordinate, width, and '
-                               'height) of the overlay.',
+                'description': 'Specification for an affine transform of the '
+                               'image overlay. Includes a 2D transform matrix, '
+                               'an X offset and a Y offset.',
                 'properties': {
-                    'upperLeftCorner': coordSchema,
-                    'width': {
+                    'xoffset': {
                         'type': 'number',
                         'minimum': 0
                     },
-                    'height': {
+                    'yoffset': {
                         'type': 'number',
                         'minimum': 0
-                    }
+                    },
+                    'transform': transformArray
                 },
-                'required': ['upperLeftCorner', 'width', 'height']
             }
         },
         'required': ['girderId'],
--- a/girder_annotation/girder_large_image_annotation/models/annotation.py
+++ b/girder_annotation/girder_large_image_annotation/models/annotation.py
@@ -450,6 +450,46 @@ class AnnotationSchema:
         ]
     }
 
+    overlaySchema = {
+        '$schema': 'http://json-schema.org/schema#',
+        'type': 'object',
+        'properties': {
+            'girderId': {
+                'type': 'string',
+                'description': 'Girder item ID containing the image to '
+                               'overlay.'
+            },
+            'opacity': {
+                'type': 'number',
+                'minimum': 0,
+                'maximum': 1,
+                'description': 'Default opacity for this image overlay. Must '
+                               'be between 0 and 1. Defaults to 1.'
+            },
+            'location': {
+                'type': 'object',
+                'description': 'Position (upper-left coordinate, width, and '
+                               'height) of the overlay.',
+                'properties': {
+                    'upperLeftCorner': coordSchema,
+                    'width': {
+                        'type': 'number',
+                        'minimum': 0
+                    },
+                    'height': {
+                        'type': 'number',
+                        'minimum': 0
+                    }
+                },
+                'required': ['upperLeftCorner', 'width', 'height']
+            }
+        },
+        'required': ['girderId'],
+        'additionalProperties': False,
+        'description': 'An image to overlay onto another like an '
+                       'annotation.'
+    }
+
     annotationElementSchema = {
         '$schema': 'http://json-schema.org/schema#',
         # Shape subtypes are mutually exclusive, so for efficiency, don't use
@@ -496,7 +536,8 @@ class AnnotationSchema:
                 'title': 'Image Markup',
                 'description': 'Subjective things that apply to a '
                                'spatial region.'
-            }
+            },
+            'overlay': overlaySchema
         },
         'additionalProperties': False
     }
--- a/girder_annotation/girder_large_image_annotation/models/annotation.py
+++ b/girder_annotation/girder_large_image_annotation/models/annotation.py
@@ -467,6 +467,10 @@ class AnnotationSchema:
         '$schema': 'http://json-schema.org/schema#',
         'type': 'object',
         'properties': {
+            'type': {
+                'type': 'string',
+                'enum': ['imageoverlay']
+            },
             'girderId': {
                 'type': 'string',
                 'pattern': '^[0-9a-f]{24}$',
@@ -494,11 +498,11 @@ class AnnotationSchema:
                         'type': 'number',
                         'minimum': 0
                     },
-                    'transform': transformArray
+                    'matrix': transformArray
                 },
             }
         },
-        'required': ['girderId'],
+        'required': ['girderId', 'type'],
         'additionalProperties': False,
         'description': 'An image to overlay onto another like an '
                        'annotation.'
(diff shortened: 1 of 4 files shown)


girder_annotation/girder_large_image_annotation/models/annotation.py
```diff
@@ -496,7 +536,8 @@ class AnnotationSchema:
                 'title': 'Image Markup',
                 'description': 'Subjective things that apply to a '
                                'spatial region.'
-            }
+            },
+            'overlay': overlaySchema
```
manthey, line 540: I'd rather this was just another possible annotation element.  We can expect the viewer to do something similar to how we handle heatmaps -- in `web_client/annotations/convertFeatures.js` we create a layer per heatmap (which isn't actually strictly necessary, as two heatmaps could share a layer) and manage it there.

girder_annotation/girder_large_image_annotation/models/annotation.py
```diff
@@ -450,6 +450,46 @@ class AnnotationSchema:
...
+            },
+            'location': {
+                'type': 'object',
+                'description': 'Position (upper-left coordinate, width, and '
+                               'height) of the overlay.',
+                'properties': {
+                    'upperLeftCorner': coordSchema,
+                    'width': {
+                        'type': 'number',
+                        'minimum': 0
```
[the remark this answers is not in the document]
manthey: We probably will eventually want to allow for an affine transform matrix instead.  We end up with six values: xoffset, yoffset, plus a 2x2 scaling matrix [[s11, s12], [s21, s22]].  For the width/height condition, this ends up being [[specified width/image width, 0], [0, specified height/image height]].  In many instances, we don't want to apply a scale or skew, but just an offset and a rotation, but we can leave that to whoever specifies the matrix.

girder_annotation/girder_large_image_annotation/models/annotation.py
```diff
@@ -450,6 +450,46 @@ class AnnotationSchema:
         ]
     }
 
+    overlaySchema = {
+        '$schema': 'http://json-schema.org/schema#',
+        'type': 'object',
+        'properties': {
+            'girderId': {
+                'type': 'string',
+                'description': 'Girder item ID containing the image to '
```
manthey, line 459: We should enforce that this is a 24 digit hexadecimal number `'pattern': '^[0-9a-f]{24}$'`.

girder_annotation/girder_large_image_annotation/models/annotation.py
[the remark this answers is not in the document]
manthey: We could add an optional field of `transform` which would be an array of exactly four numbers (and document them).  In this case, it would be nice to specify that you can't specify a transform AND a width or height.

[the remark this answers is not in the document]
naglepuff: See 95c8b45. I think if we eventually want to allow for an affine transform matrix, it makes sense to be able to specify that now.

girder_annotation/girder_large_image_annotation/models/annotation.py
```diff
@@ -450,6 +450,46 @@ class AnnotationSchema:
...
+                    'width': {
+                        'type': 'number',
+                        'minimum': 0
+                    },
+                    'height': {
+                        'type': 'number',
+                        'minimum': 0
+                    }
+                },
+                'required': ['upperLeftCorner', 'width', 'height']
```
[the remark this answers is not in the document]
naglepuff: See 95c8b45. Properties for the position of the overlay are no longer specified as required.

girder_annotation/girder_large_image_annotation/models/annotation.py
naglepuff: See d0aa882. Once I added image overlays as an option for elements, I was hitting errors on upload of a new annotation during the calculation of the bounding box.

My naive approach to bounding box calculation for image overlays was just to set it to the size of the entire base image. Perhaps a better approach would be to use the transform properties to calculate the exact coordinates of the overlay. Thoughts?

girder_annotation/girder_large_image_annotation/models/annotation.py
naglepuff: Fixed in 95c8b45

girder_annotation/girder_large_image_annotation/models/annotation.py
manthey: It would be better to calculate the actual bounding box, since, in principle, you could have several images and only need to fetch the ones that intersect some region.  Similarly, the overlay could actually extend outside of the base image.

naglepuff: I just committed 9c44499, which should do a better job of calculating the bounding box. Let me know if you have any additional thoughts/suggestions.

girder_annotation/girder_large_image_annotation/models/annotationelement.py
```diff
@@ -374,6 +415,23 @@ def _boundingBox(self, element):
             bbox['highx'] = max(x0, x1)
             bbox['highy'] = max(y0, y1)
             bbox['details'] = len(element['values'])
+        elif element.get('type') == 'imageoverlay':
+            lowx = highx = lowy = highy = highz = 0
+            try:
+                lowx, highx, lowy, highy = Annotationelement()._overlayBounds(element)
+                baseImageItem = ImageItem().load(annotation['itemId'], force=True)
+                baseImageMetadata = ImageItem().getMetadata(baseImageItem)
+                highz = baseImageMetadata.get('levels', 0)
```
manthey, line 424: In thinking about this, I don't think we want to store levels in the z bounding box -- the levels aren't conceptually heights (whereas there ARE multi-frame images that have a z-stack of frames).  I think for now, the z will always just be zero.  If we do that, then we don't have to pass the parent annotation into the `_boundingBox` method.

girder_annotation/girder_large_image_annotation/models/annotationelement.py
```diff
@@ -374,6 +415,23 @@ def _boundingBox(self, element):
...
+            except Exception as e:
+                logger.error('Error generating bounding box for image overlay annotation: %s' % e)
```
manthey, line 426: I think it would be better to have this guard in the `_overlayBounds` method rather than here.

naglepuff: See 95cf345

girder_annotation/girder_large_image_annotation/models/annotationelement.py
naglepuff: See 95cf345

manthey:
Great.  It would be nice to have some basic tests of this.  For instance, you could add a test in the `girder_annotation/test_annotation/test_annotations.py` file, in the `TestLargeImageAnnotationElement` class, adding something like:
```
    def testOverlayAnnotation(self, server, admin, fsAssetstore):
        file = utilities.uploadExternalFile('sample_image.ptif', admin, fsAssetstore)
        itemId = str(file['itemId'])
        bbox = Annotationelement()._boundingBox(
            {'type': 'imageoverlay', 'girderId': itemId})
        assert bbox == {
            'lowx': 0, 'lowy': 0, 'lowz': 0,
            'highx': 58368, 'highy': 12288, 'highz': 0,
            'details': 1,
            'size': (58368**2 + 12288**2)**0.5}
```
would test some of the bounding box code.  You'll want some additional conditions with offset and matrix values set.

girder_annotation/test_annotation/test_annotations.py
```diff
@@ -416,6 +416,55 @@ def testGetNextVersionValue(self):
...
+                'xoffset': 500,
+                'yoffset': 1000,
+                'matrix': [[0.5, 0], [0, 0.5]]
+            }
+        })
+        assert lowx == 500
+        assert lowy == 1000
+        assert highx == (58368 / 2) + 500
+        assert highy == (12288 / 2) + 1000
+
```
manthey, line 467: Great.  Can you also add a test that invokes the `_boundingBox` method to test that it invokes the `_overlayBounds` method appropriately?  It doesn't have to test the various transform options, just the `'type': 'imageoverlay'` pass-through.

girder_annotation/girder_large_image_annotation/models/annotationelement.py
```diff
@@ -336,6 +337,50 @@ def removeOldElements(self, annotation, oldversion=None):
...
+            corners = [np.matmul(np.array(corner), transformMatrix) for corner in corners]
+            offsetArray = np.array([transform.get('xoffset', 0), transform.get('yoffset', 0)])
+            corners = [np.add(corner, offsetArray) for corner in corners]
+            # use .item() to convert back to native python types
+            lowx = min([corner[0] for corner in corners]).item()
+            highx = max([corner[0] for corner in corners]).item()
+            lowy = min([corner[1] for corner in corners]).item()
+            highy = max([corner[1] for corner in corners]).item()
+        except Exception as e:
+            logger.error('Error generating bounding box for image overlay annotation: %s' % e)
```
manthey, line 381: You could just do `logger.exception(''Error generating bounding box for image overlay annotation')`

manthey (approved):
You might have to convert this to a non-draft PR in order to merge.  It looks good.

PR #750, opened 2021-12-23
Display image overlay annotations

### LINKED ISSUE #678: Support image overlay annotations

naglepuff:
Implement client support for image overlay annotation elements.

Changes include:

- [x] Draw image overlays when annotations containing them are selected in the Girder `large_image` UI
- [x] Remove image overlays when their annotations are deselected
- [x] Support rendering image overlays with opacity specified in the annotation JSON
- [x] Support image overlay offset
- [x] Support image overlay transform via affine matrix

(opening as draft PR to kick start discussion about implementation details)

Closes #678

### DIFF
--- a/girder_annotation/girder_large_image_annotation/web_client/annotations/convertFeatures.js
+++ b/girder_annotation/girder_large_image_annotation/web_client/annotations/convertFeatures.js
@@ -95,32 +95,6 @@ function convertHeatmap(record, properties, layer) {
 }
 
 /**
- * Convert an image overlay annotation to a geojs feature.
- *
- * @param record: the image overlay annotation element.
- * @param properties: a property map of additional data, such as the original
- *      annotation id.
- * @param layer: the layer where this may be added.
- */
-// function convertImageOverlay(record, properties, layer) {
-    // /* Image overlays need to be in their own layer */
-    // const overlayItemId = record.girderId;
-
-    // // Get overlay overlay metadata to inform layer params
-    // restRequest({
-        // url: `item/${overlayItemId}/tiles`
-    // }).done((response) => {
-        // const map = layer.map();
-        // const geo = window.geo;
-        // let params = geo.util.pixelCoordinateParams(
-           // map.node(), response.sizeX, response.sizeY, response.tileHeight, response.tileWidth
-        // );
-        // return params;
-    // }).fail((response) => {
-    // });
-// }
-
-/**
  * Convert a griddata heatmap annotation to a geojs feature.
  *
  * @param record: the griddata heatmap annotation element.
--- a/girder_annotation/girder_large_image_annotation/web_client/annotations/convertFeatures.js
+++ b/girder_annotation/girder_large_image_annotation/web_client/annotations/convertFeatures.js
@@ -1,5 +1,3 @@
-// import { restRequest } from '@girder/core/rest';
-
 /**
  * Create a color table that can be used for a heatmap.
  *
--- a/girder_annotation/girder_large_image_annotation/web_client/annotations/convertFeatures.js
+++ b/girder_annotation/girder_large_image_annotation/web_client/annotations/convertFeatures.js
@@ -1,3 +1,5 @@
+// import { restRequest } from '@girder/core/rest';
+
 /**
  * Create a color table that can be used for a heatmap.
  *
@@ -91,6 +93,32 @@ function convertHeatmap(record, properties, layer) {
     heatmap._ownLayer = true;
     return [heatmap];
 }
+
+/**
+ * Convert an image overlay annotation to a geojs feature.
+ *
+ * @param record: the image overlay annotation element.
+ * @param properties: a property map of additional data, such as the original
+ *      annotation id.
+ * @param layer: the layer where this may be added.
+ */
+// function convertImageOverlay(record, properties, layer) {
+    // /* Image overlays need to be in their own layer */
+    // const overlayItemId = record.girderId;
+
+    // // Get overlay overlay metadata to inform layer params
+    // restRequest({
+        // url: `item/${overlayItemId}/tiles`
+    // }).done((response) => {
+        // const map = layer.map();
+        // const geo = window.geo;
+        // let params = geo.util.pixelCoordinateParams(
+           // map.node(), response.sizeX, response.sizeY, response.tileHeight, response.tileWidth
+        // );
+        // return params;
+    // }).fail((response) => {
+    // });
+// }
 
 /**
  * Convert a griddata heatmap annotation to a geojs feature.
--- a/girder_annotation/girder_large_image_annotation/web_client/views/imageViewerWidget/geojs.js
+++ b/girder_annotation/girder_large_image_annotation/web_client/views/imageViewerWidget/geojs.js
@@ -212,9 +212,10 @@ var GeojsImageViewerWidgetExtension = function (viewer) {
                     params.layer.autoshareRenderer = false;
                     params.layer.opacity = overlay.opacity || 1;
                     let overlayLayer = this.viewer.createLayer('osm', params.layer);
-                    console.log({ overlayLayer });
+                    overlayLayer.id(overlay.id);
+                    this.viewer.scheduleAnimationFrame(this.viewer.draw, true);
                 }).fail((response) => {
-                    console.log({ response });
+                    console.error(`There was an error overlaying image with ID ${overlayItemId}`);
                 });
             });
             this._featureOpacity[annotation.id] = {};
@@ -428,6 +429,11 @@ var GeojsImageViewerWidgetExtension = function (viewer) {
                     } else {
                         this.featureLayer.deleteFeature(feature);
                     }
+                });
+                _.each(this._annotations[annotation.id].overlays, (overlay) => {
+                    const overlayLayer = this.viewer.layers().find(
+                        (layer) => layer.id() === overlay.id);
+                    this.viewer.deleteLayer(overlayLayer);
                 });
                 delete this._annotations[annotation.id];
                 delete this._featureOpacity[annotation.id];
--- a/girder_annotation/girder_large_image_annotation/web_client/views/imageViewerWidget/geojs.js
+++ b/girder_annotation/girder_large_image_annotation/web_client/views/imageViewerWidget/geojs.js
@@ -67,6 +67,9 @@ var GeojsImageViewerWidgetExtension = function (viewer) {
             const xOffset = transformInfo.xoffset || 0;
             const yOffset = transformInfo.yoffset || 0;
             const matrix = transformInfo.matrix || [[1, 0], [0, 1]];
+            if (xOffset === 0 && yOffset === 0 && matrix === [[1, 0], [0, 1]]) {
+                return '+proj=longlat +axis=enu';
+            }
             return `+proj=longlat +axis=enu +s11=${1 / matrix[0][0]} +s12=${matrix[0][1]}` +
                    ` +s21=${matrix[1][0]} +s22=${1 / matrix[1][1]} +xoff=-${xOffset} +yoff=${yOffset}`;
         },
--- a/girder_annotation/girder_large_image_annotation/web_client/views/imageViewerWidget/geojs.js
+++ b/girder_annotation/girder_large_image_annotation/web_client/views/imageViewerWidget/geojs.js
@@ -115,7 +115,7 @@ var GeojsImageViewerWidgetExtension = function (viewer) {
                     this.viewer.deleteLayer(overlayLayer);
                 });
             }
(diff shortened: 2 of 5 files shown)


manthey:
You'll probably need to merge or rebase on master for CI to have any chance of passing since Pillow 9.0.0 broke some of the tests.

manthey (changes requested):

girder_annotation/girder_large_image_annotation/web_client/annotations/convertFeatures.js
```diff
@@ -1,3 +1,5 @@
+// import { restRequest } from '@girder/core/rest';
```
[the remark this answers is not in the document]
naglepuff: Removed in 5b3f52b

girder_annotation/girder_large_image_annotation/web_client/views/imageViewerWidget/geojs.js
```diff
@@ -55,6 +56,21 @@ var GeojsImageViewerWidgetExtension = function (viewer) {
...
+         * a proj-string representation of its transform specification.
+         * @param {object} overlay An imageoverlay annotation element.
+         * @returns a proj-string representing how to overlay should be tranformed.
+         */
+        _getOverlayTransformProjString: function (overlay) {
+            const transformInfo = overlay.transform || {};
+            const xOffset = transformInfo.xoffset || 0;
+            const yOffset = transformInfo.yoffset || 0;
+            const matrix = transformInfo.matrix || [[1, 0], [0, 1]];
+            return `+proj=longlat +axis=enu +s11=${1 / matrix[0][0]} +s12=${matrix[0][1]}` +
```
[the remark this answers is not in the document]
naglepuff: Fixed in 5b3f52b

girder_annotation/girder_large_image_annotation/web_client/views/imageViewerWidget/geojs.js
```diff
@@ -190,6 +213,32 @@ var GeojsImageViewerWidgetExtension = function (viewer) {
                     }
                 });
             }
+            // draw overlays
+            if (this._annotations[annotation.id].overlays.length > 0) {
+                this.viewer.clampBoundsY(false);
+                this.viewer.clampBoundsX(false);
```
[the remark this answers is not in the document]
naglepuff: Let me know if my approach in dc210e3 works here. I agree that zoom out to recenter is very useful to have. Is there another approach to allowing image overlays  to be off the image without interfering with that functionality?

girder_annotation/girder_large_image_annotation/web_client/views/imageViewerWidget/geojs.js
```diff
@@ -190,6 +213,32 @@ var GeojsImageViewerWidgetExtension = function (viewer) {
...
+                const overlayItemId = overlay.girderId;
+                restRequest({
+                    url: `item/${overlayItemId}/tiles`
+                }).done((response) => {
+                    let params = geo.util.pixelCoordinateParams(
+                        this.viewer.node(), response.sizeX, response.sizeY, response.tileHeight, response.tileWidth
+                    );
+                    params.layer.useCredentials = true;
+                    params.layer.url = `api/v1/item/${overlayItemId}/tiles/zxy/{z}/{x}/{y}`;
+                    params.layer.autoshareRenderer = false;
```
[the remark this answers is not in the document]
naglepuff: I addressed this in 0bbbe08. Let me know if that is sufficient or if there's something I'm misunderstanding about this comment.

girder_annotation/girder_large_image_annotation/web_client/views/imageViewerWidget/geojs.js
```diff
@@ -402,8 +489,15 @@ var GeojsImageViewerWidgetExtension = function (viewer) {
...
                     }
                 });
+                _.each(this._annotations[annotation.id].overlays, (overlay) => {
+                    const overlayLayer = this.viewer.layers().find(
+                        (layer) => layer.id() === overlay.id);
+                    this.viewer.deleteLayer(overlayLayer);
+                });
                 delete this._annotations[annotation.id];
                 delete this._featureOpacity[annotation.id];
+                this.viewer.clampBoundsY(true);
```
manthey, line 499: I think we should reset to true if we didn't change it to false, so a downstream viewer could use whatever default they see fit if the flag is false.

naglepuff: 2cf5c3f

manthey:
I see two issues.  I placed a small image (one using the PIL tile source) as an overlay on a much larger image with an identity matrix specified as the transform.

Internally, I still see transform parameters being applied (`gcs: "+proj=longlat +axis=enu +s11=1 +s12=0 +s21=0 +s22=1 +xoff=-0 +yoff=0"` -- note the -0.

More importantly, though, the small image is magnified.  Fundamentally, the scale is incorrect, since the function that calculates the layer properties is making level 0 tiles appear at the same scale for the two sources.  This might be as simple as scaling down the transform matrix based on the number of tile levels between the two images, but it might be more complex than that.

naglepuff:
> Internally, I still see transform parameters being applied (gcs: "+proj=longlat +axis=enu +s11=1 +s12=0 +s21=0 +s22=1 +xoff=-0 +yoff=0" -- note the -0.

This is easy enough to fix. The issue lies with the check `matrix === [[1, 0], [0, 1]]`.

> More importantly, though, the small image is magnified. Fundamentally, the scale is incorrect, since the function that calculates the layer properties is making level 0 tiles appear at the same scale for the two sources. This might be as simple as scaling down the transform matrix based on the number of tile levels between the two images, but it might be more complex than that.

I think I see this as well. If I overlay a small image (~4k by 2k) onto a very large image (~90k by 90k), the small image is blown up well beyond the space it should take up. With my test images, the smaller has 6 zoom levels, the large has 10. How does this ratio inform how to appropriately scale the smaller image?

manthey:
GeoJS has a utility function (`pixelCoordinateParams`) that assumes the image will take up the whole map space.  The correct thing to do is to either have a different function where this assumption is not made, or where we adjust the layer parameter outputs of that function.

Alternately, we can probably twiddle the affine transform to make up for the fact that this isn't so -- by scaling the image by 2**(maxlevel of base image - maxlevel of overlayimage) to account for the number of tile levels difference and by (tile width of base image / tile width of overlay image), (tile height of base image / tile height of overlay image).  This will probably just work, but the extending or adjusting the results from pixelCoordinateParams would be "more correct" and probably by slightly more efficient.

I'll see if I can generalize this, as we've wanted it once before on a different project.

naglepuff:
> GeoJS has a utility function (pixelCoordinateParams) that assumes the image will take up the whole map space. The correct thing to do is to either have a different function where this assumption is not made, or where we adjust the layer parameter outputs of that function.

Which parameter outputs would need to be adjusted? `tilesAtZoom` and `tilesMaxBounds`?

> Alternately, we can probably twiddle the affine transform to make up for the fact that this isn't so

I've been fiddling around with this and it seems to be a decent workaround. However, I'd prefer to do something more "correct." I'm leaning towards adjusting the output of `pixelCoordinateParams`, as I don't want to increase the scope to include changes to `geojs` at this point. Let me know if you have any thoughts.

manthey:
So I dug through an old example I had where I did this somewhere else.  There is code below.  `item` is a an object with `id` as the Girder ID and `zoffset` the base image's max levels - overlay image's max levels.  The big other change is the `url` isn't just a template string, it is actually a function that returns the url.

```
    if (item.zoffset) {
      params.layer.url = (x, y, z) => 'api/v1/item/' + item.id + `/tiles/zxy/${z-item.zoffset}/${x}/${y}`;
      params.layer.minLevel = item.zoffset;
      params.layer.maxLevel += item.zoffset;
      params.layer.tilesAtZoom = (level) => {
        var scale = Math.pow(2, params.layer.maxLevel - level);
        return {
          x: Math.ceil(item.meta.sizeX / item.meta.tileWidth / scale),
          y: Math.ceil(item.meta.sizeY / item.meta.tileHeight / scale)
        };
      };
      params.layer.tilesMaxBounds = (level) => {
        var scale = Math.pow(2, params.layer.maxLevel - level);
        return {
          x: Math.floor(item.meta.sizeX / scale),
          y: Math.floor(item.meta.sizeY / scale)
        };
      };
    }
```

girder_annotation/girder_large_image_annotation/web_client/views/imageViewerWidget/geojs.js
```diff
@@ -85,11 +85,23 @@ var GeojsImageViewerWidgetExtension = function (viewer) {
...
             const matrix = transformInfo.matrix || [[1, 0], [0, 1]];
-            if (xOffset === 0 && yOffset === 0 && matrix === [[1, 0], [0, 1]]) {
-                return '+proj=longlat +axis=enu';
+            const s11 = matrix[0][0];
+            const s12 = matrix[0][1];
+            const s21 = matrix[1][0];
+            const s22 = matrix[1][1];
+
+            let projString = '+proj=longlat +axis=enu';
+            if (xOffset > 0) {
```
manthey, line 94: Is there a reason we constrain the offsets to be positive?

girder_annotation/girder_large_image_annotation/web_client/views/imageViewerWidget/geojs.js
```diff
@@ -55,6 +57,111 @@ var GeojsImageViewerWidgetExtension = function (viewer) {
...
+            params.layer.url = `api/v1/item/${overlayImageId}/tiles/zxy/{z}/{x}/{y}`;
+            if (this._countDrawnImageOverlays() <= 6) {
+                params.layer.autoshareRenderer = false;
+            } else {
+                params.layer.renderer = 'canvas';
+            }
+            params.layer.opacity = overlay.opacity || 1;
+
+            if (this.levels !== overlayImageMetadata.levels) {
+                const levelDifference = Math.abs(this.levels - overlayImageMetadata.levels);
```
manthey, line 145: This should NOT be `abs` -- it could be negative.

girder_annotation/girder_large_image_annotation/web_client/views/imageViewerWidget/geojs.js
naglepuff: In #678, `xOffset` and `yOffset` had a minimum of `0`. In rethinking that decision, I can't come up with a truly compelling reason for that constraint. There's likely a valid reason to render an overlay with its origin outside of the base image. To that end, I've pushed 86cc293

girder_annotation/girder_large_image_annotation/web_client/views/imageViewerWidget/geojs.js
```diff
@@ -55,6 +57,114 @@ var GeojsImageViewerWidgetExtension = function (viewer) {
...
+         * Generate layer parameters for an image overlay layer
+         * @param {object} overlayImageMetadata metadata such as size, tile size, and levels for the overlay image
+         * @param {string} overlayImageId ID of a girder image item
+         * @param {object} overlay information about the overlay such as opacity
+         * @returns layer params for the image overlay layer
+         */
+        _generateOverlayLayerParams(overlayImageMetadata, overlayImageId, overlay) {
+            const geo = window.geo;
+            let params = geo.util.pixelCoordinateParams(
+                this.viewer.node(), overlayImageMetadata.sizeX, overlayImageMetadata.sizeY, overlayImageMetadata.tileHeight, overlayImageMetadata.tileWidth
```
manthey, line 133: tileWidth and tileHeight are swapped in this call.

girder_annotation/girder_large_image_annotation/web_client/views/imageViewerWidget/geojs.js
naglepuff: Fixed in a1743d5

manthey (approved):

PR #752, opened 2022-01-07
Include user, group, and label in the imageoverlay schema.

manthey:
This also makes the spacing in the annotation schema docs more consistent.

### DIFF
--- a/girder_annotation/docs/annotations.rst
+++ b/girder_annotation/docs/annotations.rst
@@ -16,16 +16,16 @@ the comments are not part of a valid annotation:
 
 ::
 
-   {
-     "name": "MyAnnotationName",              # Non-empty string.  Optional
-     "description": "This is a description",  # String.  Optional
-     "attributes": {                          # Object.  Optional
-       "key1": "value1",
-       "key2": ["any", {"value": "can"}, "go", "here"]
-     },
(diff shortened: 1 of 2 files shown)


girder_annotation/docs/annotations.rst
```diff
@@ -278,21 +280,22 @@ a ``2x2`` affine matrix.
...
-            "yoffset": 0,                 # How much to shift the overlaid image down.
-            "matrix": [                   # Affine matrix to specify tranformations like scaling,
-                                          # rotation, or shearing.
-                [1, 0],
-                [0, 1]
-            ]
-        }
+  {
+    "type": "imageoverlay",            # Exact string. Required
+    <id, label>                        # Optional general shape properties
```
naglepuff, line 285: Does it make sense to keep consistent with other element types here?

```suggestion
    <id, label, group, user>           # Optional general shape properties
```

naglepuff (changes requested):

girder_annotation/docs/annotations.rst
manthey: Yes.  In my mind I had updated that line, but not in reality.

naglepuff (approved):

PR #755, opened 2022-01-10
Support discrete palettes for all tile sources.

### LINKED ISSUE #753: Categorical color palettes
If passing two HEX colors as the palette, the returned thumbnail or tiles interpolate the values between those two colors. 

Is there a way to use the color maps as categorical colors? e.g., split the min/max range into the number of colors in the palette and assign those values directly to the given colors.

manthey:
Closes #753.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -4,12 +4,16 @@
 
 ### Features
 - Added image overlay annotation element ([#742](../../pull/742))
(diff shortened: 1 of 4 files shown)


banesullivan (approved):
Excellent! Thanks, @manthey!

"discrete" is a good choice IMO

PR #761, opened 2022-01-21
Add additional behaviors to image overlays

naglepuff:
This PR rounds out some functionality for image overlays that make them more reactive in applications like [HistomicsUI](https://github.com/DigitalSlideArchive/HistomicsUI).

Specifically, this PR introduces the following changes:
- Image overlay layers now properly respond to the `globalAnnotationOpacity` of the image viewer. As a result, image overlays now change opacity based on the slider in HistomicsUI.
- Image overlays now respect the currently highlighted annotation. Image overlay elements of non-highlighted annotations will reduce their opacity by 75% like other element types. This can be demonstrated in HUI by selecting "Interactive" in the annotations widget and hovering over different annotations in the list.
- Schema for image overlays now allows for additional properties. This fixes an issue where editing an annotation (e.g. via HUI) would cause an error when trying to save, since the `id` property would be added to the overlay elements.
-  `drawAnnotation` has been updated to account for the case where an image overlay element was removed from the annotation between an initial call to `drawAnnotation` and a follow up call to `drawAnnotation`.

### DIFF
--- a/girder_annotation/girder_large_image_annotation/models/annotation.py
+++ b/girder_annotation/girder_large_image_annotation/models/annotation.py
@@ -508,7 +508,7 @@ class AnnotationSchema:
             'group': groupSchema,
         },
         'required': ['girderId', 'type'],
-        'additionalProperties': False,
+        'additionalProperties': True,
         'description': 'An image to overlay onto another like an '
                        'annotation.'
     }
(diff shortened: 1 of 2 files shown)


girder_annotation/girder_large_image_annotation/models/annotation.py
```diff
@@ -508,7 +508,7 @@ class AnnotationSchema:
             'group': groupSchema,
         },
         'required': ['girderId', 'type'],
-        'additionalProperties': False,
+        'additionalProperties': True,
```
manthey, line 511: Probably the correct thing to do is to make the overlay annotation inherit from the baseShapeSchema, except that we don't want to get the line and lineWidth properties.  We can do this in a separate PR -- baseShapeSchema would inherit from baseElementSchema which would exclude those properties and then the overlay annotation would inherit from that.

manthey (approved):
This looks right to me.  We should make a new issue or PR to adjust the schema in a more intelligent manner.

girder_annotation/girder_large_image_annotation/models/annotation.py
naglepuff: Created #762

PR #763, opened 2022-01-24
Update and make the annotation schema more uniform.

### LINKED ISSUE #762: Allow image overlay schema to contain additional properties such as ID
Probably the correct thing to do is to make the overlay annotation inherit from the baseShapeSchema, except that we don't want to get the line and lineWidth properties.  We can do this in a separate PR -- baseShapeSchema would inherit from baseElementSchema which would exclude those properties and then the overlay annotation would inherit from that.

_Originally posted by @manthey in https://github.com/girder/large_image/pull/761#discussion_r791006836_

manthey:
Resolves #762.

### DIFF
--- a/girder_annotation/girder_large_image_annotation/models/annotation.py
+++ b/girder_annotation/girder_large_image_annotation/models/annotation.py
@@ -14,6 +14,7 @@
 #  limitations under the License.
 ##############################################################################
 
+import copy
 import datetime
 import enum
 import re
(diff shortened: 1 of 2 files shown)


naglepuff:
I'm having some trouble with these changes. Here's an annotation that works (can be successfully uploaded) in `master` but not in `adjust-schema`:

```
{
	"name": "overlay_test",
	"description": "",
	"elements": [
		{
			"type": "imageoverlay",
			"girderId": "61b243b9b9de8887df5cd186",
			"opacity": 0.5,
			"group": "myGroup"	
		}
	]
}
```
In this case its the `group` property that appears to be the cause of failure. This seems to be the case for `label`, `user`, and `id`.

PR #764, opened 2022-01-26
Add multi-source source

manthey:
This adds a source that can be composited from other sources.

This is still a work-in-progress, needing some features and tests.  It *should* work except for crop and any transforms where s12 or 21 is non-zero or s11 or s22 is less than zero.

There are several ways that the same set of files can be specified.  For instance, given a set of eight files named `test_orient1.tif`, `test_orient2.tif`, ..., `test_orient8.tif`, to make a source that uses all of these files, the following yaml files are equivalent:

- (1) explicitly specify each frame and a z value
```
sources:
  - path: ./test_orient1.tif
    z: 0
  - path: ./test_orient2.tif
    z: 1
  - path: ./test_orient3.tif
    z: 2
  - path: ./test_orient4.tif
    z: 3
  - path: ./test_orient5.tif
    z: 4
  - path: ./test_orient6.tif
    z: 5
  - path: ./test_orient7.tif
    z: 6
  - path: ./test_orient8.tif
    z: 7
```
- (2) use a regular expression to select frames (they are added alphabetically by c-sort order) and specify a step value
```
sources:
  - path: .
    pathPattern: 'test_orient[1-8]\.tif'
    zStep: 1
```
- (3) use a named regular expression to populate the z value.  If the name ends in 1, it is assumed to be 1-indexed rather than 0-indexed.
```
sources:
  - path: .
    pathPattern: 'test_orient(?P<z1>[1-8])\.tif'
```

### DIFF
--- a/.circleci/make_wheels.sh
+++ b/.circleci/make_wheels.sh
@@ -31,6 +31,8 @@ cd "$ROOTPATH/sources/gdal"
 pip wheel . --no-deps -w ~/wheels && rm -rf build
 cd "$ROOTPATH/sources/mapnik"
 pip wheel . --no-deps -w ~/wheels && rm -rf build
+cd "$ROOTPATH/sources/multi"
+pip wheel . --no-deps -w ~/wheels && rm -rf build
 cd "$ROOTPATH/sources/nd2"
 pip wheel . --no-deps -w ~/wheels && rm -rf build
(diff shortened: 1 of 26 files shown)


manthey:
Merging this as it works in basic testing and to make it easier to use in other projects.

PR #767, opened 2022-01-28
Add schema for tiled pixelmap overlays

naglepuff:
This PR includes the following changes:

- Update the girder large image annotation schema to include `tiledpixelmap` elements
- Calculate bounding box for `tiledpixelmap` elements using the same technique as `imageoverlay` elements
- Update documentation to reflect schema changes.

### DIFF
--- a/girder_annotation/girder_large_image_annotation/models/annotation.py
+++ b/girder_annotation/girder_large_image_annotation/models/annotation.py
@@ -437,6 +437,63 @@ class AnnotationSchema:
         'description': 'An image overlay on top of the base resource.',
     })
 
+    segmentsToColorSchema = {
+        'type': 'array',
+        'items': colorSchema,
+        'description': 'An array where the indices correspond to pixel '
+                       'values in the pixel map image, and the values '
+                       'correspond to colors.'
+    }
+
+    categoryColorMap = {
+        'type': 'object',
+        'properties': {
+            'segments': {
+                'type': 'array',
+                'items': {'type': 'string'},
+                'description': 'An array where the indices correspond to '
+                               'pixel values in the pixel map image, and the '
+                               'values correspond to key values in the '
+                               'colormap object.'
+            },
+            'colormap': {
+                'type': 'object',
+                'additionalProperties': colorSchema,
+                'description': 'An object whose keys represent category '
+                               'names, and whose values are color strings. '
+                               'This map is used together with the segments '
+                               'array to correctly color regions of the '
+                               'pixelmap.'
+            }
+        },
+        'additionalProperties': False,
+        'required': ['segments', 'colormap'],
+    }
+
+    pixelMapDataSchema = {
+        'oneOf': [
+            segmentsToColorSchema,
+            categoryColorMap,
+        ],
+    }
+
+    tiledPixelMapSchema = extendSchema(overlaySchema, {
+        'properties': {
+            'type': {
+                'type': 'string',
+                'enum': ['tiledpixelmap'],
+            },
+            'data': pixelMapDataSchema,
+            'boundaries': {
+                'type': 'boolean',
+                'description': 'True if the boundaries of the pixelmap regions have '
+                               'their own value in the data or segments array.'
+            },
+        },
+        'required': ['data', 'boundaries'],
+        'description': 'A tiled pixelmap to overlay onto a base resource.'
+    })
+
     annotationElementSchema = {
         # Shape subtypes are mutually exclusive, so for efficiency, don't use
         # 'oneOf'
@@ -454,6 +511,7 @@ class AnnotationSchema:
             rectangleShapeSchema,
             rectangleGridShapeSchema,
             overlaySchema,
+            tiledPixelMapSchema,
         ]
     }
 
--- a/girder_annotation/docs/annotations.rst
+++ b/girder_annotation/docs/annotations.rst
@@ -294,6 +294,74 @@ a ``2x2`` affine matrix.
     }
   }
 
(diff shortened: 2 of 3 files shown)


jeffbaumes:
One thought is on naming here. Could we simplify to “image” and “pixelmap”? If the tiled nature is important to call out, maybe “tiledimage” and “tiledpixelmap” for consistency?

girder_annotation/girder_large_image_annotation/models/annotation.py
```diff
@@ -437,6 +437,63 @@ class AnnotationSchema:
...
+        'items': colorSchema,
+        'description': 'An array where the indices correspond to pixel '
+                       'values in the pixel map image, and the values '
+                       'correspond to colors.'
+    }
+
+    categoryColorMap = {
+        'type': 'object',
+        'properties': {
+            'segments': {
```
manthey, line 451: I think this would be better as 
```
'values': {
    'type': 'array',
    'items': {'type': 'integer'},
}
```
Using `values` for consistency with other annotations, and storing just indices into the category map rather than strings.  Since we could have millions of pixels, using integers here will be more efficient.

The `colormap` might conceptually be better called `categories` where each entry has a color, a label string, and an optional description to give a more verbose description of the category.

For pixelmaps with boundaries, perhaps this value array should be half as long as the actual number of number of distinct indices and then the category would have a `fillColor` and a `strokeColor` (and for non-boundaries pixelmaps, this would just use the `fillColor`).

naglepuff: I've implemented these changes in 47c74f4.

I've also removed having two representations of the `values` array. Let me know if there are any further refinements to be made.

girder_annotation/girder_large_image_annotation/models/annotation.py
```diff
@@ -401,7 +401,7 @@ class AnnotationSchema:
         'properties': {
             'type': {
                 'type': 'string',
-                'enum': ['imageoverlay']
+                'enum': ['tiledimage']
             },
```
manthey, line 405: @naglepuff, @jeffbaumes has convinced me that we should just call these annotations `image` and `pixelmap`.  When we call `geo.util.pixelCoordinateParams`, the only difference there is whether a tile size is passed to the function, and a difference in the url having zxy expansions, so it should be relatively easy to make the viewer agnostic to whether the underlying image is tiled or not.

girder_annotation/girder_large_image_annotation/models/annotation.py
```diff
@@ -679,6 +733,7 @@ def _migrateACL(self, annotation):
         return annotation
 
     def createAnnotation(self, item, creator, annotation, public=None):
+        print('\nCREATING ANNOTATION\n')
```
manthey, line 736: If you want to keep log messages, you should be able to use `logger.debug` (or another appropriate level).  For development, I'll adjust the level to info and then set it to debug once I think it is stable.

naglepuff: Removed in 24115d5

girder_annotation/girder_large_image_annotation/models/annotation.py
naglepuff: Renamed in 24115d5

girder_annotation/girder_large_image_annotation/models/annotation.py
```diff
@@ -437,6 +437,59 @@ class AnnotationSchema:
...
+                'type': 'array',
+                'items': pixelMapCategorySchema,
+                'description': 'An array used to map between the '
+                               'values array and color values. '
+                               'Can also contain semantic '
+                               'information for color values.'
+            },
+            'boundaries': {
+                'type': 'boolean',
+                'description': 'True if the boundaries of the pixelmap regions have '
```
manthey, line 485: I don't think this description is right any more.  Since we've added strokeColor into the category, this should probably read something on order of "True if the pixelmap doubles index values such that even values are the fill and odd values are the stroke of each superpixel.  In this case, the values array will be half the length of the actual index values in the image".

girder_annotation/girder_large_image_annotation/models/annotation.py
```diff
@@ -437,6 +437,59 @@ class AnnotationSchema:
...
+                               'their own value in the data or segments array.'
+            },
+        },
+        'required': ['values', 'categories', 'boundaries'],
+        'description': 'A tiled pixelmap to overlay onto a base resource.'
```
manthey, line 490: Please add `'additionalProperties': False` to this and the pixelmapCategorySchema.

girder_annotation/girder_large_image_annotation/models/annotation.py
naglepuff: Updated description in 95bab98

girder_annotation/girder_large_image_annotation/models/annotation.py
naglepuff: Added in 95bab98

girder_annotation/girder_large_image_annotation/models/annotation.py
```diff
@@ -437,6 +437,59 @@ class AnnotationSchema:
         'description': 'An image overlay on top of the base resource.',
     })
 
+    pixelMapCategorySchema = {
```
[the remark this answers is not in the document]
naglepuff: Fixed in 95bab98

manthey (approved):
Thanks for putting up with the requested changes.

PR #775, opened 2022-02-08
Improve the import time.

manthey:
Some libraries: palettable and pylibmc account for a substantial import time.  Loaded them as needed rather than on first import.


manthey:
Curiously, after this change, the slowest import in the package is `pkg_resources`, whcih we use to determine version.  We could switch to `importlib` for this purpose and shave off another 30-40 ms on initial import.  When all the sources are loaded, the slow points are the nd2 and bioformat sources.  Delaying imports on those is probably worth the speed improvement, though it would allow an installation without those packages to imply that those sources are available.

PR #776, opened 2022-02-08
Support pixelmap annotations

naglepuff:
This PR updates the `large_image` plugin to support drawing `pixelmap` annotations (see #767).

### DIFF
--- a/girder_annotation/girder_large_image_annotation/web_client/views/imageViewerWidget/geojs.js
+++ b/girder_annotation/girder_large_image_annotation/web_client/views/imageViewerWidget/geojs.js
@@ -121,6 +121,44 @@ var GeojsImageViewerWidgetExtension = function (viewer) {
         },
 
         /**
+         * Set additional parameters for pixelmap overlays.
+         * @param {object} layerParams An object containing layer parameters. This should already have
+         * generic properties for overlay annotations set, such as the URL, opacity, etc.
+         * @param {object} pixelmapElement A pixelmap annotation element
+         * @returns An object containing parameters needed to create a pixelmap layer.
+         */
+        _addPixelmapLayerParams(layerParams, pixelmapElement) {
+            // For pixelmap overlays, there are additional parameters to set
+            layerParams.keepLower = false;
+            layerParams.url = layerParams.url + `?encoding=PNG`;
+            let pixelmapData = pixelmapElement.values;
+            if (pixelmapElement.boundaries) {
+                let valuesWithBoundaries = new Array(pixelmapData.length * 2).fill(0);
+                pixelmapData = valuesWithBoundaries.map((d, i) => pixelmapData[Math.floor(i / 2)]);
+            }
+            layerParams.data = pixelmapData;
+            layerParams.style = {
+                color: (d, i) => {
+                    const categoryMap = pixelmapElement.categories;
+                    const boundaries = pixelmapElement.boundaries;
+                    if (d < 0 || d > categoryMap.length) {
+                        console.warn(`No category found at index ${d} in the category map.`);
+                        return 'rgba(0, 0, 0, 0)';
+                    }
+                    let color;
+                    let category = categoryMap[d];
+                    if (boundaries) {
+                        color = (i % 2 === 0) ? category.fillColor : category.strokeColor;
+                    } else {
+                        color = category.fillColor;
+                    }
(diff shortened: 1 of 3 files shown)


girder_annotation/girder_large_image_annotation/web_client/views/imageViewerWidget/geojs.js
```diff
@@ -120,6 +120,43 @@ var GeojsImageViewerWidgetExtension = function (viewer) {
...
+         * @param {object} pixelmapElement A pixelmap annotation element
+         * @returns An object containing parameters needed to create a pixelmap layer.
+         */
+        _addPixelmapLayerParams(layerParams, pixelmapElement) {
+            // For pixelmap overlays, there are additional parameters to set
+            layerParams.url = layerParams.url + `?encoding=PNG`;
+            let pixelmapData = pixelmapElement.values;
+            if (pixelmapElement.boundaries) {
+                let valuesWithBoundaries = new Array(pixelmapData.length * 2).fill(0);
+                pixelmapData = valuesWithBoundaries.map((d, i) => pixelmapData[Math.floor(i / 2)]);
```
manthey, line 136: You aren't using the values in the valuesWithBoundaries array.  So `fill` is a wasted operation.  This would probably be better as:
```
let valuesWithBoundaries = new Array(pixelmapData.length * 2);
for (let i = 0; i < pixelmapData.length; i += 1) {
  valuesWithBoundaries[i * 2] = valuesWithBoundaries[i * 2 + 1] = pixelmapData[i];
}
```
Function calls tend to be slow, so avoiding `map`ping something that is potentially millions long is preferred.

girder_annotation/girder_large_image_annotation/web_client/views/imageViewerWidget/geojs.js
```diff
@@ -129,7 +129,6 @@ var GeojsImageViewerWidgetExtension = function (viewer) {
          */
         _addPixelmapLayerParams(layerParams, pixelmapElement) {
             // For pixelmap overlays, there are additional parameters to set
-            layerParams.keepLower = false;
```
manthey: I think we need this set to false, since otherwise you'll get off effects if the superpixels are transparent.

I think this was fixed in https://github.com/OpenGeoscience/geojs/commit/7ea49772a89031353aacacc2387e33f7e0bf14a6.  Are you on the master build of geojs?

naglepuff: `npm` is telling me that I have `geojs@1.7.1`, which looks like the latest release.

With `layerParams.keepLower = false`, I was seeing a problem in HisotmicsUI where zooming in past the number of levels would cause the pixelmap to disappear until zoomed out again. For example, in my testing I was working with a whole slide image with magnification 20, so HUI allows zooming in to 40x. The pixelmap layer would not be visible when zooming between 20x and 40x.

Without setting `keepLower = false`, I did not see that issue. I didn't see any effects (not sure what I should be looking out for though). Is there a better way to allow this kind of zooming while keeping the layer visible?

manthey: You should obviously see the issue if the fill color is transparent and the stroke color is opaque.  We either need to change some parameters or there is a bug in GeoJS.  I'll investigate.

manthey: There will be a PR in GeoJS shortly.

naglepuff: I see the issue now. I'll go ahead and add that line back. I'm still not sure how to deal with the issue of zooming "past" the pixelmap layer that I'm seeing in HistomicsUI.

manthey: https://github.com/OpenGeoscience/geojs/pull/1178.  I'll merge when it passes CI and then it will be available.

naglepuff: Looks like that PR did the trick. I've added `keepLower = false` back in 65ae579

manthey (approved):

PR #787, opened 2022-02-22
Emit event for overlay annotations

naglepuff:
This PR makes the following changes:

1. Emit an event of type `g:drawOverlayAnnotation` when a GeoJS layer is created for an image overlay. This is useful for code that uses `large_image`, such as `HistomicsUI`, since creating these layers happens asynchronously.
2. Fix a bug where the URL for annotations of type `pixelmap` would be malformed if it had a different number of zoom levels compared to the base image layer.

### DIFF
--- a/girder_annotation/girder_large_image_annotation/web_client/views/imageViewerWidget/geojs.js
+++ b/girder_annotation/girder_large_image_annotation/web_client/views/imageViewerWidget/geojs.js
@@ -125,12 +125,17 @@ var GeojsImageViewerWidgetExtension = function (viewer) {
          * @param {object} layerParams An object containing layer parameters. This should already have
          * generic properties for overlay annotations set, such as the URL, opacity, etc.
          * @param {object} pixelmapElement A pixelmap annotation element
+         * @param {number} levelDifference The difference in zoom level between the base image and the overlay
          * @returns An object containing parameters needed to create a pixelmap layer.
          */
-        _addPixelmapLayerParams(layerParams, pixelmapElement) {
+        _addPixelmapLayerParams(layerParams, pixelmapElement, levelDifference) {
             // For pixelmap overlays, there are additional parameters to set
             layerParams.keepLower = false;
-            layerParams.url = layerParams.url + `?encoding=PNG`;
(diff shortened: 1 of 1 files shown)


girder_annotation/girder_large_image_annotation/web_client/views/imageViewerWidget/geojs.js
```diff
@@ -125,12 +125,17 @@ var GeojsImageViewerWidgetExtension = function (viewer) {
...
          * @param {object} pixelmapElement A pixelmap annotation element
+         * @param {number} levelDifference The difference in zoom level between the base image and the overlay
          * @returns An object containing parameters needed to create a pixelmap layer.
          */
-        _addPixelmapLayerParams(layerParams, pixelmapElement) {
+        _addPixelmapLayerParams(layerParams, pixelmapElement, levelDifference) {
             // For pixelmap overlays, there are additional parameters to set
             layerParams.keepLower = false;
-            layerParams.url = layerParams.url + `?encoding=PNG`;
+            if (levelDifference > 0) {
```
manthey, line 134: Why is this necessary since if this is non-zero the url would have already been `'api/v1/item/' + overlayImageId +` `/tiles/zxy/${z - levelDifference}/${x}/${y}` based on the code in `_generateOverlayLayerParam`?  The only difference I see is a `/` before the `?encoding=PNG`

naglepuff: In the case that `levelDifference` is non-zero, `layerParams.url` is a function, so concatenating the `?encoding=PNG` wasn't working in that case. 

Is there a better way to do change the function? 

I've recently pushed 14d38de to make sure the `levelDifference` check passes if it is non-zero, not just positive, and removed those extra `/` characters from the url.

manthey: Probably we should eventually abstract this to have the GeoJS utility functions generate appropriate urls directly based on the level difference.  For now, this is fine.

manthey (approved):

PR #794, opened 2022-03-10
Add mouse events to overlay layers

naglepuff:
Similar to how annotation feature elements, hook up overlay elements to mouse events.


manthey (approved):

naglepuff:
@manthey I added a test. Could you take a look at it when you get a chance?

PR #798, opened 2022-03-15
Add options to threshold near min/max based on the histogram.

manthey:
Instead of using a style range of `min` or `max`, you can now also use `min:<value>` or `max:<value>`.  This excludes a portion of the known values based on the computed histogram.  For instance, `min:0.02` would exclude the dimmest 2% of values.


manthey:
As an example:
style={"min":"min","max":"max"}
image
style={"min":"min:0.0002","max":"max:0.0002"}
image
style={"min":"min:0.002","max":"max:0.002"}
image
style={"min":"min:0.02","max":"max:0.02"}
image
style={"min":"min:0.2","max":"max:0.2"}
image

The first image using the actual min and max doesn't have much detail because a single bright point forces the rest of the image to be dark.  The second image uses a very conservative 0.02% threshold, and has better detail most places, though technically loses a trivial amount of detail around the bright point.  The third through fifth images successively threshold at 0.2%, 2% and 20%, and more features in the mid range are visible and more data is lost in the shadows and highlights.

PR #799, opened 2022-03-16
Change `libtiff` dependency to `pylibtiff`

mcovalt:
This PR updates the package name for [pearu/pylibtiff](https://github.com/pearu/pylibtiff).

It seems the official PyPI project name for [pearu/pylibtiff](https://github.com/pearu/pylibtiff) changed from [`libtiff`](https://pypi.org/project/libtiff) to [`pylibtiff`](https://pypi.org/project/pylibtiff) as of October 21, 2021.

### DIFF
--- a/sources/tiff/setup.py
+++ b/sources/tiff/setup.py
@@ -44,7 +44,7 @@ setup(
     ],
     install_requires=[
         'large-image',
-        'libtiff>=0.4.1',
+        'pylibtiff>=0.4.1',
         'tifftools>=1.2.0',
         'importlib-metadata ; python_version < "3.8"',
     ],
--- a/sources/tiff/setup.py
+++ b/sources/tiff/setup.py
@@ -44,7 +44,7 @@ setup(
     ],
     install_requires=[
         'large-image',
-        'pylibtiff>=0.4.1',
+        'pylibtiff',
         'tifftools>=1.2.0',
         'importlib-metadata ; python_version < "3.8"',
     ],


manthey:
We'll need to update the large_image_wheels to emit libtiff as pylibtiff before we merge this.

banesullivan:
Drive-by comment that the conda-forge recipe should be good here (no changes needed): https://github.com/conda-forge/large-image-source-tiff-feedstock/blob/main/recipe/meta.yaml

mcovalt:
> Drive-by comment that the conda-forge recipe should be good here (no changes needed): https://github.com/conda-forge/large-image-source-tiff-feedstock/blob/main/recipe/meta.yaml

Awesome. I think after this PR you could drop the `libtiff` requirement there? Or is `libtiff` there referring to the actual `libtiff` library?

sources/tiff/setup.py
```diff
@@ -44,7 +44,7 @@ def prerelease_local_scheme(version):
     ],
     install_requires=[
         'large-image',
-        'libtiff>=0.4.1',
+        'pylibtiff>=0.4.1',
```
manthey, line 47: We should probably remove the version limiter.  `pylibtiff` was only published as 0.4.4 (and our wheel which uses an artificially higher number to get priority).

mcovalt: Resolved in 42167f341a4286b0d0bcce0ae8aa4ab0ab844d84.

banesullivan:
> Or is libtiff there referring to the actual libtiff library?

Yes, libtiff is referring to the libtiff library while pylibtiff are the Python wrappings. That is still necessary in the conda recipe

PR #800, opened 2022-03-17
Use simplejpeg for jpeg encoding rather than PIL.

manthey:
This shaves more than 5 ms off of typical tile generation.  In one test, average tile generation time dropped from 17.4 ms to 12.1 ms.  This was with concurrent rest requests to Girder; the time was roughly twice this without concurrent requests.

This can be included with `pip install large_image[performance]` (or, of course `large_image[all]`).

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -9,7 +9,8 @@
 - Add options to threshold near min/max based on the histogram ([#798](../../pull/798))
 - Mark vsi extensions as being preferred by the bioformats source ([#793](../../pull/793))
 - Add mouse events to overlay annotations ([#794](../../pull/794))
-- Use orjson instead of ujson for annotationss ([#802](../../pull/802))
+- Use orjson instead of ujson for annotations ([#802](../../pull/802))
+- Use simplejpeg for jpeg encoding rather than PIL ([#800](../../pull/800))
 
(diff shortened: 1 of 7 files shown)


banesullivan:
Do you know what the typical response time is with PNG encoding? Simply curious

manthey:
With the default compress_level (6), the same benchmark on the same server averaged 18.2 ms per tile.  With a compress_level of 2, it averaged 15.5 ms.  Probably if we want to really speed up PNG, switching to mtpng would be the way to go, but we'd have to write a python wrapper for it (it is a rust library).

The benchmark I ran fires multiple concurrent requests into a running Girder system on a somewhat weak system.  When requests are sent sequentially, the average time per is somewhat worse that twice the time per tile listed here.  Once there are three concurrent requests, the speed seems to get no better, which seems surprising, though there is probably locking on the tile source that prevents speed ups.  Specifically, I fetch every tile (28895) from a specific svs file where all tiles end up getting encoded (whereas some other formats we can pass through at least some of the tiles without reencoding).

manthey:
And, for completeness, with a warm cache where we don't have to recompress anything, the average tile serving time from Girder is on the order of 5.3 ms (this is pulling image data from memcached and serving it over a socket).

PR #816, opened 2022-04-07
Add a vips tilesource

manthey:
This class can be used to write files as well.  This is a new capability and should be regarded as somewhat in flux.  Specifically, you can do
```
import large_image_source_vips

newsrc = large_image_source_vips.new()
for chunk, x, y in chunk_generator: # some process that generates numpy arrays or images
    newsrc.addTile(chunk, x, y)
newsrc.write('output.tiff')
```


manthey:
Eventually, this should be generalized, so that there is a `large_image.new(<some criterion>)` method that would choose a sink based on some feature and could emit multiple file formats.  There are TODOs listed in the source for this for other ways to extend it.

PR #817, opened 2022-04-08
Support style dictionary

### LINKED ISSUE #730: Allow `dict` for `style`
Currently, I think only JSON strings work when passing a `style` to `large_image.open()` -- could we support `dict`s as well? I'm not sure how this is being used under the hood, but perhaps this would be as simple as adding:

```py
if isinstance(style, dict):
    style = json.dumps(style)
```

banesullivan:
Resolves #730

I changed a few tests at random to use a dict instead of doing `json.dumps` to make sure this is tested

### DIFF
--- a/large_image/tilesource/base.py
+++ b/large_image/tilesource/base.py
@@ -126,12 +126,12 @@ class TileSource:
         self.jpegSubsampling = int(jpegSubsampling)
         self.tiffCompression = tiffCompression
         self.edge = edge
+        self._jsonstyle = style
         if style:
             if isinstance(style, dict):
                 self.style = style
                 self._jsonstyle = json.dumps(style, sort_keys=True, separators=(',', ':'))
             else:
-                self._jsonstyle = style
                 try:
                     self.style = json.loads(style)
                     if not isinstance(self.style, dict):
--- a/large_image/tilesource/base.py
+++ b/large_image/tilesource/base.py
@@ -128,12 +128,15 @@ class TileSource:
         self.edge = edge
         self._jsonstyle = style
         if style:
-            try:
-                self.style = json.loads(style)
-                if not isinstance(self.style, dict):
-                    raise TypeError
-            except TypeError:
-                raise exceptions.TileSourceError('Style is not a valid json object.')
+            if isinstance(style, dict):
+                self.style = style
+            else:
+                try:
+                    self.style = json.loads(style)
+                    if not isinstance(self.style, dict):
+                        raise TypeError
+                except TypeError:
+                    raise exceptions.TileSourceError('Style is not a valid json object.')
 
     def __repr__(self):
         return self.getState()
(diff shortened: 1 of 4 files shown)


large_image/tilesource/base.py
```diff
@@ -128,12 +128,15 @@ def __init__(self, encoding='JPEG', jpegQuality=95, jpegSubsampling=0,
         self.edge = edge
         self._jsonstyle = style
```
manthey, line 129: We use this as part of the caching key.  If this is a dict, then it has to serialize the dict each time.  I think the resolution would be to change
```
            if isinstance(style, dict):
                self.style = style
```
add
```
            if isinstance(style, dict):
                self.style = copy.deepcopy(style)
                self._jsonstyle = json.dumps(style, sort_keys=True, separators=(',', ':'))
```
I've added a deepcopy to the dictionary, as otherwise the caller could alter the style dynamically which would probably break things.

banesullivan: Push a fix for this

large_image/tilesource/base.py
```diff
@@ -126,14 +126,18 @@ def __init__(self, encoding='JPEG', jpegQuality=95, jpegSubsampling=0,
         self.jpegSubsampling = int(jpegSubsampling)
         self.tiffCompression = tiffCompression
         self.edge = edge
-        self._jsonstyle = style
```
manthey, line 129: This needs to be set even if `not style`.

banesullivan: 🤦🏻‍♂️  added this back

manthey (approved):

manthey:
My one concern is with tile source caching.  Currently, the style parameter is used for caching -- if the dictionary is identical enough, then it works as expected, but if the style dictionary differs, the caching may be less than expected (e.g., have equivalent dictionaries whose keys are differently ordered).  I'm wondering if it would be worth LRU hashing function so that if style is a dictionary it is serialized to json with sorted keys rather than just used as a python repr.

Or, maybe we don't care about efficiency in that case because a client is unlikely to reorder a style dictionary between tile source accesses.

banesullivan:
>Or, maybe we don't care about efficiency in that case because a client is unlikely to reorder a style dictionary between tile source accesses.

I think it's an unlikely edge case that the dictionary keys will be differently ordered between requests

> (e.g., have equivalent dictionaries whose keys are differently ordered). I'm wondering if it would be worth LRU hashing function so that if style is a dictionary it is serialized to json with sorted keys rather than just used as a python repr.

Should we just convert the dict to an `OrderedDict`?: https://docs.python.org/3/library/collections.html#collections.OrderedDict

and perhaps also sort the JSON blob as well?

manthey:
When we generate the JSON blob from the dict, we are ordering it.  As of Python 3.6, all dictionaries are ordered .  I think I'll merge this -- if anyone ever mutates the order of the keys in the dict and notices caching misses, we'll handle it then.

PR #818, opened 2022-04-08
Add `geospatial` property

### LINKED ISSUE #805: Not all GDAL sources are geospatial
GDAL can handle a wide variety of image formats, and there is no requirement for the images of some formats to be geospatial. This line can lead to some wacky behavior when trying to handle tile serving or region extraction (wacky behavior downstream rather than in large image itself as I do `if source.geospatial: do stuff` in my code)

https://github.com/girder/large_image/blob/0ccb63390c463d3cd6beaf993a2e7dd14f352126/sources/gdal/large_image_source_gdal/__init__.py#L94

Could we change this to a `property` that does some sort of check to make sure the source is indeed geospatial? I have some non-geospatial images that are only readable by the GDAL source that I have shared as the `IHTest_202009_Path3_Step5_BBXSWIR_12deg_DistStA` file pair here: https://data.kitware.com/#user/5edea5859014a6d84eabf0f0/folder/61809f8b2fa25629b904edc7

banesullivan:
Work in progress to resolve #805 

(using CI to test since I can't get the test env going locally at the moment)

### DIFF
--- a/sources/gdal/large_image_source_gdal/__init__.py
+++ b/sources/gdal/large_image_source_gdal/__init__.py
@@ -601,22 +601,21 @@ class GDALFileTileSource(FileTileSource, metaclass=LruCacheMetaclass):
             self._bandInfo = infoSet
         return self._bandInfo
 
-    # @property
-    # def geospatial(self):
-    #     return bool(
-    #         self.dataset.GetProjection() or
-    #         (self.dataset.GetGCPProjection() and self.dataset.GetGCPs()) or
-    #         self.dataset.GetGeoTransform(can_return_null=True) or
-    #         hasattr(self, '_netcdf'))
+    @property
+    def geospatial(self):
+        """
+        This is true if the source has geospatial information.
+        """
+        return bool(
+            self.dataset.GetProjection() or
+            (self.dataset.GetGCPProjection() and self.dataset.GetGCPs()) or
+            self.dataset.GetGeoTransform(can_return_null=True) or
+            hasattr(self, '_netcdf'))
 
     def getMetadata(self):
         with self._getDatasetLock:
             metadata = {
-                'geospatial': bool(
-                    self.dataset.GetProjection() or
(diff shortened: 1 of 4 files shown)


sources/gdal/large_image_source_gdal/__init__.py
```diff
@@ -91,7 +91,7 @@ class GDALFileTileSource(FileTileSource, metaclass=LruCacheMetaclass):
         'image/tiff': SourcePriority.LOW,
         'image/x-tiff': SourcePriority.LOW,
     }
-    geospatial = True
```
banesullivan, line 94: I didn't see anywhere obvious where this was actually used

manthey:
@banesullivan I pushed a change (and rebased to latest master) to make this work.  Change to a non-draft PR and I can approve it.

banesullivan:
In downstream projects, I have implemented the following to work around this - so this isn't crucial but would be nice to have

https://github.com/girder/django-large-image/blob/43c9ef62d83c4d6f33d3b49cc41f2c678f296584/django_large_image/tilesource.py#L54-L60

sources/gdal/large_image_source_gdal/__init__.py
```diff
@@ -601,14 +601,21 @@ def getBandInformation(self, statistics=True, dataset=None, **kwargs):
             self._bandInfo = infoSet
         return self._bandInfo
 
+    @property
+    def geospatial(self):
```
banesullivan, line 605: It may be good to put this property on the base tile source class and return False by default

manthey: It can just be a constant on the base class (see https://github.com/girder/large_image/blob/patch/geospatial-attr/large_image/tilesource/base.py#L47).

manthey (approved):

sources/gdal/large_image_source_gdal/__init__.py
banesullivan: Ah right 🤦🏻‍♂️

PR #819, opened 2022-04-08
INFO to DEBUG log level

### LINKED ISSUE #690: Change log level to debug where it makes sense
There are a ton of `INFO` level logs. Can many of these be changed to `DEBUG` level where it makes sense?

Extension of #689

banesullivan:
Resolves #690

This changes most INFO level logs to DEBUG - some were changed to ERROR where I thought appropriate

### DIFF
--- a/girder/girder_large_image/rest/large_image_resource.py
+++ b/girder/girder_large_image/rest/large_image_resource.py
@@ -59,7 +59,7 @@ def createThumbnailsJobTask(item, spec):
         except TileGeneralError as exc:
             status['failed'] += 1
             status['lastFailed'] = str(item['_id'])
-            logger.info('Failed to get thumbnail for item %s: %r' % (item['_id'], exc))
+            logger.error('Failed to get thumbnail for item %s: %r' % (item['_id'], exc))
         except AttributeError:
             raise
         except Exception:
@@ -190,7 +190,7 @@ def createThumbnailsJob(job):
                         JobStatus.ERROR: 'stopped due to error',
                     }[None if not job else job.get('status')]
                     msg = 'Large image thumbnails job %s' % cause
-                    logger.info(msg)
+                    logger.debug(msg)
                     # Cancel any outstanding tasks.  If they haven't started,
                     # they are discarded.  Those that have started will still
                     # run, though.
@@ -208,7 +208,7 @@ def createThumbnailsJob(job):
         # Clean up the task pool asynchronously
         pool.shutdown(False)
     job, msg = createThumbnailsJobLog(job, status, 'Finished: ', JobStatus.SUCCESS)
-    logger.info(msg)
+    logger.debug(msg)
 
 
 class LargeImageResource(Resource):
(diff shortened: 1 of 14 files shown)


manthey:
I don't understand the point of changing a lot of these -- many of them are intended to be visible information, not just extra debug information.  I think you really just don't want to see logs from the library, not change their level.

Maybe we should change the default logger that is set up instead.  Python recommends libraries have:
logging.getLogger(__name__).addHandler(logging.NullHandler())
as the default handler for their logger.  So changing large_image/config.py to have the fallback logger use the `NullHandler` instead of the `StreamHandler` is probably a better choice than changing log levels.

As an example where I'd rather not change from INFO to DEBUG, tiling frames can be slow, so logging information every 10 seconds is useful to see that something is progressing.  Similarly, the logs in the thumbnail task in Girder effectively vanish if we set them to debug, losing feedback from the job.

banesullivan:
> have the fallback logger use the NullHandler instead of the StreamHandler is probably a better choice than changing log levels.

I didn't know about that. I'll see if I can give this PR another go to implement that and make sure the changes are more useful/relevant. We'd need to change all `logger.` to `config.getConfig('logger')`, right?

> I think you really just don't want to see logs from the library, not change their level.

Somewhat - there are a few that should be changed to an ERROR level rather than INFO but those are minor. It also makes it easier to filter on papertrail to know what truly is DEBUG info vs INFO info vs logged ERRORs

manthey:
> I didn't know about that. I'll see if I can give this PR another go to implement that and make sure the changes are more useful/relevant. We'd need to change all `logger.` to `config.getConfig('logger')`, right?

I just checked, and outside of the Girder packages, we already are exclusively using the logger from the config.  In some girder code we are using the girder logger.  The converter utility also uses its own logger rather than the general large-image logger.  So, no changes needed for that purpose.

girder/girder_large_image/rest/large_image_resource.py
```diff
@@ -59,7 +59,7 @@ def createThumbnailsJobTask(item, spec):
         except TileGeneralError as exc:
             status['failed'] += 1
             status['lastFailed'] = str(item['_id'])
-            logger.info('Failed to get thumbnail for item %s: %r' % (item['_id'], exc))
+            logger.error('Failed to get thumbnail for item %s: %r' % (item['_id'], exc))
```
manthey, line 62: I think when `imageKey` is present, this should actually do nothing in the exception rather than increment failed and log anything.

manthey:
Regarding changing some logging to ERROR level -- let's keep that, though I've commented on one log message that sometimes shouldn't occur; if we don't address that on this PR, we should make an issue so it can be addressed separately.

manthey:
I think with #840, we should either close this PR without merging, or just reduce it to changing the non-girder log messages that are being promoted to error level (which I think is just one in the gdal source).

banesullivan:
I'm fine just closing this.

My qualm was that ANY logs from large-image were being captured by sentry.io for django-large-image. So I just did this to mitigate it: https://github.com/girder/django-large-image/commit/13273f5eabf422ce657df58a0997045ceb62d3e9

PR #831, opened 2022-04-19
Add scm version fallback.

manthey:
If setuptools_scm is installed and a local deploy is made that isn't a github repo, then packages fail to install unless there is a fallback version.


banesullivan (approved):
Let's merge this

PR #834, opened 2022-04-26
Add Python-slim Docker image

banesullivan:
Builds a Python3.9-slim Docker image for `master`, each tag, and each PR and publishes the image to the GitHub container registry.

This image does NOT install the girder dependencies.

This image is ~2.2 GB when built locally (so not all that slim). Considering this size, I have added a separate slim-geospatial image that only has the sources/utilities relevant for geospatial formats which is ~1.15 GB under the `-geo` tag suffix


.dockerignore
```diff
@@ -1,9 +1,2 @@
 build
 test
-.git
-girder
-girder_annotation
-large_image
-sources
-utilities
-examples
```
banesullivan, line 9: This might be causing problems for CircleCI

slim.Dockerfile
```diff
@@ -0,0 +1,31 @@
...
+LABEL repo="https://github.com/girder/large_image"
+# NOTE: this does not install any girder3 packages
+RUN pip install \
+    --find-links https://girder.github.io/large_image_wheels \
+    --find-links=/opt/wheels \
+    'pylibmc>=1.5.1' \
+    matplotlib \
+    pyvips \
+    simplejpeg \
+    $(ls -1  /opt/wheels/large_image*.whl)
```
banesullivan, line 31: This still installs `large_image_tasks` which pulls some of the girder stack

.dockerignore
manthey: The existing dockerfile is used for one of the test environments and shouldn't care.  It probably should be renamed (to `test.Dockerfile`?) to disambiguate the purposes for the new Dockerfile.  You should only need to change the build line in .circleci/config.yml to handle such a name change.

slim.Dockerfile
```diff
@@ -0,0 +1,50 @@
+# Build wheels
+FROM python:3.9-slim as build
```
manthey, line 2: Should we use Python 3.10 in preference to 3.9?  Or, what should be our criterion for choosing a base python?

slim.Dockerfile
```diff
@@ -0,0 +1,50 @@
...
+
+# Need git for setuptools_scm
+RUN apt-get update \
```
manthey, line 5: If we are using setuptools_scm, we probably don't want `.git` in the `.dockerignore` file, as we won't pick up the version correctly.

banesullivan: I took `.git` out of the `.dockignore` for this

slim.Dockerfile
banesullivan: With that said, we can easily add a build argument and publish 3.8, 3.9, and 3.10 images

.dockerignore
manthey: And, if you think is semantically clearer, feel free to move the existing Dockerfile to some subdirectory within the project.

banesullivan: I'll rename that to `test.Dockerfile` to be more clear.

slim.Dockerfile
banesullivan: Further thoughts - we should use the same Python version that the Girder4 cookie-cutter uses: https://github.com/girder/cookiecutter-girder-4/blob/d1912b887133ae2407277f772f6329c082fafb73/%7B%7B%20cookiecutter.project_slug%20%7D%7D/dev/django.Dockerfile#L1

banesullivan:
From my perspective, this is ready to merge and the versioning issue I was noticing earlier is now resolved.

Any changes to the Python version or adding addition tags should come in follow up PRs as needed

manthey (approved):

PR #835, opened 2022-04-26
Add validate_cog method to GDALFileTileSource

banesullivan:
This adds a `validate_cog` method to `GDALFileTileSource` to use the `osgeo_utils` script that will verify a COG.

I'm wondering if we should add another `validate_*` method to the `TiledTiffDirectory` that uses `tifftools` to check the IFDs and tiling of an image

### DIFF
--- a/sources/gdal/large_image_source_gdal/__init__.py
+++ b/sources/gdal/large_image_source_gdal/__init__.py
@@ -41,7 +41,9 @@ from large_image.cache_util import CacheProperties, LruCacheMetaclass, methodcac
 from large_image.constants import (TILE_FORMAT_IMAGE, TILE_FORMAT_NUMPY,
                                    TILE_FORMAT_PIL, SourcePriority,
                                    TileInputUnits, TileOutputMimeTypes)
-from large_image.exceptions import TileSourceError, TileSourceFileNotFoundError
+from large_image.exceptions import (TileSourceError,
+                                    TileSourceFileNotFoundError,
+                                    TileSourcePyramidFormatError)
 from large_image.tilesource import FileTileSource
 from large_image.tilesource.utilities import getPaletteColors
 
@@ -56,6 +58,10 @@ try:
 except PackageNotFoundError:
     # package is not installed
     pass
+try:
+    from osgeo_utils.samples.validate_cloud_optimized_geotiff import validate as gdal_validate
+except ImportError:
+    gdal_validate = None
 
 TileInputUnits['projection'] = 'projection'
 TileInputUnits['proj'] = 'projection'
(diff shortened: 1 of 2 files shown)


manthey:
We should figure out how to harmonize this with the "inefficient warning" here https://github.com/girder/large_image/blob/master/sources/tiff/large_image_source_tiff/__init__.py#L324-L339

manthey:
If a file is small enough, even if it isn't a COG, we probably shouldn't consider it a problem (e.g., if a file is only a few kilobytes, the fact it isn't tiled or optimized is probably irrelevant).  Since there are formats that are efficient that aren't tiff, so we probably can only intelligently detect that a file is inefficient in a small set of cases.

banesullivan:
> If a file is small enough, even if it isn't a COG, we probably shouldn't consider it a problem ...

I'm not sure if I'd like this new `validate_cog` method to handle that, I think this method should solely inform users on whether it is a valid COG or not regardless of size.

IMO, any downstream library that is calling this function should handle the cursory check of: *is this under my file size threshold?* no? then check if valid COG.

> We should figure out how to harmonize this with the "inefficient warning"

Hm. should I add `_checkForInefficientDirectories()` as an abstract method on `FileTileSource` and let the GDAL source override it as:

```py
    def _checkForInefficientDirectories(self, warn=True):
        # TODO: perhaps do a cursory check of the file size to see if it's under some minimum
        try:
            return self.validate_cog()
        except TileSourcePyramidFormatError as e:
            pass
        if warn:
            config.getConfig('logger').warning(f'Raster is not Cloud Optimized: {e}')
        self._inefficientWarning = True
```

sources/gdal/large_image_source_gdal/__init__.py
```diff
@@ -1185,6 +1191,44 @@ def getRegion(self, format=(TILE_FORMAT_IMAGE, ), **kwargs):
...
+
+        Parameters
+        ----------
+        check_tiled : bool
+            Set to False to ignore missing tiling.
+        full_check : bool
+            Set to True to check tile/strip leader/trailer bytes.
+            Might be slow on remote files
+        strict : bool
+            Enforce warnings as exceptions. Set to False to simple warn and not
```
manthey, line 1210: s/simple/simply/

manthey:
> Hm. should I add `_checkForInefficientDirectories()` as an abstract method on `FileTileSource` and let the GDAL source override it as:
> 
> ```python
>     def _checkForInefficientDirectories(self, warn=True):
>         try:
>             return self.validate_cog(warn=warn)
>         except TileSourcePyramidFormatError:
>             pass
>         self._inefficientWarning = True
> ```

Let's defer this to a later conversation or PR.  There is another project where we may try to determine this for more different sources and we can take it back up then.

manthey:
I pushed a change to the test that checks which sources can read which files so that the new test files don't cause tests to fail.

sources/gdal/large_image_source_gdal/__init__.py
```diff
@@ -56,6 +58,10 @@
 except PackageNotFoundError:
     # package is not installed
     pass
+try:
```
manthey, line 61: Since this is only used in the validateCOG method, would it be more appropriate to just move the import there (and then we don't have to have `gdal_validate = None`, since it would just raise the error directly?

banesullivan: Agreed - will change

manthey (approved):

PR #842, opened 2022-04-28
Improve example usage

manthey:


manthey:
Thank you @Leengit for the feedback.

docs/example_usage.rst
```diff
@@ -75,18 +76,22 @@ You can specify the size in physical coordinates:
...
     tile0 = source.getTile(0, 0, 0)
     # tile0 is the lowest resolution tile that shows the whole image.  It will
     # be a JPEG or PNG or some other image format depending on the source
     tile002 = source.getTile(0, 0, 2)
-    # tile002 will be a tile a tile representing no more than 1/4 the width of
-    # image in the upper-left corner.
+    # tile002 will be a tile representing no more than 1/4 the width of the
+    # image in the upper-left corner.  Since the z (third parameter) is 2, the
+    # level will have up to 2**2 x 2**2 (4 x 4) tiles.  Since the image may not
+    # be a power of two and square, there may be fewer tiles than this.
```
Leengit, line 94: I don't know the answer to this question but I ask it just in case: is it "power of two" or "multiple of four" that is applicable in this case?

Also, maybe change "and square" to something more like "in size and square".

Leengit: Another question I don't know the answer to.  I the very last sentence here, is it fewer tiles or fewer pixels?

manthey: Maybe this comment should read something like
```
    # tile002 will be a tile representing no more than 1/4 the width of the
    # image in the upper-left corner.  Since the z (third parameter) is 2, the
    # level will have up to 2**2 x 2**2 (4 x 4) tiles.  An image doesn't
    # necessarily have all tiles in that range, as the image may not be square.
```

and then below the code blocks we could elaborate:

Tile ``x``, ``y``, and ``z`` values are based on the tile width and height and on the image width and height.  For example if an image is 58368 x 12288 pixels in size and the tiles are 256 x 256, there are nine levels of tiles (numbered 0 through 8).  The highest level (8) will be composed of 228 x 48 tiles, so if ``z = 8``, then ``x`` can range from 0 to 227 inclusive and ``y`` can range from 0 to 47 inclusive.  Each lower level has half as many tiles in each direction, rounded up.  For instance, if ``z = 5``, ``x`` can range from 0 to 28 and ``y`` can range from 0 to 5.  If ``z = 2``, ``x`` can range from 0 to 3 and ``y`` can only be 0.


I'm trying to balance conciseness with clarity.  Maybe it would be better to also have a table showing off the example:


z | horizontal tiles | vertical tiles | valid x values | valid y values
--- | --- | --- | --- | ---
8 | 228 | 48 | 0 - 227 | 0 - 47
7 | 114 | 24 | 0 - 113 | 0 - 23
6 | 57 | 12 | 0 - 56 | 0 - 11
5 | 29 | 6 | 0 - 28 | 0 -5
4 | 15 | 3 | 0 - 14 | 0 - 2
3 | 8 | 2 | 0 - 7 | 0 - 1
2 | 4 | 1 | 0 - 3 | 0
1 | 2 | 1 | 0 - 1 | 0
0 | 1 | 1 | 0 | 0

Leengit: Something like?: "The returned tile will be of size `(tileWidth, tileHeight)` regardless of the selected level.
When `level==max_level` the upper-left-hand corner  of the requested tile will be `(x * tileWidth, y * tileHeight)` as measured at maximum resolution.  When `level==max_level - n` for a positive integer `n`, the tile data will come from a tile of size `(tileWidth * 2**n, tileHeight * 2**n)` with upper-left-hand corner at `(x * tileWidth * 2**n, y * tileHeight * 2**n)` as measured at maximum resolution."

Leengit: "In particular, `x * tileWidth * 2**n` must be strictly less than `sizeX` and `y * tileHeight * 2**n` must be strictly less than `sizeY`."

manthey:
I'm going to merge this to get the typo fixes in -- I'll make another PR if we want to adjust it further.

docs/example_usage.rst
```diff
@@ -97,6 +102,8 @@ Some methods such as ``getRegion`` and ``getThumbnail`` allow you to specify for
     tile0 = source.getTile(0, 0, 0)
     # tile is now guaranteed to be a PNG
 
+Tiles are always ``tileWidth`` by ``tileHeight`` in pixels.  At the maximum level (``z = levels - 1``), the number of tiles in that level will range in ``x`` from ``0`` to strictly less than ``sizeX / tileWidth``, and ``y`` from ``0`` to strictly less than ``sizeY / tileHeight``.  For each lower level, the is a power of two less tiles.  For instance, when ``z = levels - 2``, ``x`` ranges from ``0`` to less than ``sizeX / tileWidth / 2``; at ``z = levels - 3``, ``x`` is less than ``sizeX / tileWidth / 4``.
```
Leengit, line 105: `levels-2` should be a factor of 4; `levels-3` should be a factor of 8.

manthey: No.  `levels` returned from the `getMetadata` is the number of levels.  The z value can range from 0 to `levels - 1` (e.g., an image with 9 levels has z-values [0, 8].  Hence `levels - 1` is the full res version.

PR #844, opened 2022-04-29
Allow polygons with holes to be specified.

manthey:
This will *render* polygons with holes, but doesn't yet support editing them.

Some scalability tests are needed.


manthey:
I'll note that in HistomicsUI where this can be displayed, we already correctly handle computing area and perimeter for polygons with holes.

PR #846, opened 2022-05-02
Speed up validation of polygons with a mix of holes.

manthey:


manthey:
Schema validation could be parallelized to speed it up further.

PR #849, opened 2022-05-03
Reduce the strictness of version checks in dev branches

manthey:
Otherwise, you pretty much have to do `tox -r` after every change if you want to test the results.

### DIFF
--- a/girder/setup.py
+++ b/girder/setup.py
@@ -27,7 +27,7 @@ try:
     from setuptools_scm import get_version
 
     version = get_version(root='..', local_scheme=prerelease_local_scheme)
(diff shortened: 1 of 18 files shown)


manthey:
I was fooling with a test and found that every time I ran tox it would complain with an error like
```
raise VersionConflict(dist, req).with_context(dependent_req)
pkg_resources.VersionConflict: (large-image 1.14.3.dev11+g9d83e92.d20220503 (.../large_image/build/tox/test-py39/lib/python3.9/site-packages), Requirement.parse('large_image>=1.14.3.dev11+g9d83e92')
```

PR #854, opened 2022-05-04
Add Jupyter docker images

banesullivan:
Publishes Jupyter docker images as well


manthey:
Should we add the published tags of the docker image to the README?

manthey:
And, with this, it would be great to add a Jupyter notebook example with geo and non-geo sources in that would be in the docs (with instructions to run it on a local Jupyter docker instance and a link to open it in colab or something similar).

PR #862, opened 2022-05-20
Add scaleWithZoom property to heatmaps

naglepuff:
This PR updates the annotation schema for heatmap elements to include a `scaleWithZoom` property. It also updates documentation to reflect this change and sets the property in the style dict when creating new heatmaps.

### DIFF
--- a/girder_annotation/docs/annotations.rst
+++ b/girder_annotation/docs/annotations.rst
@@ -185,6 +185,9 @@ near by values aggregate together when viewed.
     "normalizeRange": true             # If true, the rangeValues are normalized to [0, 1].  If
                                        # false, the rangeValues are in the
                                        # value domain.  Defaults to true.  Optional
+    "scaleWithZoom": true              # If true, scale the size of points with the zoom level of
+                                       # the map. In this case, the radius property refers to the
+                                       # size of points at zoom level 0. Defaults to false. Optional
   }
 
 Grid Data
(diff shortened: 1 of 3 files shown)


girder_annotation/docs/annotations.rst
```diff
@@ -185,6 +185,9 @@ near by values aggregate together when viewed.
     "normalizeRange": true             # If true, the rangeValues are normalized to [0, 1].  If
                                        # false, the rangeValues are in the
                                        # value domain.  Defaults to true.  Optional
+    "scaleWithZoom": true              # If true, scale the size of points with the zoom level of
```
manthey, line 188: Can you add a comma after `"normalizeRange": true` for consistency with the rest of the examples?

manthey:
In considering this, I think we should diverge from the GeoJS method where scaleWithZoom is based on level 0.  For geospatial sources, this makes some sense because there is no native or maximum zoom, and you need some reference.  For images, we could make the radius the size at the native pixel level.  Internally, this means that we get to divide by 2**(number of levels -1), but I think handling that programmatically will have less cognitive dissonance than having the user know how big tiles are and how many levels there are.

For this PR, this means changing the wording.

naglepuff:
> For this PR, this means changing the wording.

The description has been updated in 12ffeb0 such that zoom level 0 is no longer mentioned. Feel free to suggest a different wording than I have if it is not up to snuff.

PR #871, opened 2022-05-31
Reduce the chance of blinking on annotation layers.

manthey:
When the base image is panned, sometimes an update occurs before the reset features are updated.  This forces the update sooner.


ds2268:
@manthey I checked, as described in https://github.com/DigitalSlideArchive/digital_slide_archive/issues/214 and it does not solve the issue.

I also checked dsa_girder docker container and the large_image contains the latest changes, so I should have the patches included. I suggest that we have a call where I demo to you the bottlenecks and blinking/refreshing stuff.

ds2268:
Ah, I have now cleaned all the old docker images manually and redeployed the DSA and now the bug seems to be fixed. It's really much more responsive! Next week we will start real-world usage again on a larger scale and will report back if there are any bottlenecks left. Thanks!

PR #876, opened 2022-06-16
Abstract caching and support entrypoints

banesullivan:
Abstracts the caching mechanism so that external projects can implement a cache interface and use entrypoints to make it available to large_image.

Generally, this is working for django-large-image with https://github.com/girder/django-large-image/pull/32

Todo:

- [x] Can the `MemCache` cache be chosen appropriately? Is it ignored when not available and is it chosen over `python` when available?
- [x] Should `loadCaches()` only be called once rather than on every `CacheFactory.getCache()` call?
- [x] Address inline TODO statements
- [x] What is `CacheFactory.getCacheSize()`and `pickAvailableCache()` and how are they affected by these changes?
- [ ] What happens if `BaseCache.getCache()` returns a `None` cache lock? can/should we support that?
- [x] Test downstream in `django-large-image`

### DIFF
--- a/large_image/cache_util/cachefactory.py
+++ b/large_image/cache_util/cachefactory.py
@@ -16,7 +16,6 @@
 
 import math
 import threading
-from collections import OrderedDict
 
 import cachetools
 
@@ -37,10 +36,12 @@ try:
 except ImportError:
     MemCache = None
 
-availableCaches = OrderedDict()
+# DO NOT MANUALLY ADD ANYTHING TO `_availableCaches`
+#  use entrypoints and let loadCaches fill in `_availableCaches`
+_availableCaches = {}
 
 
-def loadCaches(entryPointName='large_image.cache', sourceDict=availableCaches):
+def loadCaches(entryPointName='large_image.cache', sourceDict=_availableCaches):
     """
     Load all caches from entrypoints and add them to the
     availableCaches dictionary.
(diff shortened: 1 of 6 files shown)


large_image/cache_util/cachefactory.py
```diff
@@ -89,32 +121,23 @@ def getCacheSize(self, numItems, cacheName=None):
...
+            config.setConfig('cache_backend', cacheBackend)
+
         if cacheBackend:
             cacheBackend = str(cacheBackend).lower()
+
         cache = None
-        if cacheBackend == 'memcached' and MemCache and numItems is None:
-            # lock needed because pylibmc(memcached client) is not threadsafe
-            cacheLock = threading.Lock()
+        # TODO: why have this numItems check?
```
manthey, line 137: numItems is None for the tile caches and not None for the source cache.  The source cache has to use in-process caching (hence, python), since it includes open file handles.  There should probably be a clearer mechanism for this (e.g., some flag like "in-process") that would be used instead of this cryptic code.

banesullivan:
@manthey, this seems to be working and is tested downstream.I'm ready for a final review if you have time

manthey:
Excellent.

manthey (approved):

manthey:
> What happens if BaseCache.getCache() returns a None cache lock? can/should we support that?
That should work correctly.

manthey:
> Can we also tag 1.14.6 (or should this be 1.15 since there is a new interface) after this merges?

It will be 1.15.0 since I think this is enough of a new interface.

PR #883, opened 2022-07-05
Support more style range options

manthey:


large_image/tilesource/base.py
```diff
@@ -1194,6 +1218,13 @@ def _applyStyle(self, image, style, x, y, z, frame=None):  # noqa
                 else:
                     output[:, :, channel] = numpy.maximum(
                         output[:, :, channel], numpy.where(keep, clrs, 0))
+        if dtype == 'uint16':
+            output = (output * 65535 / 255).astype(numpy.uint16)
+        elif dtype == 'float':
+            output /= 255
+        print(axis, output.shape)
```
banesullivan, line 1225: remove print statement?

manthey: Thanks.

banesullivan:
I'm curious when `full` is wanted/needed?

banesullivan:
I suppose I've had a few `float` rasters but the data range was always much larger than 255 and thus the tiles would come back looking a single color. Setting `full` to 1 might still have the same issue?

banesullivan:
All makes sense! Thanks for the explanation!

PR #894, opened 2022-07-25
Copy tilesource instances to add styles

manthey:
Instead of reopening a tilesource when adding a style, copy the unstyled class and add the style information.

I've marked this as draft as I think it requires more tests to verify it does no harm.

### DIFF
--- a/large_image/cache_util/cache.py
+++ b/large_image/cache_util/cache.py
@@ -1,3 +1,4 @@
+import copy
 import functools
 import threading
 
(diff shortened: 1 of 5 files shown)


manthey:
@banesullivan This should speed up generating regions and tiles with different styles from the same source file by avoiding reopening the file handles.  I *think* it should work without issue, but, since file handles and locks are shared across classes, I think there is a potential that some additional lock may be needed in some case.

Please try it out and let me know if you see any issues when using multiple styles and sources.  I'll convert it to a non-draft when I've added more tests.

banesullivan:
Thanks for pinging me on this. I'd like to stress test this with Django-large-image when I get back from PTO (tomorrow)

manthey:
I've added some tests to ensure sources don't close handles prematurely.  @banesullivan If your testing is successful, I think we can merge this.

banesullivan:
LGTM. Tests in `django-large-image` are happy, and my poking and prodding went well

banesullivan (approved):

PR #909, opened 2022-08-04
Fix alpha on some GDAL sources

manthey:
This affected sources with a projection that have an explicit alpha channel.

Fixes https://github.com/girder/django-large-image/issues/39

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -18,11 +18,14 @@
 - Speed up styling by doing less ([892](../../pull/892))
 - Add local color definitions ([858](../../pull/858))
 - Inheritable config files ([897](../../pull/897))
-- Add geospatial property ([818](../../pull/818))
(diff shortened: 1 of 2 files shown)


banesullivan (approved):

banesullivan:
Turns out this didn't completely fix the issue:

In the following example, the alpha is still improperly handled

```py
import large_image

path = '/Users/bane/Desktop/rasters/Django-Large-Image/4-ortho-rgb-cog.tif'

src = large_image.open(path, projection='EPSG:3857', style={
 'bands': [
    {'band': 1, 'palette': 'R'},
    {'band': 2, 'palette': 'G'},
    {'band': 3, 'palette': 'B'}
  ]
})
thumb = src.getThumbnail(encoding='PNG')[0]
thumb
```
download

PR #916, opened 2022-08-08
cleaning and error in the gdal file

12rambau:
As I'm working on the creation of a rasterio provider I need to fully understand the gdal file. I have a lot of questions so I oppened this draft to notify you of my findings. 

Please comment if something I post here is not making sense (I'll use the comment functionality of the PR to link code).

When I'll be done I'll remove the draft mode and I assume you'll squash my commits. What do you think ?

### DIFF
--- a/sources/gdal/large_image_source_gdal/__init__.py
+++ b/sources/gdal/large_image_source_gdal/__init__.py
@@ -872,14 +872,8 @@ class GDALFileTileSource(FileTileSource, metaclass=LruCacheMetaclass):
                 'left and right and at least one of top and bottom is '
                 'specified.')
         if not self.projection:
-            pleft, ptop = self.toNativePixelCoordinates(
-                right if left is None else left,
-                bottom if top is None else top,
-                units)
-            pright, pbottom = self.toNativePixelCoordinates(
-                left if right is None else right,
-                top if bottom is None else bottom,
-                units)
+            pleft, ptop = self.toNativePixelCoordinates(left or right, top or bottom, units)
+            pright, pbottom = self.toNativePixelCoordinates(right or left, bottom or top, units)
             units = 'base_pixels'
         else:
             inProj = self._proj4Proj(units)
--- a/sources/gdal/large_image_source_gdal/__init__.py
+++ b/sources/gdal/large_image_source_gdal/__init__.py
@@ -942,9 +942,9 @@ class GDALFileTileSource(FileTileSource, metaclass=LruCacheMetaclass):
             if right is None:
                 right = bounds['xmax'] if width is None else left + width
             if top is None:
-                top = bounds['ymax'] if bottom is None or width is None else bottom - width
+                top = bounds['ymax'] if bottom is None or height is None else bottom + height
             if bottom is None:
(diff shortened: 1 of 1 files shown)


sources/gdal/large_image_source_gdal/__init__.py
```diff
@@ -942,9 +942,9 @@ def _getRegionBounds(self, metadata, left=None, top=None, right=None,
             if right is None:
                 right = bounds['xmax'] if width is None else left + width
             if top is None:
-                top = bounds['ymax'] if bottom is None or width is None else bottom - width
+                top = bounds['ymax'] if bottom is None or height is None else bottom + height
             if bottom is None:
-                bottom = bounds['ymin'] if width is None else top + width
+                bottom = bounds['ymin'] if height is None else top - height
```
12rambau, line 947: - and + where inverted as well. We agree that height is positive ?

sources/gdal/large_image_source_gdal/__init__.py
```diff
@@ -872,14 +872,8 @@ def _convertProjectionUnits(self, left, top, right, bottom, width, height,
...
         if not self.projection:
-            pleft, ptop = self.toNativePixelCoordinates(
-                right if left is None else left,
-                bottom if top is None else top,
-                units)
-            pright, pbottom = self.toNativePixelCoordinates(
-                left if right is None else right,
-                top if bottom is None else bottom,
-                units)
+            pleft, ptop = self.toNativePixelCoordinates(left or right, top or bottom, units)
```
12rambau, line 875: reduce complexity with nice use of the or parameter

manthey: None != 0, so this doesn't actually do the same thing.  Specifically, this is if right is not specified, use left as a fall-back.  You might want a 0 coordinate depending on the crs (e.g., it could be the prime meridian of your reference system).

sources/gdal/large_image_source_gdal/__init__.py
manthey: `width` should be `height` in the `bottom =` line, but bottom > top depends on the coordinate system.  I'll have to investigate further.

manthey:
@12rambau Thanks for working on this.

sources/gdal/large_image_source_gdal/__init__.py
12rambau: `left is None or right` will do the trick then right ?

12rambau: sorry didn't thought about actual zero value (Flake8 would have complained anyway)

12rambau:
https://github.com/12rambau/large_image/blob/b1f41afb065702edad3a97cba5008be135b5e532/sources/gdal/large_image_source_gdal/__init__.py#L183

what's the use of this lock ? It's never called in this file

12rambau:
https://github.com/girder/large_image/blob/2ca878e1318eaa43c754a27587530746625a9b83/sources/gdal/large_image_source_gdal/__init__.py#L395 

Shouldn't it be `getURLHash` ?

manthey:
This has been superceded by #1115.

PR #927, opened 2022-08-11
Rasterio source

### LINKED ISSUE #898: create a rasterio driver for large-image?
don't get me wrong the question is not "is geospatial version of large_image necessary" but more about the use of GDAL itself. 

GDAL is a great library but it has the really bad habit of being the worst possible neighbor in a python environment. I started to use https://localtileserver.banesullivan.com/installation/index.html for my project and since I publish the latest version all the pip-related environments failed because of gdal inconsistency and version crash. 

I search in the GDAL repository itself and there is a very long thread of about a potential wheel distribution for GDAL https://github.com/OSGeo/gdal/issues/3060 that seems on stale. 

As large_image seems to be developed for python environments being pip-friendly seems like a desirable objective. Do you already considered using a pure rasterio driver ?

12rambau:
Fix #898 

at the moment this is just to show that I'm moving forward. 

I think I still missed some very specific tricks with projections but we are getting there. If you want to have a look be my guest, I still need to test it but that's complicated in my environment. 

> **Note**
> don't trigger the build yet, it's not ready at all


test/test_source_rasterio.py
```diff
@@ -0,0 +1,604 @@
...
+    assert image[:len(utilities.PNGHeader)] == utilities.PNGHeader
+    
+    # We get a different thumbnail with a projection
+    source = large_image_source_rasterio.open(imagePath, projection='EPSG:3857')
+    image2, mimeType = source.getThumbnail(encoding='PNG')
+    assert image2[:len(utilities.PNGHeader)] == utilities.PNGHeader
+    assert image != image2
+
+
+def testPixel():
```
12rambau, line 166: @manthey, I think I'll need some support satrting from here, I cannot manage to make the test to work and it's coming from outside of the actual source file so I think I'll loose my time trying to understand what's wrong. 

When I run the test for test/test_source_rasterio.py::testPixel, I always get an empty rgb values. could you have a look ?

manthey: I'll take a look.

manthey: If you open a sample file (e.g., test/test_files/rgb_geotiff.tiff) without a projection and then call getThumbnail, you'll see that returns an all black result.  getPixel is just the victim of this...  I haven't looked into what is happening without a projection yet.

12rambau: hmmm,I need to check then, because when I was looking at the generated jpeg it seemed to me that it was working

12rambau:
thanks @banesullivan for the updates, we started a house project with my partner, I have very little time to dedicate to FOSS these day. I will be back when the kitchen is fully operational

banesullivan:
No worries, @12rambau! enjoy some time away from the keyboard! Thank you for getting so far along in this PR -- @manthey and I should be able to wrap it up and merge it soon here (perhaps not in a perfect state, but far enough that this source can be used downstream and make the installation process much easier)

PR #937, opened 2022-08-17
Add annotation creation access flag

eagw:
Adds a new girder access flag which demarcates whether or not a user has permission to create annotations, allowing users to create annotations without having write access to a folder/item.  Also creates a new endpoint indicating whether or not the user has creation permissions (via write access or the flag) and makes current annotation creation procedures check for the access flag.

### DIFF
--- a/girder_annotation/girder_large_image_annotation/__init__.py
+++ b/girder_annotation/girder_large_image_annotation/__init__.py
@@ -15,6 +15,7 @@
 #############################################################################
 
 from girder import events
+from girder.constants import registerAccessFlag
 from girder.exceptions import ValidationException
 from girder.plugin import GirderPlugin, getPlugin
 from girder.settings import SettingDefault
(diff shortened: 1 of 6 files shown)


girder_annotation/girder_large_image_annotation/__init__.py
```diff
@@ -58,6 +59,10 @@ def validateBoolean(doc):
     constants.PluginSettings.LARGE_IMAGE_ANNOTATION_HISTORY: True,
 })
 
+# Access flags
+
+registerAccessFlag('createAnnots', 'Create annotations', 'Allow user to create annotations')
```
manthey, line 64: Can we define this access flag (the `createAnnots`) in `girder/girder_large_image/constants.py` (maybe at the root level of that modules as `ANNOTATION_ACCESS_FLAG = 'large_image_create_annotations'`), and then use the constant instead of the explicit string?

manthey:
Other than preferring a defined constant to a string, this looks good.

girder_annotation/girder_large_image_annotation/__init__.py
eagw: See 99ada00

manthey (approved):

PR #959, opened 2022-09-07
Improve scaled overlays.

manthey:
Overlay annotations at scales other than the base image could show eccentric tile layers.


manthey:
@naglepuff This should look much better with superpixels generated at scales other than the base image scale.

PR #960, opened 2022-09-12
Use functions in styles.

### LINKED ISSUE #688: Band math when for getTile and getRegion
Would it be possible to provide some sort of callable that will perform a mathematical operation on a single band when tile serving?

Let's say I have some massive Digital Elevation Model (DEM) and I want to dynamically produce Hillshade maps from it. Ideally, I'd like to have the tile server do this automatically when serving the tiles.

I imagine this being some sort of option for both `getRegion` and `getTile` so that I can serve tiles with the math applied or get a small ROI with the math applied.

Example math:

<details>

```py
import numpy as np

# https://github.com/earthlab/earthpy/blob/9ad455e85002a2b026c78685329f8c5b360fde5a/earthpy/spatial.py#L564
def hillshade(arr, azimuth=30, altitude=30):
    """Create hillshade from a numpy array containing elevation data.
    Parameters
    ----------
    arr : numpy array of shape (rows, columns)
        Numpy array with elevation values to be used to created hillshade.
    azimuth : float (default=30)
        The desired azimuth for the hillshade.
    altitude : float (default=30)
        The desired sun angle altitude for the hillshade.
    Returns
    -------
    numpy array
        A numpy array containing hillshade values.
    """
    try:
        x, y = np.gradient(arr)
    except ValueError:
        raise ValueError("Input array should be two-dimensional")

    if azimuth <= 360.0:
        azimuth = 360.0 - azimuth
        azimuthrad = azimuth * np.pi / 180.0
    else:
        raise ValueError(
            "Azimuth value should be less than or equal to 360 degrees"
        )

    if altitude <= 90.0:
        altituderad = altitude * np.pi / 180.0
    else:
        raise ValueError(
            "Altitude value should be less than or equal to 90 degrees"
        )

    slope = np.pi / 2.0 - np.arctan(np.sqrt(x * x + y * y))
    aspect = np.arctan2(-x, y)

    shaded = np.sin(altituderad) * np.sin(slope) + np.cos(
        altituderad
    ) * np.cos(slope) * np.cos((azimuthrad - np.pi / 2.0) - aspect)

    return 255 * (shaded + 1) / 2

```

</details>

NOTE: This is for a personal project

manthey:
Resolves #688.

This can be used, for instance, to get a mask from a superpixel image. For instance, suppose you want a mask for the superpixel with an index of 82, you'd use a style like:
`{"function":{"name":"large_image.tilesource.stylefuncs.maskPixelValues","context":true,"parameters":{"values":[82]}},"bands":[]}` (or, if you are using superpixels with boundaries, this would be `{"function":{"name":"large_image.tilesource.stylefuncs.maskPixelValues","context":true,"parameters":{"values":[164,165]}},"bands":[]}`). The resulting image (assuming you are returning a PNG), will be opaque black where this value is not present and transparent white where it is (or specify the `negative` and `positive` parameters to get different colors).

@naglepuff This should be useful in showing just the superpixels from an image.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,6 +1,9 @@
 # Change Log
 
-## 1.16.3
+## 1.17.0
+
+### Features
+- Style functions ([960](../../pull/960))
 
 ### Improvements
 - Reduce rest calls to get settings ([953](../../pull/953))
--- a/docs/tilesource_options.rst
+++ b/docs/tilesource_options.rst
@@ -80,7 +80,7 @@ A band definition is an object which can contain the following keys:
 
   The function parameter can be a single function or a list of functions.  Items in a list of functions can, themselves, be lists of functions.  A single function can be an object or a string.  If a string, this is shorthand for ``{"name": <function>}``.  The function object contains (all but ``name`` are optional):
 
-  - ``name``: The name of a Python module and function that is installed in the same environment as large_image.  For instance, ``large_image.tilesource.utilities.maskPixelValues`` will use the function ``maskPixelValues`` in the ``large_image.tilesource.utilities`` module.  The function must be a Python function that takes a numpy array as the first parameter (the image) and has named parameters or kwargs for any passed parameters and possibly the style context.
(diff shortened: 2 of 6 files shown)


banesullivan:
We should definitely add some examples, like for hillshade as I outlined in https://github.com/girder/large_image/issues/688#issue-1059482283

banesullivan:
I think hillshade is common enough that having a defined function in large-image is warranted

manthey:
> We should definitely add some examples, like for hillshade as I outlined in [#688 (comment)](https://github.com/girder/large_image/issues/688#issue-1059482283)

Definitely.  I'm torn between adding these to `tilesource.utilities` and making a new module `tilesource.stylefuncs`.  We can add that in a follow-up PR.

banesullivan:
+1 for a new module

PR #974, opened 2022-10-03
Add a metadata search mode for items and annotations

manthey:


ant0nsc:
Thanks @manthey for getting that out so quickly ❤️

PR #982, opened 2022-10-18
Fix an issue when compositing multiple frames.

manthey:


manthey:
This fixes an issue that was introduced in https://github.com/girder/large_image/pull/892.

PR #1005, opened 2022-11-29
Add a dicom source.

manthey:
This uses wsidicom to read dicom files.  So far it has only been tested with files that were generated with the wsi2dcm docker from the 1.0.3 deb from https://github.com/GoogleCloudPlatform/wsi-to-dicom-converter, which seems to have issues with some files.

This may be refactored to use a different base library, and internal details should be considered provisional.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,6 +1,9 @@
 # Change Log
 
-## 1.17.4
+## 1.18.0
+
+### Features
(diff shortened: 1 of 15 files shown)


manthey:
As an implementation note, I had looked into using the dicomslide library rather than the wsidicom library.  dicomslide *requires* numerous dicom fields to be set that are not and cannot be set via the wsi2dcm program, and therefore doesn't seem to be an easy solution.

PR #1037, opened 2023-01-26
Apply ICC profiles.

### LINKED ISSUE #856: ICC Color Profiles
For tile sources where it is available, we should explicitly either apply or not apply the ICC profile.

I imagine there would be three options: default, apply, ignore, where default should apply the profile to image outputs but not to numpy outputs.

manthey:
This is done on tiff and openslide tiff-based sources.

Closes #856.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,6 +1,9 @@
 # Change Log
 
-## 1.19.4
+## 1.20.0
+
+### Features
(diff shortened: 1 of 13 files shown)


manthey:
Upper left is without ICC correction, lower right is ICC to sRGB correction.

image

cooperlab:
Bravo!

PR #1051, opened 2023-02-07
Algorithm progression script

annehaley:


examples/algorithm_progression.py
```diff
@@ -0,0 +1,81 @@
...
+    ):
+        altered_data = algorithm(tile['tile'])
+        new_source.addTile(altered_data, tile['x'], tile['y'])
+    return new_source
+
+
+def save_result(source, iteration, output_folder):
+    filename = f'iteration_{iteration}.tiff'
+    filepath = pathlib.Path(output_folder, filename)
+    source.write(filepath, lossy=False)
```
manthey, line 23: Sadly, I think `write` currently needs a string file path and not a pathlib filepath.

manthey:
I think you currently have an odd mix of flexibility and hard-coded values.  I could see making this a more general utility to show things off.  In which case, I think we'd want to:
- properly use argparse or similar to specify input file and other parameters (discussed below)
- specify output directory and/or maybe base output name, so when we have an improved API for output it makes it more obvious
- specify algorithm to run (this could be the package.function)
- use a parameter sweep rather than an iteration (I think iterations will be a rare case, and can be handled via a parameter sweep where the parameter is the number of iterations)
- allow multiple parameters to vary (I'd use itertools.product to get the space that would be used)
- if we have more than one parameter varying, we wouldn't want to include the source image, since I'm not sure where it would be located in a z/t/xy/etc. matrix
- if we do include the source image as an output frame in the multi source, we should probably copy it to the output directory and use the relative path to make the output self contained and not leaking the host machine's pathing information

This might result in a command like:
`algorithm_sweep.py source.svs --output=/tmp/algoutput --function=algorithm.positive_pixel_count --sweep hue_value,0,0.95,0.05 --sample hue_width,0.05,0.1,0.125,0.15,0.175,0.2,0.25` 

In some future fancier version, you'd also specify tile size, overlap, and combination methods.

We currently are doing `new_source = large_image.new(); new_source._addTile(); new_source.write()` for each frame and then a bunch of code to generate a yaml file, which results in a pile of files.  In the future, each parameter set would use the same new source and there wouldn't be any of the yaml generate and a single output file.  When refactoring code, any way to make these changes more obvious would be good as an illustration of the api differences.

manthey:
And, if we want to use positive pixel count to generate images, it should be (extracted from HistomicsTK):
```
class Labels:
    NEGATIVE = 0
    WEAK = 1
    PLAIN = 2
    STRONG = 3


def rgb_to_hsi(im):
    """
    Convert to HSI the RGB pixels in im.

    Adapted from
    https://en.wikipedia.org/wiki/HSL_and_HSV#Hue_and_chroma.
    """
    im = np.moveaxis(im, -1, 0)
    if len(im) not in (3, 4):
        raise ValueError('Expected 3-channel RGB or 4-channel RGBA image; '
                         'received a {}-channel image'.format(len(im)))
    im = im[:3]
    hues = (np.arctan2(3**0.5 * (im[1] - im[2]),
                       2 * im[0] - im[1] - im[2]) / (2 * np.pi)) % 1
    intensities = im.mean(0)
    saturations = np.where(
        intensities, 1 - im.min(0) / np.maximum(intensities, 1e-10), 0)
    return np.stack([hues, saturations, intensities], -1)


def positive_pixel_count(
        image, hue_value=0.83, hue_width=0.15, saturation_minimum=0.05,
        intensity_upper_limit=0.95, intensity_lower_limit=0.05,
        intensity_weak_threshold=0.65, intensity_strong_threshold=0.35):

    image_hsi = rgb_to_hsi(image / 255)
    mask_all_positive = (
        (np.abs((image_hsi[..., 0] - hue_value + 0.5 % 1) - 0.5) <=
         hue_width / 2) &
        (image_hsi[..., 1] >= saturation_minimum) &
        (image_hsi[..., 2] < intensity_upper_limit) &
        (image_hsi[..., 2] >= intensity_lower_limit)
    )
    all_positive_i = image_hsi[mask_all_positive, 2]
    mask_weak = all_positive_i >= intensity_weak_threshold
    mask_strong = all_positive_i < intensity_strong_threshold
    mask_pos = ~(mask_weak | mask_strong)

    label_image = np.full(image.shape[:-1], Labels.NEGATIVE, dtype=np.uint8)
    label_image[mask_all_positive] = (
        mask_weak * Labels.WEAK +
        mask_pos * Labels.PLAIN +
        mask_strong * Labels.STRONG
    )
    color_map = np.empty((4, 4), dtype=np.uint8)
    color_map[Labels.NEGATIVE] = 255, 255, 255, 255
    color_map[Labels.WEAK] = 60, 78, 194, 255
    color_map[Labels.PLAIN] = 221, 220, 220, 255
    color_map[Labels.STRONG] = 180, 4, 38, 255
    ppcimg = color_map[label_image]
    return ppcimg
```

examples/algorithm_progression.py
```diff
@@ -0,0 +1,177 @@
...
+        if param.name != 'data':
+            user_input = input(f'{param.name} ({param.default}): ')
+            if len(user_input.split(',')) == 2:
+                user_range = [float(i.strip()) for i in user_input.split(',')]
+                num_values = int(
+                    input(
+                        f'Enter a number of values to generate in the range {user_range}:'
+                    )
+                )
+                input_params[param.name] = np.linspace(
```
manthey, line 123: Some parameters we will want to range on the semi open-ended interval, e.g. hue-value makes sense to range from [0, 1), while others are conceptually easier to think about in the closed interval, hue_width might range from [0.05, 0.25].  This is a flag to `linspace`.

examples/algorithm_progression.py
```diff
@@ -0,0 +1,177 @@
...
+        i = 0
+        while Path(output_dir).exists():
+            i += 1
+            output_dir = f'{algorithm.__name__}_output_{i}'
+        os.mkdir(output_dir)
+    else:
+        if not Path(output_dir).exists():
+            os.mkdir(output_dir)
+
+    user_provided = input(
```
manthey, line 157: Having user input is fine as a feature, but tedious for reproducible testing.  I'd rather be able to say `--param=hue_value,0,1,250,open --param=hue_width=0.05,0.25,9`.  There is a lot we could include: open-ended/close-ended interval, number of steps versus step size, axis name (and we now can support arbitrary axes), etc., but I think it would suffice that if any algorithm parameters were added on the command line we'd skip asking the user.

annehaley:
With [6d8b2b6](https://github.com/girder/large_image/pull/1051/commits/6d8b2b66689968ec2bc3d6ca00b84f22596a8360), you can now use a command like this to pass in the parameter space for the algorithm:
 
```
python3 algorithm_progression.py ppc [path_to_input_image] -o sample_output -w 5 -p hue_value,0,1,250,open -p hue_width,0.05,0.25,9
```

manthey:
I'm getting an error: `AttributeError: 'Future' object has no attribute 'write'`

annehaley:
@manthey Thanks. I've now discovered that I wasn't properly awaiting futures. However, when I do make the changes to actually run the `apply_algorithm` function asynchonously, I run into the problem of large image sources being unpickleable. They cannot be passed as arguments to a task nor returned from a task if we use 'ProcessPoolExecutor'. 

We can pass them through 'ThreadPoolExecutor' though.

manthey:
> @manthey Thanks. I've now discovered that I wasn't properly awaiting futures. However, when I do make the changes to actually run the `apply_algorithm` function asynchonously, I run into the problem of large image sources being unpickleable. They cannot be passed as arguments to a task nor returned from a task if we use 'ProcessPoolExecutor'.
> 
> We can pass them through 'ThreadPoolExecutor' though.

Sources aren't actually pickleable because they have open file handles.  But paths are pickleable and large_image caches sources internally to avoid reopening files if called multiple times with the same source.  Some of the tile sources block on certain reads if there are multiple threads in the same process because the underlying reader isn't thread safe.  Using processes avoids that bottleneck.

manthey:
And maybe we should make a way to pickle sources (with in the large_image library ) so you wouldn't have to think about it....  Let me look up how we can do that.

annehaley:
Ok, how do you suggest I send the information from the newly-created source (after the algorithm is applied) back in a pickleable form and turn it into a source again?

manthey:
If we parallelize at the algorithm level, then each process would open the original source, apply the specific parameters and create a single output file, and then return the created file path (so it would save the file).

In our hypothetical improved method, we would need to be able to either open the hdf5 or xarray file from multiple processes OR be able to pickle enough information to handle that (but that is future work).

annehaley:
That's a good idea, I'll try that out.

annehaley:
@manthey Could you try this version?

examples/algorithm_progression.py
```diff
@@ -0,0 +1,172 @@
...
+                apply_algorithm,
+                algorithm,
+                input_filename,
+                output_dir,
+                combo,
+                unique,
+            )
+            for combo, unique in zip(param_space_combos, combos_uniques)
+        ]
+        for future in enumerate(concurrent.futures.as_completed(futures)):
```
manthey, line 108: If you use `enumerate`, you are iterating over a tuple `(index, future)`, not just the values in the futures.

examples/algorithm_progression.py
```diff
@@ -0,0 +1,172 @@
...
+    help='A parameter to pass to the algorithm; '
+         'instead of using the default value, '
+         'Pass every item in the number space, '
+         'specified as `--param=param_name,start,end,num_items[,open]`',
+)
+
+
+def apply_algorithm(algorithm, input_filename, output_dir, params, unique_params):
+    source = large_image.open(input_filename)
+    filename = f'{algorithm.__name__}_{"_".join(["{:.2f}".format(v) for v in unique_params])}.tiff'
```
manthey, line 57: `.2f` isn't guaranteed to be enough precision (e.g., --param=hue_value,0,1,256,open` will have collisions.

examples/algorithm_progression.py
```diff
@@ -0,0 +1,172 @@
...
+import algorithms
+import numpy as np
+import yaml
+
+import large_image
+
+VARIABLE_LAYERS = ['z', 'c', 't']
+
+
+argparser = argparse.ArgumentParser(
```
manthey, line 17: It is more common to have all of the argparse stuff into the `if __name__ == '__main__':` conditional, or to make a function that creates the parser and call that from within that conditional.

examples/algorithm_progression.py
```diff
@@ -0,0 +1,172 @@
...
+)
+
+
+def apply_algorithm(algorithm, input_filename, output_dir, params, unique_params):
+    source = large_image.open(input_filename)
+    filename = f'{algorithm.__name__}_{"_".join(["{:.2f}".format(v) for v in unique_params])}.tiff'
+    filepath = Path(output_dir, filename)
+    desc = {'path': filename}
+    for i, v in enumerate(unique_params):
+        desc[VARIABLE_LAYERS[i]] = float(v)
```
manthey, line 61: We actually want the values for `z`, `t`, etc. to be 0-based integers, and not the floating point parameter values.

manthey:
As a basic test, I would expect the following to work:
```
python examples/algorithm_progression.py ppc -o /tmp/sample build/tox/externaldata/sample_Easy1.png --param hue_value,0,1,256,open --param=hue_width,0.05,0.25,9 
python -c "import large_image;ts = large_image.open('/tmp/sample/results.yml');print(ts.sizeX, ts.sizeY, ts.frames)"
```

examples/algorithm_progression.py
annehaley: Can we add arbitrary keys to the description that contain the true (float) values of the params, then? z, t, and c would be unique indexes, but we need some way to attach the true values to each result image.

manthey: Yes -- You can arbitrary keys at the frame level.  There isn't any standard on different WSI formats on how to store actual values.  Maybe add a `parameters` dictionary with `hue_value: 0.123`, etc. in the is example and `parameters_axes` with `hue_value: z` (or `z: hue_value`?).

manthey: Those values will presently be mostly ignored by the reader (ideally we'd eventually try to impose some consistency between tile sources on this process -- OME Tiff has some examples, but no regularity).

annehaley: For example, could the content of `results.yml` look like this?
```
- path: positive_pixel_count_1.00000_0.05000.tiff
  hue_width: 1.0
  z: 1
  hue_value: 0.05
  c: 0
- path: positive_pixel_count_0.00000_0.15000.tiff
  hue_value: 0.0
  z: 0
  hue_width: 0.15000000000000002
  c: 1
- path: positive_pixel_count_1.00000_0.15000.tiff
  hue_width: 0.15000000000000002
  z: 1
  c: 1
- path: positive_pixel_count_0.00000_0.05000.tiff
  hue_value: 0.05
  z: 0
  c: 0
```

manthey: Yes, but I think it'd be better to nest the algorithm values under a subkey and indicate which axes have been used:
```
- path: positive_pixel_count_1_1.tiff
  parameters:
    values:
      hue_value: 1.0
      hue_width: 0.15000000000000002
      saturation_minimum: 0.05
      intensity_upper_limit: 0.95
      intensity_lower_limit: 0.05
      intensity_weak_threshold: 0.65
      intensity_strong_threshold: 0.35
    axes:
      z: hue_value
      c: hue_width
  z: 1
  c: 1
```
Part of my reasoning is that we have expanded what can be specified in the multi-source source in the past, so limiting the frame level keys might be better.

And, if we are sweeping many parameters, we can use custom axis names (add a top-level key `axes` with a list of axes names, then you can use those axes just like `z`, `t`, `xy`, and `c`.  You shouldn't use `x`, `y`, `s`, or `f`.

annehaley: Nevermind, I remembered that large_image does not accept arbitrary keys in each source, so I just put all that information in the description. `results.yml` now looks like this:

```
name: Positive Pixel Count iterative results
description: Positive Pixel Count algorithm performed on /home/anne/Documents/testing_grounds/large_image/data/sample_jp2k_33003_TCGA-CV-7242-11A-01-TS1.1838afb1-9eee-4a70-9ae3-50e3ab45e242.svs,
  where z represents change in hue_value as an index of [0. 1.],c represents change
  in hue_width as an index of [0.05 0.15],saturation_minimum remains unchanged per
  run with a value of 0.05,intensity_upper_limit remains unchanged per run with a
  value of 0.95,intensity_lower_limit remains unchanged per run with a value of 0.05,intensity_weak_threshold
  remains unchanged per run with a value of 0.65,intensity_strong_threshold remains
  unchanged per run with a value of 0.35
sources:
- path: /home/anne/Documents/testing_grounds/large_image/data/sample_jp2k_33003_TCGA-CV-7242-11A-01-TS1.1838afb1-9eee-4a70-9ae3-50e3ab45e242.svs
- path: positive_pixel_count_1_1.tiff
  z: 1
  c: 1
- path: positive_pixel_count_0_1.tiff
  z: 0
  c: 1
- path: positive_pixel_count_1_0.tiff
  z: 1
  c: 0
- path: positive_pixel_count_0_0.tiff
  z: 0
  c: 0

```

This is successful with the checking command you provided below, and the output of that command is `23021 23162 4`.

examples/algorithm_progression.py
annehaley: Changed in [8f12e60](https://github.com/girder/large_image/pull/1051/commits/8f12e60fa621de791b9463118c999c13ddb2e88a)

examples/algorithm_progression.py
annehaley: Changed in [8f12e60](https://github.com/girder/large_image/pull/1051/commits/8f12e60fa621de791b9463118c999c13ddb2e88a)

examples/algorithm_progression.py
annehaley: Changed in [8f12e60](https://github.com/girder/large_image/pull/1051/commits/8f12e60fa621de791b9463118c999c13ddb2e88a)

examples/algorithm_progression.py
```diff
@@ -0,0 +1,179 @@
...
+    new_source.write(str(filepath), lossy=False)
+    return desc
+
+
+def sweep_algorithm(algorithm, input_filename, input_params, output_dir, max_workers):
+    algorithm_name = algorithm.__name__.replace('_', ' ').title()
+    yaml_dict = {
+        'name': f'{algorithm_name} iterative results',
+        'description': f'{algorithm_name} algorithm performed on {input_filename}',
+        'sources': [{'path': input_filename}],
```
manthey, line 87: If we output multiple axes (or even one axis), the input file doesn't really fit in the matrix of output.  Let's _not_ add it.  If we did add it, it should be on some distinct axis location, as otherwise it will just be overwritten by the output at the 0 values.

annehaley: Ok, I'll take it out.

examples/algorithm_progression.py
```diff
@@ -0,0 +1,179 @@
...
+                algorithm,
+                input_filename,
+                output_dir,
+                combo,
+                input_params
+            )
+            for combo in param_space_combos
+        ]
+        for future in concurrent.futures.as_completed(futures):
+            print('Completed run.')
```
manthey, line 114: Can we print the actual number of frames processed?  Maybe maintain a done counter and `print(f'Completed {done}/{len(param_space_combos)}')`

annehaley: Changed in [29902be](https://github.com/girder/large_image/pull/1051/commits/29902be9031c53f83d89011752c5e1c0481e5071)

examples/algorithm_progression.py
```diff
@@ -0,0 +1,181 @@
...
+                apply_algorithm,
+                algorithm,
+                input_filename,
+                output_dir,
+                combo,
+                input_params
+            )
+            for combo in param_space_combos
+        ]
+        num_done += 1
```
manthey, line 114: This needs to be in the `for future` loop.

examples/algorithm_progression.py
```diff
@@ -0,0 +1,181 @@
...
+        required=False,
+        help='Name for a new directory in this location, wherein result images will be stored',
+    )
+    argparser.add_argument(
+        '-w',
+        '--num_workers',
+        required=False,
+        default=4,
+        type=int,
+        help='Number of workers to use for processing alogrithm iterations',
```
manthey, line 41: s/alogrithm/algorithm/

examples/algorithm_progression.py
```diff
@@ -0,0 +1,181 @@
...
+    new_source.addTile(altered_data, tile['x'], tile['y'])
+    new_source.write(str(filepath), lossy=False)
+    return desc
+
+
+def sweep_algorithm(algorithm, input_filename, input_params, output_dir, max_workers):
+    algorithm_name = algorithm.__name__.replace('_', ' ').title()
+    yaml_dict = {
+        'name': f'{algorithm_name} iterative results',
+        'description': f'{algorithm_name} algorithm performed on {input_filename}',
```
manthey, line 86: Can we also add `'uniformSources': True` to this dict?  Doing so tells the multi-source that all sources are structurally similar (which should be the case for this script), so it doesn't have to open *every* source to verify opening the yaml file.

annehaley: Changed in [9d45193](https://github.com/girder/large_image/pull/1051/commits/9d45193e21b46b6121053daf87825898fdbcce1f)

examples/algorithm_progression.py
annehaley: Changed in [9d45193](https://github.com/girder/large_image/pull/1051/commits/9d45193e21b46b6121053daf87825898fdbcce1f)

examples/algorithm_progression.py
```diff
@@ -0,0 +1,182 @@
...
+        os.mkdir(output_dir)
+    else:
+        if not Path(output_dir).exists():
+            os.mkdir(output_dir)
+
+    params = {
+        param.name: (
+            input_params[param.name]
+            if param.name in input_params
+            else algorithms.ALGORITHM_DEFAULT_PARAM_SPREADS[algorithm_code][param.name]
```
manthey, line 173: When we want to sweep just one parameter (say hue_value in the PPC algorithm), we currently have to explicitly say to use a single value for other default params.  I wonder if it would be better to skip using the defaults if any input parameters are given.  @annehaley What do you think?  Otherwise, this is working as desired.

annehaley: Ok I can do that! hanged in [98a33e1](https://github.com/girder/large_image/pull/1051/commits/98a33e194ca9b8d1d95b13880bba24202bb122cd)

manthey (approved):
Excellent.  Thanks for your patience with all of the requested changes.

PR #1052, opened 2023-02-07
Allow turning on or off ICC correction from the plugin settings.

manthey:


manthey:
Since ICC color correction can now be toggled in the config, I think this can be merged as a feature rather than a breaking change.

PR #1064, opened 2023-02-21
Fix how some sources are referenced in setup.py.

manthey:
Fix how sources that don't work with older python verisons are referenced.

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -1,6 +1,5 @@
 import itertools
 import os
-import sys
 
 from setuptools import setup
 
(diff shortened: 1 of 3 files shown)


setup.py
```diff
@@ -45,10 +44,12 @@ def prerelease_local_scheme(version):
 sources = {
     'bioformats': [f'large-image-source-bioformats{limit_version}'],
     'deepzoom': [f'large-image-source-deepzoom{limit_version}'],
+    'dicom': [f'large-image-source-dicom{limit_version} ; python_version < "3.8"'],
```
banesullivan, line 47: Shouldn't this be `python_version >= "3.8"`? That's what it is with what was removed.

As is, this means anyone above Python 3.7 cannot install this source

manthey: Yep.  Fixed and testing.

PR #1065, opened 2023-02-22
Add ipyleaflet representation

### LINKED ISSUE #1059: _repr_html_ for interactive tile viewing in Jupyter
`_repr_html_` is a fantastic way to add an interactive representation style for objects in Jupyuter and we could easily expose a slippy-maps style viewer that fetches tiles from a large-image source.

For example, I implemented this in `localtileserver`: https://localtileserver.banesullivan.com/user-guide/index.html#user-guide

I'd like to wrap up the rasterio source and consolidate a new "GeoBaseTileSource", then think about implementing this as:

1. `_repr_html_` on the base `TileSource` class that shows an slippy maps style image viewer (non-geospatially referenced)
2. `_repr_html_` on the base "GeoBaseTileSource" that has a geospatial reference

## Implementation

We need to launch a webserver to host the endpoints for our JS-based viewer to fetch tiles from. Launching the server isn't too bad and as I've shared in plenty of places, we can do this easily with [server-thread](https://github.com/banesullivan/server-thread) and a simple wsgi/asgi app.

Here is a minimal, yet fairly robust example:

<details>

```py
import io

from fastapi import FastAPI
from server_thread import ServerThread
from starlette.responses import StreamingResponse


def get_server(src):

    app = FastAPI()

    @app.get("/metadata")
    async def get_metadata():
        return src.getMetadata()

    @app.get("/tile/{z}/{x}/{y}.png")
    async def get_tile(z: int, x: int, y: int):
        tile_binary = src.getTile(x, y, z)
        return StreamingResponse(io.BytesIO(tile_binary), media_type="image/png")

    # Launch app in a background thread
    server = ServerThread(app)

    return server
```

With usage looking like:

```py
import large_image
from tileserver import get_server
from ipyleaflet import Map, TileLayer

source = large_image.open('path/to/raster.tif')

server = get_server(source)

t = TileLayer(
    url=f"http://{server.host}:{server.port}/tile/{{z}}/{{x}}/{{y}}.png",
    attribution="Served with large-image",
)

m = Map()
m.add_layer(t)
m
```

</details>

### Complications

An issue arises from gaining access to that webserver when running on managed, deployed environments (think MyBinder, university JupyterLab deployments, Sagemaker, etc.), we'll likely need to rely on [`jupyter-server-proxy`](https://github.com/jupyterhub/jupyter-server-proxy) to proxy the port so that the user's web browser can access this. Thing is, it has to be configured slightly differently in just about every managed environment -- a maintenance and support burden we're facing on PyVista with its webserver-based Jupyter viewer. 

Though, a good chunk of the implementation to support most platforms in a standard way can be taken from `localtileserver`: https://github.com/banesullivan/localtileserver/blob/main/localtileserver/configure.py

### Which viewer do we use?

We could custom bundle a viewer based on GeoJS, CesiumJS, LeafletJS, etc and this wouldn't require any additional dependencies. However, this would be somewhat of "oh neat,... but I can't do anything with that" situation. I think it would behoove us to add an `extra_requires` section for Jupyter that would leverage [`ipyleaflet`](https://ipyleaflet.readthedocs.io/en/latest/) for creating these viewers in Jupyter. This would enable users to add annotations, add additional tile sources, and customize to their heart's desire.

Using `ipyleaflet` will require nailing down the non-geospatial case as it's still a bit weird: https://localtileserver.banesullivan.com/user-guide/non-geo.html

banesullivan:
A first pass at adding an `_ipython_display_` repr to visualize a tile source in Jupyter dynamically

resolves #1059 

I'm having a bit of trouble getting this to work consistently for non-geospatial tile sources

This adds a new `rest.py` module which is leveraged to launch a minimal tile server for a `TileSource` using a tornado web server (the same web server Jupyter uses, thus requiring no extra dependencies).

The `TileLayer` creation is abstracted in `getIpyleafletTileLayer()` so that users can build the tile layer more easily and/or create custom maps.

Screen Shot 2023-02-22 at 11 47 06 AM

```py
import large_image

path = 'oahu-dense.tiff'

source = large_image.open(
    path,
    encoding="PNG",
    projection="EPSG:3857",
)
source
```

```py
from ipyleaflet import Map

t = source.getIpyleafletTileLayer(attribution="My tile source!!")

m = Map(center=source.getCenter(srs="EPSG:4326"))
m.add_layer(t)
m
```

## Potential remaining tasks

- [ ] Add testing
- [ ] Figure out why the tornado enpoints only work in JupyterLab and break classic notebook. Perhaps switch to using a background thread
- [ ] Verify orientation of medical/non-geo images


banesullivan:
@zachmullen and I are experiencing issues where the tornado endpoints here are breaking Classic Notebook, but this works fin in JupyterLab

banesullivan:
@manthey, this is ready for review. There are some quirks and known issues, but I think this is at a point where it is highly valuable/useful.

1. This does not work in Classic Notebook, **only JupyterLab**. We may have to switch to using FastAPI over Tornado for Classic Notebook, but I don't think as many people still use Classic Notebook
2. See the inline comment above about potential improvements to the map/projection
3. Testing is just a smoke test to ensure these functions run and don't error (requests are not made)

manthey:
Let's merge this and refine it in future PRs.

manthey (approved):
Let's make an issue to revisit the coordinate system of non-geospatial data.

Issue #1073, opened 2023-03-02
Handle massive json annotations


manthey:
Currently, we parse annotations by loading the entire annotation into memory.  Usinb an iterative approach for larger files (e.g., via https://github.com/ICRAR/ijson) would be mord scalable.

manthey:
A quick but limited solution to this would be to pick a bigger size than Girder's default of 16MB and read those using the current technique.

PR #1075, opened 2023-03-02
Increase maximum size of json file that will be read for annotations.

manthey:
This is still constrained more than desirable, and could use refactoring.  This is not the whole solution for #1073.


manthey:
Note that prior to #989, we would try to load arbitrarily large json files BUT only from filesystem assetstores.

PR #1077, opened 2023-03-06
Compare xarray sweep and hdf5 sweep methods

annehaley:
```
python3 examples/algorithm_progression/main.py ppc test/test_files/test_RGB_16.png -o tmp/xarray -f xarray -p hue_width,0,1,8 -p hue_value,0,1,4

python3 examples/algorithm_progression/main.py ppc test/test_files/test_RGB_16.png -o tmp/hdf5 -f hdf5 -p hue_width,0,1,8 -p hue_value,0,1,4
```


examples/algorithm_progression/xarray_sweep.py
```diff
@@ -0,0 +1,67 @@
...
+                    source_metadata["sizeX"],
+                    4,
+                )
+            )
+            for tile in source.tileIterator(
+                format=large_image.tilesource.TILE_FORMAT_NUMPY,
+                tile_size=dict(width=2048, height=2048),
+            ):
+                output_data[
+                    tile["x"]: tile["x"] + tile["width"],
```
manthey, line 36: Arrays default to y, x, samples not x, y, samples

annehaley: Changed in [502bcdd](https://github.com/girder/large_image/pull/1077/commits/502bcddb39506680d2ceec423f4f4b2e171dc808)

examples/algorithm_progression/xarray_sweep.py
```diff
@@ -0,0 +1,67 @@
...
+        self.output_files.append(result)
+
+    def complete_storage(self):
+        results = [xr.open_dataarray(filename) for filename in self.output_files]
+        if len(results) > 0:
+            dims = dict(
+                **{k: len(v) for k, v in self.input_params.items()},
+                **results[0].sizes,
+            )
+            composite_xr = xr.DataArray(
```
manthey, line 57: It looks like this is putting everything in memory, so it fails if the results are larger than memory+swap.  Is there a way to ask xarray to not have it all in memory?  My test was to ask for a data set that was larger than my memory size and the process had to be OOM killed.

annehaley: We would have to store the array as zarr to do so:
https://discourse.pangeo.io/t/processing-large-too-large-for-memory-xarray-datasets-and-writing-to-netcdf/1724

annehaley: One of the fundamental problems we're facing here is the idea of editing the same data store from multiple processes. For the multisource sweep, this worked because each iteration writes its own file and the composite file (`results.yml`) is pretty small. When we try this for the xarray or hdf5 sweep, we want to end up with one large file. I haven't found a solution yet that allows chunked access of the file from multiple processes. What do you think is a good pattern to follow?

examples/algorithm_progression/xarray_sweep.py
```diff
@@ -0,0 +1,87 @@
...
+            **{
+                'y': self.source_metadata['sizeY'],
+                'x': self.source_metadata['sizeX'],
+                'c': 4,
+            },
+        )
+
+        ds = xr.Dataset(
+            {
+                'results': (
```
manthey, line 23: We'd ideally like to NOT allocate all the space up front.  Can we create this with expandable dimensions?

examples/algorithm_progression/xarray_sweep.py
```diff
@@ -0,0 +1,87 @@
...
+        self.zarr_path = Path(self.output_dir, 'results.zarr')
+        ds = self.create_ds()
+        ds.to_zarr(self.zarr_path, mode='w', compute=False)
+
+    def apply_algorithm(self, param_combo):
+        current_coords = {
+            k: [param_combo[i]] for i, k in enumerate(self.input_params.keys())
+        }
+        output_data = np.empty(
+            (
```
manthey, line 50: We really don't want to allocate an array the size of our image -- some don't fit into memory.

examples/algorithm_progression/xarray_sweep.py
```diff
@@ -0,0 +1,87 @@
...
+        output_ds['results'].loc[current_coords] = output_data
+
+        current_index_region = {
+            k: list(v.values).index(current_coords[k][0])
+            for k, v in output_ds['results'].coords.items()
+        }
+        current_index_region_slice = {
+            k: slice(i, i + 1) for k, i in current_index_region.items()
+        }
+        output_ds.isel(current_index_region_slice).to_zarr(
```
manthey, line 75: This is astonishingly slow.  When I run `main.py ppc -o ppc_easy_x sample_Easy1.png --param hue_value,0,1,1000,open --format xarray` it is taking on the order of 15 seconds per parameter set.

examples/algorithm_progression/hdf5_sweep.py
```diff
@@ -0,0 +1,86 @@
...
+
+        self.filepath = Path(self.output_dir, 'results.h5')
+        self.iteration_dims = {
+            'y': self.source_metadata['sizeY'],
+            'x': self.source_metadata['sizeX'],
+            'c': 4,
+        }
+        self.param_dims = {k: len(v) for k, v in self.input_params.items()}
+        self.layout = h5py.VirtualLayout(
+            shape=tuple(dict(**self.param_dims, **self.iteration_dims).values()),
```
manthey, line 24: We can make the h5py expandable and not need to know what size it will be before hand by starting with a shape of all 1s for the param_dims and something modest for the iteration_dims, and then specify a `maxshape=tuple([None] * (total number of dimensions))` and specify `chunks=True`.  If you start with an axis length of 1, h5py will chunk that axis in chunks of 1, which is super inefficient for the spatial dimensions.  We know the output image size, but we might not in other contexts, so we probably want to make the minimum x / y / bands be something like 1024 / 1024 / 4, and (if we make this a tile sink) track the actual used values.

I haven't used the VirtualLayout option, but `h5py.File(<path>, 'w').create_dataset('images', shape=..., maxshape=..., chunks=True)`.  You seemingly can't open h5py files for writing from multiple processes, so we might still need to collect each frame as a separate h5py file and then combine them at the end.

examples/algorithm_progression/xarray_sweep.py
annehaley: Changed in [980bc50](https://github.com/girder/large_image/pull/1077/commits/980bc50d5ca759524577d1656b72dfa62f6aab96)

examples/algorithm_progression/xarray_sweep.py
annehaley: I'm not able to replicate this, I'm getting expected speeds. Is your copy of `Easy1.png` about 3 MB? And are you noticing this 15 second gap in between each print of "Completed {i} of {n} runs"?

examples/algorithm_progression/xarray_sweep.py
annehaley: Changed in [980bc50](https://github.com/girder/large_image/pull/1077/commits/980bc50d5ca759524577d1656b72dfa62f6aab96)

PR #1086, opened 2023-03-21
New Frame selector Vue component

### LINKED ISSUE #498: Better frame selection
Currently, on a multiframe image, we show a simple control for selecting the frame:
image
In many instance we have more information and know that the frames have further information: Z (z-slice), T (time), XY (xy location, also called view), and C (channel)  values that each have a range.  When this is the case, instead of having a single slider for frame, it would be nicer to have multiple sliders.

Further, for single-channel frames where the is a known C range, it would be good to have an option to composite multiple C frames together using false color.  This could be done, for instance, by having a radio button picking between the slider and a drop-down list of channels that could be checked on or off.  As a further refinement, either as a whole or individually, channels could have composition options.  These would include:
- range to map to the false color channel, which could be:
  - full-range
  - non-zero histogram range
  - extended non-zero histogram range (where the histogram excluding a small amount at either end is ranged to pixel values 1-254 so that the extremes fall on 0 and 255 even if they are far from the rest of the range)
  - manually specified
- color to map the channel to

### LINKED ISSUE #1108: Put fixed width on labels in frame selector

### LINKED ISSUE #1111: Increase z value of color picker

annehaley:
Resolves #498, resolves #1108, resolves #1111

Improvements for future PRs:
- Histograms
- Save configurations?
- Reset button


manthey:
@annehaley See the changes I made and the comments on the last commit.  If you think this would be clearer rebased onto master, please do so.

annehaley:
@manthey thank you for fixing `geojs.js`! It works smoothly now. I'll rebase and get tests passing, then I'll throw it back to you for review

annehaley:
@manthey could you help me figure out why the following failure is happening?

```
FAILED girder/test_girder/test_web_client.py::testWebClientNoStream[imageViewerSpec.js]
```

manthey:
> @manthey could you help me figure out why the following failure is happening?
> 
> ```
> FAILED girder/test_girder/test_web_client.py::testWebClientNoStream[imageViewerSpec.js]
> ```

I used `URLSearchParams` to extract a value to help with cache control.  Our client tests run on an older javascript framework which apparently doesn't define this but since there is a timing factor, it only sometimes fails.  Let me switch how this is done or add a polyfill.

girder/girder_large_image/web_client/vue/components/CompositeChannels.vue
```diff
@@ -0,0 +1,184 @@
+<script>
+import { Chrome } from 'vue-color';
+
+const CHANNEL_COLORS = {
```
manthey, line 4: For channel colors, we want to have a dictionary of known channel names with predefined colors (see https://github.com/Kitware/UPennContrast/blob/master/src/store/model.ts#L539).  Eventually this could be overridden by the folder-level config (via the .large_image_config.yaml mechanism, but that can be a future PR).  If the channel name isn't present in our map, we would fall back to a list that is much more ordered on primary color channels (https://github.com/Kitware/UPennContrast/blob/master/src/store/model.ts#L474 is an simple way to have such as list, but maybe it could be generated programmatically).  The idea is that the channels would use as independent as possible colors.

girder/girder_large_image/web_client/vue/components/CompositeChannels.vue
```diff
@@ -0,0 +1,184 @@
...
+    background-color: #000;
+    cursor: pointer;
+}
+.picker-offset {
+    position: absolute;
+    z-index: 100;
+    margin-left: 50px;
+}
+.table-container {
+    max-height: 200px;
```
manthey, line 178: I think I'd rather have a bigger max-height here.  five channels requires scrolling, which is not that uncommon (the most I've seen is 60 channels).

manthey:
I don't want to merge anything that shows unfinished functionality.  So, we either need to handle the unfinished parts (e.g., bands) or finish them.  Once we've done that and addressed my color comments, we could merge this and leave everything else to future PRs.

Ultimately, I haven't seen real-world data sets that have both more than three bands AND channels, so I think that we could have our modes be "Sliders / Composite Frames / Composite Bands".  Composite Bands could be an option even on single band (greyscale) images with only one frame, since that would be a way of exposing the min/max controls.

For bands, the colors should default based on the band interpretation (red/green/blue/gray) and then use unused colors for other interpretations (including alpha).  The default initial checked bands should just be the first three bands (as opposed to channels where we default to all channels being on and composited).

When we use this in the HistomicsUI context, we'll want these controls to fit in one of the side panels -- we'll be able to use CSS to adjust this.  We just need to be aware that controls will be tight (more relevant as we add histograms and options for min/max besides numbers).

The use of "0" as undefined for min/max isn't appropriate.  Technically, the unset value can be different than 0 for min (and is determined based on datatype for max), and with float-based samples, 0 might actually be a sensible max in some odd case.  We eventually want to have a mode for min and max that would be "default/value/min(or max)" where "default" wouldn't use a number, "value" would be the current mode (but there can be data where this is floating point on a scale of [0,1] or [-1,1] or other values, so the controls can't restrict to integers unless we check that makes sense), and "min" or "max" would let you specify a numerical value from [0,1] (or maybe from [0%, 100%]) which would be the amount of the histogram to exclude from the ends (e.g., "min:0.002" sets the minimum such that 0.2% of the samples are below the picked value).

And, for non-composited axes, we'll eventually want a "max-merge" option (which would disable the slider).  This might conceptually have other choices (so maybe it is a drop-down control of slider/max-merge rather than a checkbox).

When we do show this histogram, it is relatively cheap to get the histogram for all the bands of the current frame.  For multiple channels, we end up needing to currently ask for the histogram for all channels at the current frame position.  Since each frame has its own histogram, we might want to overlay two histograms -- one of the current frame/channel (possibly applying max-merge) and one of the combined histograms of all frames with that channel.  Getting that combined histogram is initially expensive in time (and when we have it, we might want an additional min/max mode of combined min/combined max).  We'll want to show the min/max values on the histogram as vertical bars.  

And, in my list of wants, we may eventually want a checkbox to enable/disable all channels (or bands) so that you can quickly re-enable them or go to a single channel/band.

manthey:
When we construct the style, it is somewhat more efficient to specify the base frame as the `frame` parameter and the style with `framedelta` rather than `frame` per band (see https://girder.github.io/large_image/tilesource_options.html#style).

girder/girder_large_image/web_client/vue/components/CompositeChannels.vue
annehaley: Changed in [560ec8e](https://github.com/girder/large_image/pull/1086/commits/560ec8e65cfdca9193753d0fdf427867352c3eb7)

girder/girder_large_image/web_client/vue/components/CompositeChannels.vue
annehaley: Changed in [560ec8e](https://github.com/girder/large_image/pull/1086/commits/560ec8e65cfdca9193753d0fdf427867352c3eb7)

annehaley:
> When we construct the style, it is somewhat more efficient to specify the base frame as the `frame` parameter and the style with `framedelta` rather than `frame` per band (see https://girder.github.io/large_image/tilesource_options.html#style).

Could you provide an example? I don't understand what this would look like from the documentation.

manthey:
I see some linting errors in the latest commit.  Running `tox -e format,formatclient` should resolve most such issues.

manthey:
> > When we construct the style, it is somewhat more efficient to specify the base frame as the `frame` parameter and the style with `framedelta` rather than `frame` per band (see https://girder.github.io/large_image/tilesource_options.html#style).
> 
> Could you provide an example? I don't understand what this would look like from the documentation.

Sure.  Suppose you have four channels and a z index with a stride of 4.  You might render 3 of the channels at z = 0 like so:
```
style={"bands":[{"frame":0,"palette":"#0000FF"},{"frame":2,"palette":"#FF0000"},{"frame":3,"palette":"#FF8000"}]}
```
and z = 2 as
```
style={"bands":[{"frame":8,"palette":"#0000FF"},{"frame":10,"palette":"#FF0000"},{"frame":11,"palette":"#FF8000"}]}
```
but, the style changes each time you change z, which makes for some extra book keeping in the server.

Instead, for z = 0, specifying
```
frame=0
style={"bands":[{"framedelta":0,"palette":"#0000FF"},{"framedelta":2,"palette":"#FF0000"},{"framedelta":3,"palette":"#FF8000"}]}
```
and for z = 2, exactly the same style but `frame=8`.

annehaley:
@manthey Thank you for the explanation. See [369a1d9](https://github.com/girder/large_image/pull/1086/commits/369a1d9d37583d561e845006c25cb6360e426562) for the change.

girder/girder_large_image/web_client/views/imageViewerSelectWidget.js
```diff
@@ -65,6 +66,18 @@ var ImageViewerSelectWidget = View.extend({
         return this;
     },
 
+    _createVueModal(imageMetadata, frameUpdate) {
```
manthey, line 69: Why is this called `_createVueModal` when it doesn't create a modal dialog?  Perhaps it should be renamed to `_createVue`?

annehaley: Sure. This was a remnant I copied over from the previous work. Changed in [e19b401](https://github.com/girder/large_image/pull/1086/commits/e19b401793b7e963a18dfdc8e843041fde15cd67)

girder/girder_large_image/web_client/vue/colors.js
```diff
@@ -0,0 +1,87 @@
...
+    '#FFC040',
+    '#FF40C0',
+    '#40FFC0',
+    '#C0FF40',
+    '#C040FF',
+    '#40C0FF'
+];
+
+export function getCompositeLayerColor(layerName, usedColors) {
+    if (layerName in CHANNEL_COLORS) {
```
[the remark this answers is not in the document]
annehaley: I saw this problem appear and my solution was to sort the array before assigning colors:
`.sort((channel) => !Object.keys(CHANNEL_COLORS).includes(channel))`
The problem did not reappear after this.

girder/girder_large_image/web_client/vue/components/CompositeChannels.vue
```diff
@@ -0,0 +1,189 @@
...
+                }
+                styleArray.push(styleEntry);
+            });
+            this.$emit('updateStyle', {bands: styleArray});
+        },
+    },
+    mounted() {
+        const usedColors = []
+        this.compositeChannelInfo = Object.fromEntries(this.channels
+        .sort((channel) => !Object.keys(CHANNEL_COLORS).includes(channel))
```
manthey, line 85: `sort` takes a compare function that takes two parameters (`a` and `b`).  I'm not sure what only treating one value does, but the sort doesn't work the way we want.

manthey:
If we don't have channel names, the compositing controls don't appear.

girder/girder_large_image/web_client/vue/colors.js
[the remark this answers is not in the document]
annehaley: Changed in [6ab53dd](https://github.com/girder/large_image/pull/1086/commits/6ab53dd5c4abcca1a4ed5f7ea47b979ece5b8fc9)

annehaley:
> If we don't have channel names, the compositing controls don't appear.

Fixed in [4950ded](https://github.com/girder/large_image/pull/1086/commits/4950dedd33646f2875a9d49b7818761a331faf9f)

girder/girder_large_image/web_client/vue/components/FrameSelector.vue
```diff
@@ -0,0 +1,194 @@
...
+        },
+    },
+    mounted() {
+        if (
+            (!this.imageMetadata.channels || this.imageMetadata.channelmap)
+            && Object.keys(this.imageMetadata.IndexRange).includes('IndexC')
+        ) {
+            this.imageMetadata.channelmap = Object.fromEntries(
+                [...Array(this.imageMetadata.IndexRange['IndexC']).keys()].map(
+                    (i) => [`Channel_${i}`, i]
```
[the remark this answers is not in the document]
annehaley: Ok, I'll just force push the change on the same commit.

girder/girder_large_image/web_client/vue/components/FrameSelector.vue
```diff
@@ -0,0 +1,194 @@
...
+            });
+            const useStyle = this.currentModeId === 1
+                && this.currentChannelCompositeModeId === 1
+            let style = useStyle ? this.style : undefined
+            this.frameUpdate(frame, style);
+        },
+    },
+    mounted() {
+        if (
+            (!this.imageMetadata.channels || this.imageMetadata.channelmap)
```
manthey, line 76: I think this was supposed to be `|| !this.imageMetadata.channelmap)`

manthey:
If you have enough channels that you need to scroll, the color pickers appear in the wrong place instead of scrolling with the channels (and, for an item with 40 channels, a select all/none would be really handy)

girder/girder_large_image/web_client/vue/components/FrameSelector.vue
annehaley: Thanks for catching that. Changed in [81b50f0](https://github.com/girder/large_image/pull/1086/commits/81b50f0462293ec3b4e530bf95056dffe608ca04)

annehaley:
@manthey Could you take a look at [81b50f0](https://github.com/girder/large_image/pull/1086/commits/81b50f0462293ec3b4e530bf95056dffe608ca04) so far? I left a couple TODOs where I didn't know how to proceed.

girder/girder_large_image/web_client/vue/components/CompositeLayers.vue
```diff
@@ -68,63 +68,65 @@ export default {
...
+            ).filter((layer) => layer.enabled);
             const styleArray = []
-            activeChannels.forEach((channel) => {
-                const styleEntry = {
-                    frameDelta: channel.number,
-                };
-                if (channel.falseColor) {
-                    styleEntry['palette'] = channel.falseColor;
+            activeLayers.forEach((layer) => {
+                // TODO: what should each style entry look like for band compositing?
```
manthey, line 98: Instead of a `"framedelta"` parameters (technically it could be in addition to rather than instead of), you use a `"band"` parameter.  Its value is either the unique interpretation string or the 1-based band number.  See https://girder.github.io/large_image/tilesource_options.html#style.

annehaley: Thanks. Changed in [1ba2522](https://github.com/girder/large_image/pull/1086/commits/1ba252219be40e7cc804ad437c32977a10bf958c)

girder/girder_large_image/web_client/vue/components/FrameSelector.vue
annehaley: Changed in [81b50f0](https://github.com/girder/large_image/pull/1086/commits/81b50f0462293ec3b4e530bf95056dffe608ca04)

manthey:
When I have the axis mode and slide the channel slide to a non-zero value and then switch to composite channels, it still adds frame = <channel based value> so it ends up with the wrong frames being composited or fails if the last channel is picked.

manthey:
Ideally, when we switch between frame and axis or vice versa we would preserve the view (that is frame is directly equivalent to a set of axis values and vice versa).

annehaley:
@manthey is [1b15871](https://github.com/girder/large_image/pull/1086/commits/1b15871b724f0c65fd1d7027cbebd78d7fcc49ed) what you mean?

girder/girder_large_image/web_client/views/imageViewerWidget/geojs.js
```diff
@@ -67,43 +67,20 @@ var GeojsImageViewerWidget = ImageViewerWidget.extend({
...
-                url: this._getTileUrl('{z}', '{x}', '{y}', {encoding: 'PNG', projection: 'EPSG:3857'}),
-                useCredentials: true,
-                maxLevel: this.levels - 1
-            };
-            // the metadata levels is the count including level 0, so use one
-            // less than the value specified
+        var params = {
+            useCredentials: true,
+            autoshareRenderer: false,
+            keepLower: false,
```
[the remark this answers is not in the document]
annehaley: I had included this because it was in the `params` object on line 98 previously, but now I understand that this was only intended for the geospatial cases. I can move this to the geospatial `if` block instead.

manthey:
If you think my latest changes are good and CI passes, go ahead and merge.

manthey (approved):

PR #1089, opened 2023-03-27
Add events to feature, not layer

naglepuff:
Previously, when rendering overlay layers, several geojs mouse events were added to the layer. This would cause double-firing of these events. Now, the events are only added to the non-quad features on the layer to prevent the double firing.


manthey:
This was a bug in GeoJS.  Fixed here: https://github.com/OpenGeoscience/geojs/pull/1265

manthey:
Closing this as it was fixed in GeoJS.

PR #1115, opened 2023-04-18
rasterio source module

banesullivan:
Supersedes #927 -- making PR from a dedicated branch in this repo to make my life easier

This branch implements a new tile source module based on [rasterio](https://rasterio.readthedocs.io/en/stable/) to simplify the usage of geospatial tile sources where installing GDAL might not be easy.

This further refactors some of the existing GDAL and Mapnik tile sources so that they better share a base class of common geospatial code and similarly, the tests for these modules have been refactored to verify that all three source modules produce the same results.

- Limitation: TILED outputs can only be done when the source is opened with a projection
- Mapnik issue: see `testPixel` and `testTileLinearStyleFromGeotiffs` in `test/test_source_mapnik.py`. These are the only two remaining tests with differing results from the GDAL and rasterio sources. The styling for mapnik needs to be fixed (well maybe the styling should be fixed for all sources), then these tests should be removed to let the base class tests run.
-  `rgba_geotiff.py`: this is a pesky test file -- I will follow up in a separate PR with details on why. Currently only the GDAL source can handle this file and files like it.
- Cross testing: I've cross tested with `django-large-iamge` to verify the GDAL source is still in working order and to verify the `rasterio` source *works as expected* (still missing `TILED` `getRegion()` support)
- It's not recommended using the rasterio source without a projection -- though most things should work.

Still to address:

- [x] `large_image.open()` may not prefer the new rasterio source
- [x] Issues with CI testing (all passing locally)


banesullivan:
@manthey, this is ready to begin reviewing! Note that CI isn't running the full test suite for some reason

banesullivan:
Though I do want to continue to try to clean up the code and make sure there isn't duplicate code between the new `GDALBaseFileTileSource` and the Mapnik/GDAL/rasterio sources so significant changes still likely to come

large_image/tilesource/geo.py
```diff
@@ -0,0 +1,519 @@
...
+import pyproj  # TODO: import issues
+
+from large_image.cache_util import CacheProperties, methodcache
+from large_image.constants import TileInputUnits
+from large_image.exceptions import TileSourceError
+
+from .base import FileTileSource
+from .utilities import JSONDict, getPaletteColors
+
+TileInputUnits['projection'] = 'projection'
```
manthey, line 12: Let's just move these synonym definitions into the main constants.py file.

large_image/tilesource/geo.py
```diff
@@ -0,0 +1,519 @@
...
+        }
+        gdal_options.update(options)
+        vsi = f'/vsicurl?{urlencode(gdal_options)}'
+    return vsi
+
+
+class GeoBaseFileTileSource(FileTileSource):
+    """Abstract base class for geospatial tile sources."""
+
+    geospatial = True
```
manthey, line 48: We want this to be ` _geospatial_source = True`.  Remember, we changed the internal flag name so that the `geospatial` property was if the opened source had geospatial information.

sources/rasterio/large_image_source_rasterio/__init__.py
```diff
@@ -0,0 +1,1078 @@
...
+    from importlib.metadata import version as _importlib_version
+except ImportError:
+    from importlib_metadata import PackageNotFoundError
+    from importlib_metadata import version as _importlib_version
+try:
+    __version__ = _importlib_version(__name__)
+except PackageNotFoundError:
+    # package is not installed
+    pass
+
```
manthey, line 55: We want to filter out some additional warnings (or suppress them just where they occur), probably with 
`warnings.filterwarnings('ignore', category=rio.errors.NotGeoreferencedWarning, module='rasterio')`

manthey:
We should add `rio-cogeo` to some extras_require flag and have that installed in the tox test environment.  Maybe in an `all` key.  This would get added to the various requirements...txt files (as `sources/rasterio[all]` and in the core setup.py).  See the pil source for an example.

sources/rasterio/large_image_source_rasterio/__init__.py
```diff
@@ -0,0 +1,1078 @@
...
+            # set the large_image path
+            self._largeImagePath = self._getLargeImagePath()
+
+            # open the file with rasterio and display potential warning/errors
+            with self._getDatasetLock:
+                try:
+                    self.dataset = rio.open(self._largeImagePath)
+                except RasterioIOError:
+                    raise TileSourceError('File cannot be opened via rasterio.')
+                except FileNotFoundError:
```
manthey, line 128: When the file doesn't exist, this isn't the code path it is following.  Maybe in the section where we catch RasterIOError, we add
```
            if not os.path.isfile(self._largeImagePath):
                raise TileSourceFileNotFoundError(self._largeImagePath) from None
```

sources/rasterio/large_image_source_rasterio/__init__.py
```diff
@@ -0,0 +1,1078 @@
...
+                    raise TileSourceFileNotFoundError(self._largeImagePath)
+
```
manthey, line 130: We probably want to explicitly refuse to open netCDF files until we test appropriately with them  Maybe something like
```
        if self.dataset.driver == 'netCDF':
            raise TileSourceError('netCDF file will not be read via GDAL source')
```

manthey:
Awesome.  With a few changes as mentioned, the tests all pass.  The major rasterio code section that isn't covered is asking for `encoding='TILED'` output and the function `pixelToProjection`  that is called from it.

.gitignore
```diff
@@ -64,3 +71,6 @@ docs/source/*
 !docs/source/*.py
 !docs/source/*.rst
 .pre-commit-config.yaml
+
+*.ipynb
+NOTES.md
```
banesullivan, line 76: These are from my dev environment. Maybe remove?

manthey (approved):

banesullivan:
Recommend merging as is. Though there will need to be follow-up work to:

- More code cleanup
- Ensuring `large_image.open()` properly chooses `rasterio` source (should always have lower priority than GDAL though)
- Investigating slight differences in produced thumbnails/tiles -- seems like a difference in the sampling algorithm choices

Otherwise, this is generally complete

banesullivan:
@12rambau, thank you for getting this started!! You handled a ton of the legwork and made it possible for us to come in and finalize this. We'll tag a release and this should ship shortly.

manthey:
🎉

12rambau:
amazing I was completly unable to finalizing it due to work constraints from my side, I'm super glad you made it !

PR #1133, opened 2023-04-25
Allow putting a config file.

manthey:


manthey:
@naglepuff This can be use from curl via something like `curl '<server>/api/v1/folder/<folder id>/yaml_config/.histomicsui_config.yaml' -X 'PUT' -H 'Girder-Token: <token>' -H "Content-Type: text/yaml" --data-binary @sample.yaml`

PR #1136, opened 2023-04-28
Add hidden divs for frame_id and frame_styling

annehaley:


girder/girder_large_image/web_client/vue/components/FrameSelector.vue
```diff
@@ -135,6 +137,8 @@ export default Vue.extend({
 
 <template>
     <div class="image-frame-control-box">
+        <div id="frame_id" style="display: none;">{{ currentFrame }}</div>
```
manthey, line 140: I'd like to propose different element IDs for these: `current_image_frame` and `current_image_style`.

manthey:
In trying this out, I realize that in slicer_cli_web's code, we had imagined that we would be getting the value of an input control rather than the contained text of a div.  That could be changed and then this will work as is.

manthey (approved):

PR #1144, opened 2023-05-03
Add tile source _dtype attribute

annehaley:


sources/deepzoom/large_image_source_deepzoom/__init__.py
```diff
@@ -113,6 +114,7 @@ def getTile(self, x, y, z, pilImageAllowed=False, numpyAllowed=False, **kwargs):
             overlap if x else 0, overlap if y else 0,
             self.tileWidth + (overlap if x else 0),
             self.tileHeight + (overlap if y else 0)))
+        self._dtype = numpy.array(tile)
```
banesullivan, line 117:
```suggestion
        self._dtype = numpy.array(tile).dtype
```

banesullivan (commented):
This would be nice to have; curious what's motivating this?

manthey:
> This would be nice to have; curious what's motivating this?

When we show controls for picking the minimum and maximum values for a band, it would be nice to be intelligent about what values are allowed.  uint8 data should be limited to [0,255]; uint16 to [0,65535], float should suggest from [0,1].  We can glean this information from the histogram data, but that is expensive to compute.

banesullivan:
Ah, makes sense! One request, can we put this in the metadata?

manthey:
> Ah, makes sense! One request, can we put this in the metadata?

Yes.  That is one of the goals.

dgutman:
Not sure how this is implemented;  is the _dtype computed during the initial file import, or is this done lazily so it's computed the first time the tileInfo is requested, and then cached?

Depending on performance, may also want to have a maintenance function that could pre-compute this at a folder / item level to prevent the user having to wait for this the first time an image is loaded.

large_image/tilesource/base.py
```diff
@@ -202,6 +203,27 @@ def _setStyle(self, style):
...
         return self._style
 
+    @property
+    def dtype(self):
+        if not self._dtype:
+            with self._styleLock:
+                if not hasattr(self, '_skipStyle'):
+                    self._setSkipStyle(True)
+                    try:
+                        sample, format = self.getRegion(
```
manthey, line 213: This should be
```
sample, format = self.getRegion(
                            region=dict(left=0, top=0, width=1, height=1),
                            format=TILE_FORMAT_NUMPY)
```
Setting `left` and `right` both to zero would get a zero-width result which might invoke a quick response with no data and therefore some default dtype (probably uint8); we want `left` and `top`.  The region endpoint uses `regionWidth` and `regionHeight`, but the actual region API call uses `width` and `height` as a property of the region parameter.  Specifying the output width and height would be done via `output=dict(maxWidth=1, maxHeight=1)`, but is unnecessary since we are just getting one pixel anyway.

sources/bioformats/large_image_source_bioformats/__init__.py
```diff
@@ -638,6 +638,7 @@ def getTile(self, x, y, z, pilImageAllowed=False, numpyAllowed=False, **kwargs):
             retile[0:min(tile.shape[0], finalHeight), 0:min(tile.shape[1], finalWidth)] = tile[
                 0:min(tile.shape[0], finalHeight), 0:min(tile.shape[1], finalWidth)]
             tile = retile
+        self._dtype = tile.dtype
```
manthey, line 641: Rather than modifying any tile source's `getTile`, we could add code to tilesource/base.py's `_outputTile` method that, before applying a style, would do something like
```
if self._dtype is None:
    if tileEncoding == TILE_FORMAT_NUMPY:
        self._dtype = tile.dtype
    elif tileEncoding == TILE_FORMAT_PIL:
        self._dtype = numpy.uint8 if ';16' not in tile.mode else numpy.uint16
    else:
        self._dtype = _imageToNumpy(tile)[0].dtype
```

large_image/tilesource/base.py
```diff
@@ -1638,6 +1661,9 @@ def getMetadata(self):
             'magnification': mag['magnification'],
             'mm_x': mag['mm_x'],
             'mm_y': mag['mm_y'],
+            # Use private attribute; public property
+            # incurs calcuation cost if _dtype is still None
+            'dtype': self._dtype,
```
manthey, line 1666: We should measure just what the time impact is on our test set (that is, on all the files in the externaldata and test/test_files locations) and decide if the performance hit is significant.

manthey (changes requested):

large_image/tilesource/base.py
banesullivan: For local files, I suspect the performance hit will be marginal. I'm concerned about the performance hit for remote files with GDAL-based sources when going through the [virtual filesystem mechanisms](https://gdal.org/user/virtual_file_systems.html) -- and especially so if the image isn't cloud optimized as then `getMetadata()` could require downloading the entire image

large_image/tilesource/base.py
annehaley: Fixed in [e015def](https://github.com/girder/large_image/pull/1144/commits/e015deff103cedccef31b6f1ed9d634240e2b2ae)

sources/bioformats/large_image_source_bioformats/__init__.py
annehaley: Changed in [4f41f62](https://github.com/girder/large_image/pull/1144/commits/4f41f620606f4dc3408a1102341eec65d3ec3e41)

large_image/tilesource/base.py
annehaley: I've started a timing test result sheet here: https://docs.google.com/spreadsheets/d/1gY-VMdph4RxI5dsw1i0y5SSaZ8oaAJ0lZZ_Fi97gd9g/edit?usp=sharing (kitware access only)

manthey:
The CI is still failing because the DummyTileSource doesn't have a `_classkey` (deliberately, as it isn't eligible for styling or caching).  I'll push a change to handle that.

manthey (approved):

PR #1175, opened 2023-05-26
Histogram range editor

### LINKED ISSUE #1109: Histogram in frame selector

### LINKED ISSUE #1117: More min/max options in frame controls
We want to add the following features:
- [ ] show the frame's actual min and max value (we can get these from the histogram), maybe as a tool tip
- [ ] Support specifying `min` or `max` and `min:<value>` or `max:<value>` in the controls.  The min and max could have a drop down or toggle to switch between "value" and "min" (or "max").  In value mode, it would behave as it currently does.  In min/max mode, the editable box would be on a range of [0, 1) and be the value specified after `min:` or `max:`
- [ ] Add a option to set the min and max for all channels / bands.  This could be the same controls as per channel and band.  We could either have it so if this is unset the individual channel/band values are used OR so if individual channel / bands are unset this is used.  I'm not sure which would be more intuitive (the second would allow a default value with the ability to override per channel).

### LINKED ISSUE #1122: Frame controls should keep values when changing mode
Some modes have equivalents which should track together if possible.  For instance, frame = 4 might be the same as channel 1, z 1.  Switching between axes mode and frame mode should not visually change the displayed image.

If you switch to a compositing mode, then to frame mode and back to compositing mode, your compositing settings are reset.  They should remain if you haven't navigated away from and back to the item page.

annehaley:
Resolves #1117, resolves #1109, resolves #1122

Histograms include a mode to "exclude distribution tails". When this mode is enabled, the inputs for min and max become percentages from the ends. These automatically chop off 1% from each end (but if a bucket includes more than 1%, it may end up as more - in the below example, the minimum excludes 2%). The user can still change these values.
image

Histograms are drawn with some logic to shorten the height of the max frequency if it eclipses all other values.

From @manthey via Slack:

> let v0 be the highest value in the histogram and v1 be the second highest.  Make the top of the histogram min(v0, v1 * 1.5), so that the second highest value is 2/3 the total height unless making it so would make the highest value less than the total height.

An example without max frequency shortening:
image

The same histograms drawn with max frequency shortening:
image


girder/girder_large_image/web_client/vue/components/FrameSelector.vue
```diff
@@ -172,10 +174,27 @@ export default Vue.extend({
             />
         </table>
 
-        <div v-if="currentModeId > 1" class="image-frame-simple-control">
+        <!-- Use composite layers component twice so state for each one is maintained while invisible -->
+        <!-- Use styling instead of v-if to make each invisible so that the components are not unmounted -->
+        <div  class="image-frame-simple-control" style="width: 500px">
```
manthey, line 179: Please don't use inline styles -- they are hard to override when we reuse the component elsewhere  (like in HistomicsUI)

annehaley: Oh thanks, I hadn't meant to leave that in there. I wanted to see how it would look when squished for HistomicsUI. Removed in [4f10794](https://github.com/girder/large_image/pull/1175/commits/4f10794f9e22a0bf795e7c544b2af064834579fd)

girder/girder_large_image/web_client/vue/components/CompositeLayers.vue
```diff
@@ -57,6 +58,44 @@ export default {
...
+                // layers are channels; each layer has a frame delta
+                this.layers.forEach((layer) => {
+                    restRequest({
+                        type: 'GET',
+                        url: 'item/' + this.itemId + '/tiles/histogram',
+                        data: {
+                            frame: this.currentFrame + this.compositeLayerInfo[layer].framedelta,
+                            width: 1024,
+                            height: 1024,
+                            bins: 512,
```
manthey, line 74: We also need `resample: false` in the query, or the scaling step may end up with histograms on 8-bit data even when the source is not 8-bit.  We might want to have a single place for these parameters rather than repeat them in the bands section.

manthey:
I see a lot of extraneous parameters in the tile style queries (layerName, custom, defaultMin, etc).

Also, it doesn't seem to switch to the composited image until I fiddle with a setting -- I was expecting the previous behavior where all of the channels are composited without any specified min/max values until some value has been specified.  This means that the range we display should probably be based purely on the dtype until we fiddle with values.

I don't seem to be able to type fractional percentages in the scale range (I type 0.2 and immediately it reverts to 0).

manthey:
Eventually (in another PR), it'd be nice to be able to specify a default for any channel that has blank min/max values.  That is, by default all channels would start with blank min/max values.  If the overall defaults are set to 0.2% / 0.2% or 100 / 20000, then any channel with blank values would use those values, but channels with specified values would use those specific values.  Clearing the numerical values in a channel would go back to using the default, and clearing the default would mean min / max are unspecified in the style.

girder/girder_large_image/web_client/vue/components/CompositeLayers.vue
annehaley: Changed in [b44d581](https://github.com/girder/large_image/pull/1175/commits/b44d581d3a457415036a8f2d46bbcb8d8e2dfbe0)

manthey:
We need to change the color of the grab bars on the histograms -- red and black have no contrast in my vision, which makes the grab bars look like bars on the histogram to me.

The icons for showing the histograms are clunky looking and arranged oddly in my view:
image

manthey:
Can we make the default auto-range be 0.2% (not 0.02%) and prevent the user from dragging it past 50%?

girder/girder_large_image/web_client/vue/components/CompositeLayers.vue
```diff
@@ -216,15 +321,116 @@ export default {
...
+    position: absolute;
+    left: 0px;
+    width: calc(100% - 10px);
+    margin: 30px 0px 0px;
+    height: 40px;
+}
+</style>
+
+<style>
+input {
```
manthey, line 422: These need to be scoped or made more exact with some class; otherwise this affects input fields everywhere.  We probably want to add some class specific to this component.

manthey: Maye just make the rule for `.composite-layer-table input`?

annehaley:
I removed the `vue-switches` dependency in [d5b27c4](https://github.com/girder/large_image/pull/1175/commits/d5b27c419ffaedc68ebeb0b575e32df175607e60) and now the tests are passing :+1:

girder/girder_large_image/web_client/vue/components/CompositeLayers.vue
```diff
@@ -216,15 +321,116 @@ export default {
...
+.slider.checked:before {
+  -webkit-transform: translateX(22px);
+  -ms-transform: translateX(22px);
+  transform: translateX(22px);
 }
 .table-container {
-    overflow-x: auto;
-    overflow-y: hidden;
+    overflow: auto;
+    max-height: 300px;
```
manthey, line 396: This makes it surprising to edit the colors, as the color dialog gets hidden near the bottom.  Also, if you expand a row such that it causes the scroll bar to appear, then something about the histogram gets cropped and I can't grab the right control handle:
image
The red band was visible before expanding all, and here the right handle isn't grabbable  any more.

girder/girder_large_image/web_client/vue/components/FrameSelector.vue
```diff
@@ -83,7 +84,7 @@ export default Vue.extend({
                 this.imageMetadata.bands = Object.values(this.imageMetadata.bands).map(
                     (b, i) => {
                         if (!b.interpretation) {
-                            return `Band ${i}`
+                            return `Band ${i + 1}`
```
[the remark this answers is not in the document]
manthey: This change could be in a separate PR.

manthey:
When I clear the entry in the min or max edit fields, I see a console error: `Error: <text> attribute x: Expected length, "NaN".`

manthey:
When I try to show all the histograms of a many channel image, it is asking for the wrong frame for the histogram.  Rather than asking for frames 0,1,2,3,4,... it is asking for 0,1,3,6,10,...  My guess is that frame and framedelta are both set.

girder/girder_large_image/web_client/vue/components/CompositeLayers.vue
annehaley: For the color pickers, I made them relative to the table container in [0799c52](https://github.com/girder/large_image/pull/1175/commits/0799c52cf27bf2a1595217fd5be5107b5b906720). For the scrollbar/canvas resize conflict, my solution was to make the scrollbar always visible: [35601fc](https://github.com/girder/large_image/pull/1175/commits/35601fc8de52546db289484ca68a8588c229ff63)

girder/girder_large_image/web_client/vue/components/FrameSelector.vue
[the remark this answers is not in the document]
annehaley: Should I take out the `+1`s here?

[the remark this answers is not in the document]
manthey: No.  It doesn't have to be in a different PR.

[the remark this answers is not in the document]
manthey: The main difference with https://github.com/girder/large_image/pull/1214 is we now know for all sources how many bands there are and can name them when there are <= 4 bands.

girder/girder_large_image/web_client/vue/components/CompositeLayers.vue
annehaley: Changed in [f2f544a](https://github.com/girder/large_image/pull/1175/commits/f2f544a9fff4295393dffb0cc1dd74da01da0808)

annehaley:
> When I clear the entry in the min or max edit fields, I see a console error: Error: <text> attribute x: Expected length, "NaN".

Fixed in [4a46617](https://github.com/girder/large_image/pull/1175/commits/4a466178197dcef9454ebf404db70e21bd0e8412)

> When I try to show all the histograms of a many channel image, it is asking for the wrong frame for the histogram. Rather than asking for frames 0,1,2,3,4,... it is asking for 0,1,3,6,10,... My guess is that frame and framedelta are both set.

Fixed in [448ef26](https://github.com/girder/large_image/pull/1175/commits/448ef26b36a59a505ec8006faea88f74e1dfabbe)

manthey:
When you have scrolled down the list of channels, the color picker is out of view (scrolling up shows it).

annehaley:
> When you have scrolled down the list of channels, the color picker is out of view (scrolling up shows it).

Changed in [3e2fd3a](https://github.com/girder/large_image/pull/1175/commits/3e2fd3a80c275d427dbeff123c96f6a0d2c8e740). Rather than having a color picker component for every row in the table with each being relative to the color picker column, we now have one color picker component relative to the table header, which is sticky during scrolling.

girder/girder_large_image/web_client/vue/components/FrameSelector.vue
```diff
@@ -33,7 +34,7 @@ export default Vue.extend({
     },
     methods: {
         updateStyle(style) {
-            this.style = style
+            this.style = Object.assign(this.style, style)
```
manthey, line 37: What does doing `this.style = ` in front of the `Object.assign` do?

manthey (approved):
Great.

PR #1181, opened 2023-05-29
Ship Docker images with Python 3.10

banesullivan:


banesullivan:
Actually this is unimportant right now

PR #1182, opened 2023-05-29
Remove rasterio dep on pyproj

banesullivan:
Draft work to remove the `rasterio` source module's dependence on `pyproj`.

Considering how easy it is to install `pyproj` these days, I don't think this is totally necessary but it's probably best to have the rasterio source module use `rasterio` directly for everything rather than pulling an additional dependency.

There is still one remaining function `getPixelSizeInMeters()` that I wasn't immediately sure on how to reproduce with `rasterio`

### DIFF
--- a/large_image/tilesource/geo.py
+++ b/large_image/tilesource/geo.py
@@ -233,7 +233,7 @@ class GDALBaseFileTileSource(GeoBaseFileTileSource):
         if not has_pyproj:
             # Estimate based on great-cirlce distance
             def great_circle(lon1, lat1, lon2, lat2):
-                from math import acos, asin, cos, radians, sin, sqrt
+                from math import acos, cos, radians, sin
                 lon1, lat1, lon2, lat2 = map(radians, [lon1, lat1, lon2, lat2])
                 return None, None, 6.378e+6 * (
                     acos(sin(lat1) * sin(lat2) + cos(lat1) * cos(lat2) * cos(lon1 - lon2))
--- a/large_image/tilesource/geo.py
+++ b/large_image/tilesource/geo.py
@@ -1,13 +1,17 @@
 from urllib.parse import urlencode, urlparse
 
-import pyproj  # TODO: import issues
-
 from large_image.cache_util import CacheProperties, methodcache
 from large_image.constants import SourcePriority, TileInputUnits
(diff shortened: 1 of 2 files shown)


manthey:
We could make a dependency on `geographiclib`, which is an MIT-licensed pure python library.  Then
```
geod = geographiclib.geodesic.Geodesic.WGS84
s1 = geod.Inverse((bounds['ul']['x'], bounds['ul']['y'], bounds['ur']['x'], bounds['ur']['y'])['s12']
...
```

banesullivan:
I think geographiclib is C++ with Python wrappings? Either way, I think the goal with this PR should be to reduce dependencies for this operation

banesullivan:
It's worth trying this to at least see how different the results can be for our testing data. @manthey, is that math something your familiar with or should I do some research into how to do this?

banesullivan:
@12rambau, do you have any thoughts on depending on `pyproj` with the `rasterio` source module?

manthey:
> It's worth trying this to at least see how different the results can be for our testing data. @manthey, is that math something you are familiar with or should I do some research into how to do this?

We put the Vincenty functions directly into geojs, though we probably should have just pulled in a package.  They aren't hard, but fussy and not the sort of thing I think we should have since so many other packages implement that.  The simple great circle distance on a sphere is just trig (see https://en.wikipedia.org/wiki/Great-circle_distance) to get an angle multiplied by the average radius of the Earth; if we are doing a fallback to avoid a package, that is probably all we want.

banesullivan:
Sounds good and simple enough! Will add to my todo list to do the great circle estimation then see if it skews results too differently

banesullivan:
That was easy! Tests pass for me locally and as far as I can tell (few surface level tests), there isn't any visible impact on tiles or thumbnails

manthey:
The biggest difference I see is in sample file test/test_files/rgba_geotiff.tiff, which reports a pixel size of 8.181m rather than 8.170 m.

manthey (approved):

12rambau:
oups I'm a bit late to the party. I'm usually depending on the CRS module of rasterio which is sufficient for my applications. It seems a bit more complex from your side but the solution you found is elegant enough.

PR #1201, opened 2023-06-09
Allow rounding histogram bin edges and reducing bin counts

manthey:
For integer data types, using full bin counts can result in jagged looking results because some bins will have fewer or no discrete values than others.


manthey:
@annehaley Once this PR is merged, for the frame histogram control, we'll want to pass a `style` of `{}` (see #1200) and a `roundRange` of `true` to the histogram endpoint.

PR #1227, opened 2023-06-28
Fix updating the hidden style field for the frame selector

### LINKED ISSUE #1225: For frame compositing, defer initial histogram query
With the recent frame controls, we now always make a histogram query.  For images that don't have multiple frames, it will be uncommon to want this data and it would be better to not make an initial histogram query.

manthey:
This ensure it maintains reactivity and updates for just the current mode, including clearing for non-compositing modes.

Resolves #1225.

Sets default band names more appropriately when there are other than 3 bands.


manthey:
@annehaley If you have time, please look at this.  I may merge it relatively soon, as the hidden style field is inconsistent with the selector state without this.

PR #1233, opened 2023-07-01
Frame selection keyboard shortcuts

### LINKED ISSUE #1134: Frame control hot keys
It would nice to have hotkeys associated with the frame controls so that a user can quickly toggle off and on channels or bands.  Since we already use number keys by themselves for some map actions, perhaps this would be shift+number.  This doesn't answer what to do when there are more than 10 channels.  A different possibility would be to use something like mousetrap and use key sequences.  Suggestions welcome.

annehaley:
Resolves #1134


manthey:
In Chrome, `ctrl-<number>` switches tabs (I had never used that before).  We either need to suppress that or do something different.

manthey (approved):

PR #1237, opened 2023-07-10
Create, select, and delete image view presets

annehaley:
This PR adds a widget for Image View Presets on the righthand side of the image controls. 
The user doe the following: 
1. select from existing presets to apply one
2. use the plus button to name and save the current state as a new preset
3. delete the selected preset 

Requesting @manthey to take a look, since this will probably need a few iterations.

image

### DIFF
--- a/girder/girder_large_image/web_client/vue/components/PresetsMenu.vue
+++ b/girder/girder_large_image/web_client/vue/components/PresetsMenu.vue
@@ -97,7 +97,7 @@ export default {
         Image View Presets:
         <select v-model="selectedPreset">
             <option v-if="availablePresets.length === 0" :value="undefined" selected disabled>
-                No presets available.
+                No presets available
             </option>
             <option v-else :value="undefined" selected disabled>
                 Select a preset
--- a/girder/girder_large_image/web_client/vue/components/PresetsMenu.vue
+++ b/girder/girder_large_image/web_client/vue/components/PresetsMenu.vue
@@ -1,0 +1,133 @@
+<script>
+import { restRequest } from '@girder/core/rest';
+
+export default {
+    props: ['itemId', 'currentMode', 'currentFrame', 'currentStyle'],
+    emits: ['setCurrentMode', 'setCurrentFrame', 'updateStyle'],
+    data() {
+        return {
+            availablePresets: [],
+            selectedPreset: undefined,
+            showPresetCreation: false,
+            newPresetName: undefined,
+        }
+    },
+    methods: {
+        getPresets() {
+            restRequest({
+                type: 'GET',
+                url: 'item/' + this.itemId + '/internal_metadata/presets',
+            }).then((presets) => {
+                if (presets) {
+                    this.availablePresets = presets
+                }
+            })
+        },
+        addPreset() {
+            const newPreset = {
+                'name': this.newPresetName || this.generatedPresetName,
+                'mode': this.currentMode,
+                'frame': this.currentFrame,
+                'style': this.currentStyle,
+            }
+            this.availablePresets.push(newPreset)
+            this.selectedPreset = newPreset.name
+            this.savePresetsList()
+        },
+        deleteSelectedPreset() {
+            this.availablePresets = this.availablePresets.filter((p) => p.name !== this.selectedPreset)
+            this.selectedPreset = undefined
+            this.savePresetsList()
+        },
+        savePresetsList() {
+            restRequest({
+                type: 'PUT',
+                url: 'item/' + this.itemId + '/internal_metadata/presets',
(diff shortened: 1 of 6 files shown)


manthey:
Is there any way to close the "new preset" after it has been expanded?  Maybe clicking the plus again would hide it.  It feel like once you start creating a present you are stuck doing so.

girder/girder_large_image/web_client/vue/components/PresetsMenu.vue
```diff
@@ -0,0 +1,148 @@
...
+    },
+}
+</script>
+
+<template>
+    <div class="presets-menu">
+        Image View Presets:
+        <select v-model="selectedPreset">
+            <option v-if="availablePresets.length === 0" :value="undefined" selected disabled>
+                No presets available.
```
manthey, line 100: For appearance consistency, I think we should remove the period at the end of `No presets available.`

annehaley: Ok, changed in [eee3513](https://github.com/girder/large_image/pull/1237/commits/eee351327fa1cff7c07b01107958740a95bd9cc9)

annehaley:
> Is there any way to close the "new preset" after it has been expanded? Maybe clicking the plus again would hide it. It feel like once you start creating a present you are stuck doing so.

Yes, I can do that. Changed in [312e6ed](https://github.com/girder/large_image/pull/1237/commits/312e6ed43aa861f7d7e698531a7b96b697d332c3)

manthey:
This is an opinion question (not necessarily a problem).  I created a preset with a selection.  I then decided it wasn't as good as it could be, so I twiddled some settings.  To replace an existing preset without recreating the settings manually, I then have to (1) create a new preset with a name I don't want, (2) switch to the old version, (3) delete it, (4) switch to the undesired name, (5) create a new preset with the original name, (6) switch back to the undesired name, (7) delete it.

Obviously, replacing an existing preset should have some warning, but refusing to replace it makes updating a present hard.  What if instead when there was a name collision the "Save preset" button changed to "Update preset"?

girder/girder_large_image/web_client/vue/components/PresetsMenu.vue
```diff
@@ -0,0 +1,148 @@
...
+            })
+        },
+        addPreset() {
+            const newPreset = {
+                'name': this.newPresetName || this.generatedPresetName,
+                'mode': this.currentMode,
+                'frame': this.currentFrame,
+                'style': this.currentStyle,
+            }
+            if (this.availablePresets.find((p) => p.name === newPreset.name)) {
```
manthey, line 34: We should trim the name before comparison and saving it.  I made a preset called "foo" and it let me create "foo ", which looks like a name collision.

manthey:
1. I create a preset with some channels selected and one with all channels selected.  When I switch between them, it visually looks right, but the check boxes don't clear when channels are deselected.  
2. When a preset has all channels selected, the "all" check mark isn't checked.
3. The auto range now looks like 0.002% rather than 0.2%.

manthey:
Would it be possible to compare the current frame and style values and switch the preset drop down to "Select a preset" when things are changed and switch to a specific preset when they match?

manthey:
You can't switch back to a preset with spaces at the end of its name.

manthey:
On a multi-axis preset, I set each axis to a different value and it then render it as if the first axis was the overall frame number and the other axes were 0.

annehaley:
> Obviously, replacing an existing preset should have some warning, but refusing to replace it makes updating a present hard. What if instead when there was a name collision the "Save preset" button changed to "Update preset"?

Good idea. Changed in [4ced213](https://github.com/girder/large_image/pull/1237/commits/4ced213d26745444af6afdac2a81075de94c649f).

annehaley:
> Would it be possible to compare the current frame and style values and switch the preset drop down to "Select a preset" when things are changed and switch to a specific preset when they match?

Added in [7f37a10](https://github.com/girder/large_image/pull/1237/commits/7f37a10a5138814ad1ae47319ef2fab464152f58)

annehaley:
> On a multi-axis preset, I set each axis to a different value and it then render it as if the first axis was the overall frame number and the other axes were 0.

Can you give a more specific example? When I try this on `multi_all.yml`, it works as expected. 
image

manthey:
> > On a multi-axis preset, I set each axis to a different value and it then render it as if the first axis was the overall frame number and the other axes were 0.
> 
> Can you give a more specific example? When I try this on `multi_all.yml`, it works as expected. image

On DDX58_AXL_EGFR_well2.nd2, I set the following:
image
Then save the preset and I see 
image

manthey (approved):

PR #1248, opened 2023-07-26
Presets from Large Image Configuration file

annehaley:
This PR allows a user to specify image view presets in their `.large_image_config.yaml` configuration file at the folder level. Presets specified in the configuration file will be added to the presets available for every image in that folder (as long as the preset is applicable to that image). 

We can write more complicated logic for determining whether a preset is applicable to an image, but for now, the `presetApplicable` function simply evaluates whether the preset's `mode` and `frame` values are available for the image.

This PR also adds a section to the documentation explaining how to add presets to the configuration file. This section includes some basic example presets.

### DIFF
--- a/docs/girder_config_options.rst
+++ b/docs/girder_config_options.rst
@@ -19,7 +19,7 @@ The yaml file has the following structure:
 
     ---
     # most settings are key-value pairs, where the value could be another
-    # directionary with keys and values, lists, or other valid data.
+    # dictionary with keys and values, lists, or other valid data.
     <key>: <value>
     # The access key is special
     access:
@@ -191,6 +191,139 @@ By default, item metadata can contain any keys and values.  These can be given b
         # Exclusive values can be specified instead
         # exclusiveMinimum: 0
         # exclusiveMaximum: 10
+
+Image View Presets
+...........
+
+This is used to specify a list of default presets for viewing images in the folder.
+Presets can be customized and saved in the GeoJS Image Viewer.
+To retrieve saved presets, use http://[serverURL]/api/v1/item/[itemID]/internal_metadata/presets.
+You can convert the response to YAML and paste it into the `imagePresets` key in your config file.
+
+Each preset can specify a name, a view mode, an image frame, and style options.
+- The name of a preset can be any string which uniquely identifies the preset.
+- There are four options for mode:
+  1. Frame control
+    id: 0
+    name: Frame
+  2. Axis control
+    id: 1
+    name: Axis
+  3. Channel Compositing
+    id: 2
+    name: Channel Compositing
+  4. Band Compositing
+    id: 3
+    name: Band Compositing
+- The frame of a preset is a 0-based index representing a single frame in a multiframe image.
+  For single-frame images, this value will always be 0.
+  For channel compositing, each channel will have a `framedelta` value which represents distance from this base frame value.
+  The result of channel compositing is multiple frames (calculated via framedelta) composited together.
+- The style of a preset is a dictionary with a schema similar to the [style schema for tile retrieval](tilesource_options.rst#style).
+  The value for a preset's style consists of a band definition, where each band may have the following:
+  1. `band`: A 1-based index of a band within the current frame
+  2. `framedelta`: An integer representing distance from the current frame, used for compositing multiple frames together
+  3. `palette`: A hexidecimal string beginning with "#" representing a color to stain this frame
+  4. `min`: The value to map to the first palette value
+  5. `max`: The value to map to the last palette value
+  6. `autoRange`: A shortcut for excluding a percentage from each end of the value distribution in the image. Express as a float.
+
+The YAML below includes some example presets.
+
+::
+
+    ---
+    # If present, each preset in this list will be added to the preset list
+    # of every image in the folder for which the preset is applicable
+    imagePresets:
+    - name: Frame control - Frame 4
+      frame: 4
+      mode:
+        id: 0
+        name: Frame
+    - name: Axis control - Frame 25
+      frame: 25
+      mode:
+        id: 1
+        name: Axis
+    - name: 3 channels
+      frame: 0
+      mode:
+        id: 2
+        name: Channel Compositing
+      style:
+        bands:
+        - framedelta: 0
+          palette: "#0000FF"
+        - framedelta: 1
+          palette: "#FF0000"
+        - framedelta: 2
+          palette: "#00FF00"
+    - name: 3 bands
+      frame: 0
+      mode:
+        id: 3
+        name: Band Compositing
+      style:
+        bands:
+        - band: 1
+          palette: "#0000FF"
+        - band: 2
+          palette: "#FF0000"
+        - band: 3
+          palette: "#00FF00"
+    - name: Channels with Min and Max
+      frame: 0
+      mode:
+        id: 2
+        name: Channel Compositing
+      style:
+        bands:
+        - min: 18000
+          max: 43000
+          framedelta: 0
+          palette: "#0000FF"
+        - min: 18000
+          max: 43000
+          framedelta: 1
+          palette: "#FF0000"
+        - min: 18000
+          max: 43000
+          framedelta: 2
+          palette: "#00FF00"
+        - min: 18000
+          max: 43000
+          framedelta: 3
+          palette: "#FFFF00"
+    - name: Auto Ranged Channels
+      frame: 0
+      mode:
+        id: 2
+        name: Channel Compositing
+      style:
+        bands:
+        - autoRange: 0.2
+          framedelta: 0
+          palette: "#0000FF"
+        - autoRange: 0.2
+          framedelta: 1
+          palette: "#FF0000"
+        - autoRange: 0.2
+          framedelta: 2
+          palette: "#00FF00"
+        - autoRange: 0.2
+          framedelta: 3
+          palette: "#FFFF00"
+        - autoRange: 0.2
+          framedelta: 4
+          palette: "#FF00FF"
+        - autoRange: 0.2
+          framedelta: 5
+          palette: "#00FFFF"
+        - autoRange: 0.2
+          framedelta: 6
+          palette: "#FF8000"
+
 
 Editing Configuration Files
 ---------------------------
--- a/docs/girder_config_options.rst
+++ b/docs/girder_config_options.rst
@@ -193,7 +193,7 @@ By default, item metadata can contain any keys and values.  These can be given b
         # exclusiveMaximum: 10
 
 Image View Presets
-...........
+..................
 
 This is used to specify a list of default presets for viewing images in the folder.
 Presets can be customized and saved in the GeoJS Image Viewer.
--- a/docs/girder_config_options.rst
+++ b/docs/girder_config_options.rst
@@ -201,24 +201,28 @@ To retrieve saved presets, use http://[serverURL]/api/v1/item/[itemID]/internal_
 You can convert the response to YAML and paste it into the `imagePresets` key in your config file.
 
 Each preset can specify a name, a view mode, an image frame, and style options.
+
 - The name of a preset can be any string which uniquely identifies the preset.
+
 - There are four options for mode:
-  1. Frame control
-    id: 0
-    name: Frame
-  2. Axis control
-    id: 1
-    name: Axis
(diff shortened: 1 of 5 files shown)


docs/girder_config_options.rst
```diff
@@ -192,6 +192,143 @@ By default, item metadata can contain any keys and values.  These can be given b
...
+This is used to specify a list of default presets for viewing images in the folder.
+Presets can be customized and saved in the GeoJS Image Viewer.
+To retrieve saved presets, use http://[serverURL]/api/v1/item/[itemID]/internal_metadata/presets.
+You can convert the response to YAML and paste it into the `imagePresets` key in your config file.
+
+Each preset can specify a name, a view mode, an image frame, and style options.
+
+- The name of a preset can be any string which uniquely identifies the preset.
+
+- There are four options for mode:
```
manthey, line 207: When I preview this on GitHub, it renders differently than I think was intended.

docs/girder_config_options.rst
```diff
@@ -192,6 +192,143 @@ By default, item metadata can contain any keys and values.  These can be given b
...
+        - band: 3
+          palette: "#00FF00"
+    - name: Channels with Min and Max
+      frame: 0
+      mode:
+        id: 2
+        name: Channel Compositing
+      style:
+        bands:
+        - min: 18000
```
manthey, line 286: Using this style directly causes an issue - `currentLayerStyle.min.includes is not a function"`, because we expect min/max to be a string in the javascript and it is a number here.  We should cast it in the javascript to handle that.

docs/girder_config_options.rst
```diff
@@ -192,6 +192,143 @@ By default, item metadata can contain any keys and values.  These can be given b
...
+  6. `autoRange`: A shortcut for excluding a percentage from each end of the value distribution in the image. Express as a float.
+
+The YAML below includes some example presets.
+
+::
+
+    ---
+    # If present, each preset in this list will be added to the preset list
+    # of every image in the folder for which the preset is applicable
+    imagePresets:
```
manthey, line 242: I'm thinking the key should be `imageFramePresets` or `framePresets`, and the title of this section of the docs would be `Image Frame Presets` instead of `Image View Presets`.

manthey:
This ends up adding all of the presets from the config file to the item's internal values when I create a different preset.  We probably want to _not_ add any preset to an item that matches a config preset exactly.  Otherwise, for example, I change the name of the config preset and now I have two presets rather than one.

docs/girder_config_options.rst
annehaley: I added a casting step in [544945a](https://github.com/girder/large_image/pull/1248/commits/544945a9e690aa83d791f48d7ebf73dd98768fed)

annehaley:
> This ends up adding all of the presets from the config file to the item's internal values when I create a different preset. We probably want to _not_ add any preset to an item that matches a config preset exactly. Otherwise, for example, I change the name of the config preset and now I have two presets rather than one.

You're right, I changed this so that I'm keeping the item presets separate from the folder presets rather than grouping them together in `availablePresets`. When we save the preset list, [544945a](https://github.com/girder/large_image/pull/1248/commits/544945a9e690aa83d791f48d7ebf73dd98768fed) only sends `itemPresets` now.

manthey:
I'm getting an error: `Error in mounted hook: "TypeError: styleArray.find is not a function"` when I try to select a channel compositing preset.

docs/girder_config_options.rst
annehaley: I can't find where you can look at the docs preview on Github. I've been using the VSCode preview. What looks off?

manthey: You can just view the file directly, or navigate to the branch and the file:  (https://github.com/girder/large_image/blob/presets-from-config/docs/girder_config_options.rst)

manthey: The section "there are four options for mode" isn't rendering in the structure I think we want.

manthey: The sphinx-generated preview can also be seen via the circle-ci job artifacts.  Click the check or x next to the ci/circleci: list_and_docs job, choose the Artifacts tab, then build/docs.index.html.

girder/girder_large_image/web_client/vue/components/PresetsMenu.vue
```diff
@@ -123,6 +146,16 @@ export default {
                     this.$emit('setCurrentFrame', preset.frame)
                 }
                 if (preset.style && Object.keys(preset.style.bands).length) {
+                    preset.style.bands = Object.fromEntries(
```
manthey, line 149: This converts bands into a dictionary, but it needs to be an array.

docs/girder_config_options.rst
annehaley: Even though the preview for this works well, the lint still fails it:
```
Warning, treated as error:
/root/project/docs/girder_config_options.rst:210:Unexpected indentation.
```

girder/girder_large_image/web_client/vue/components/PresetsMenu.vue
annehaley: Is [f26394c](https://github.com/girder/large_image/pull/1248/commits/f26394c28c502314d0b7f54b29d06408d260a4a1) a sufficient fix?

manthey:
When I switch to a preset that is just a frame or just a channel, I get an error:
```
Error in callback for watcher "currentFrame": "TypeError: Cannot read properties of undefined (reading 'bands')"
...
vue.js:1906 TypeError: Cannot read properties of undefined (reading 'bands')
    at VueComponent.checkPresetMatch (PresetsMenu.vue:95:1)
```
It seems like in these instances, `checkPresetMatch` is being called while `this.currentStyle` is undefined.

girder/girder_large_image/web_client/vue/components/PresetsMenu.vue
manthey: Yes, it appears to be.

annehaley:
> It seems like in these instances, `checkPresetMatch` is being called while `this.currentStyle` is undefined.

Ok, thanks for catching that. I just added an extra `if` check to that function in [51345c0](https://github.com/girder/large_image/pull/1248/commits/51345c061e48ebf508f516b8903d4fc245a41929)

manthey:
> > I noticed an issue not related to this branch, but related to presets. I have two presets which only different based on if they have auto-range set on the active channels. Switch to the auto-on preset works fine, switching to the auto-off preset doesn't turn off the auto buttons.
> 
> I added an `else` statement in [57514a4](https://github.com/girder/large_image/pull/1248/commits/57514a495d961bd9ff692448aebba9f2c48ef336) that should fix this. Could you try with this change?

I'm not seeing any difference.

I'm using WD-76845-097.ome.tif and have the presets listed below.

I also don't see it change properly from preset "CD4RO" and "CD4RO Auto"
```
[
    {
        "frame": 15,
        "mode": {
            "id": 1,
            "name": "Axis"
        },
        "name": "Aortic"
    },
    {
        "frame": 0,
        "mode": {
            "id": 2,
            "name": "Channel Compositing"
        },
        "name": "Auto Ranged Channels",
        "style": {
            "bands": [
                {
                    "autoRange": 0.2,
                    "framedelta": 0,
                    "palette": "#FF0000"
                },
                {
                    "autoRange": 0.2,
                    "framedelta": 1,
                    "palette": "#00FF00"
                },
                {
                    "autoRange": 0.2,
                    "framedelta": 2,
                    "palette": "#0000FF"
                },
                {
                    "autoRange": 0.2,
                    "framedelta": 4,
                    "palette": "#FF00FF"
                }
            ]
        }
    },
    {
        "frame": 0,
        "mode": {
            "id": 2,
            "name": "Channel Compositing"
        },
        "name": "DNA auto",
        "style": {
            "bands": [
                {
                    "autoRange": 0.2,
                    "framedelta": 0,
                    "palette": "#FF0000"
                },
                {
                    "autoRange": 0.2,
                    "framedelta": 4,
                    "palette": "#FF00FF"
                },
                {
                    "autoRange": 0.2,
                    "framedelta": 8,
                    "palette": "#00FF80"
                },
                {
                    "autoRange": 0.2,
                    "framedelta": 12,
                    "palette": "#FF8080"
                },
                {
                    "autoRange": 0.2,
                    "framedelta": 16,
                    "palette": "#80FFFF"
                },
                {
                    "autoRange": 0.2,
                    "framedelta": 20,
                    "palette": "#00FF40"
                },
                {
                    "autoRange": 0.2,
                    "framedelta": 24,
                    "palette": "#FF4040"
                },
                {
                    "autoRange": 0.2,
                    "framedelta": 28,
                    "palette": "#40FFFF"
                },
                {
                    "autoRange": 0.2,
                    "framedelta": 32,
                    "palette": "#00FFC0"
                }
            ]
        }
    },
    {
        "frame": 0,
        "mode": {
            "id": 2,
            "name": "Channel Compositing"
        },
        "name": "DNA",
        "style": {
            "bands": [
                {
                    "framedelta": 0,
                    "palette": "#FF0000"
                },
                {
                    "framedelta": 4,
                    "palette": "#FF00FF"
                },
                {
                    "framedelta": 8,
                    "palette": "#00FF80"
                },
                {
                    "framedelta": 12,
                    "palette": "#FF8080"
                },
                {
                    "framedelta": 16,
                    "palette": "#80FFFF"
                },
                {
                    "framedelta": 20,
                    "palette": "#00FF40"
                },
                {
                    "framedelta": 24,
                    "palette": "#FF4040"
                },
                {
                    "framedelta": 28,
                    "palette": "#40FFFF"
                },
                {
                    "framedelta": 32,
                    "palette": "#00FFC0"
                }
            ]
        }
    },
    {
        "frame": 11,
        "mode": {
            "id": 1,
            "name": "Axis"
        },
        "name": "CD4RO"
    },
    {
        "frame": 0,
        "mode": {
            "id": 2,
            "name": "Channel Compositing"
        },
        "name": "CD45RO Auto",
        "style": {
            "bands": [
                {
                    "autoRange": 0.2,
                    "framedelta": 11,
                    "palette": "#FFFFFF"
                }
            ]
        }
    }
]
```

annehaley:
Oh I see, there are two places where I needed an `else`. Amended in [75347c1](https://github.com/girder/large_image/pull/1248/commits/75347c1af4643b5a7d8eb6c46dd771da5e435bc3). Your example case works for me now.

manthey:
> Oh I see, there are two places where I needed an `else`. Amended in [75347c1](https://github.com/girder/large_image/pull/1248/commits/75347c1af4643b5a7d8eb6c46dd771da5e435bc3). Your example case works for me now.

Now the auto range doesn't seem to be turning on for me.  The autorange controls then don't change when clicked (like they are known to be on) and clicking the global autorange suddenly makes the style update.

docs/girder_config_options.rst
annehaley: You're right, I had just needed to refresh my page. Thanks!

manthey:
I still see some issues: (a) after going to a present, if I click the auto-range all button, all the channels turn on (they don't show that they are turned on, but the image changes).  (b) sometimes when switching between two presets where one has auto-range on (e.g., my samples with CD4RO and the auto-ranged version), the auto-ranged version isn't shown (the controls looks right, but the image isn't). (c) If a folder-level preset has the same name as a local preset, they are both shown, but you can't select the folder-level preset -- I don't think it should be shown.  (d) ideally, we shouldn't include parameters in the style used for fetching images if they aren't used, since that reduces caching (e.g., preset=true, autoRange).

manthey:
I still see: On the CD4RO / CD4RO auto, I select the first (not auto) and it shows properly.  I select the auto and it briefly shows the auto and then shows the non-auto.

manthey (approved):
This is working in all of my test cases.  Though I do see an occasional error in the console (probably unrelated to this specific PR): `HistogramEditor.vue:151 Error: <rect> attribute width: A negative value is not valid. ("-9")`

annehaley:
> This is working in all of my test cases. Though I do see an occasional error in the console (probably unrelated to this specific PR): `HistogramEditor.vue:151 Error: <rect> attribute width: A negative value is not valid. ("-9")`

Ok, thank you for testing. I'm not able to replicate this, though. Perhaps you could submit an issue with how to replicate and I can address it in a separate PR?

PR #1256, opened 2023-08-09
Fix frame selection bugs

annehaley:
This PR fixes several bugs within the frame selection widget; each bug is described in the commit message that provides the respective fix.


manthey:
We don't want to commit the package-lock.json file because of the way girder builds things.

girder/girder_large_image/web_client/vue/components/HistogramEditor.vue
```diff
@@ -1,7 +1,8 @@
 <script>
-import { restRequest } from '@girder/core/rest';
-import { nextTick } from 'vue';
-import { makeDraggableSVG } from '../utils/drag';
+import {restRequest} from '@girder/core/rest';
+import {nextTick} from 'vue'; // eslint-disable-line
```
manthey, line 3: We could also do `import Vue from 'vue';` and then use `Vue.nextTick` instead of `nextTick`, if you think that is better than the eslint disable.

manthey (approved):

PR #1281, opened 2023-08-29
Add a Jupyter notebook example

manthey:


manthey:
See https://github.com/girder/large_image/blob/notebook-example/docs/large_image_examples.ipynb

PR #1300, opened 2023-09-14
Band Behavior

### LINKED ISSUE #1287: Band compositing isn't working
The band compositing isn't working.  Fundamentally, it looks like we expect layers to be a list, but for bands it is a dictionary.

### LINKED ISSUE #1295: Show band compositing even on single frame / single band images
I have a sample single band / single frame image that is 16-bit and it would be handy to use the histogram scaling to better see the values in the image. See r11c13-1252942648.tiff.gz from the omero samples, for instance.

annehaley:
[746b0e8](https://github.com/girder/large_image/pull/1300/commits/746b0e8cf35266200a4c145b2ec41d968e0c021a) resolves #1287.

 [a649a88](https://github.com/girder/large_image/pull/1300/commits/a649a88d5b2b0121fa3a49237dc3c01f684fb574) resolves #1295 .

### DIFF
--- a/girder/girder_large_image/web_client/vue/components/FrameSelector.vue
+++ b/girder/girder_large_image/web_client/vue/components/FrameSelector.vue
@@ -173,7 +173,7 @@ export default Vue.extend({
                     {id: 2, name: 'Channel Compositing'}
                 );
             }
-            if (this.metadata.bandCount > 1) {
+            if (this.metadata.bandCount) {
                 this.sliderModes.push(
                     {id: 3, name: 'Band Compositing'}
                 );
--- a/girder/girder_large_image/web_client/vue/components/FrameSelector.vue
+++ b/girder/girder_large_image/web_client/vue/components/FrameSelector.vue
@@ -250,24 +250,24 @@ export default Vue.extend({
     <!-- Use styling instead of v-if to make each invisible so that the components are not unmounted -->
     <div class="image-frame-simple-control">
       <composite-layers
-        v-if="imageMetadata.channels && modesShown[2]"
+        v-if="metadata.channels && modesShown[2]"
         key="channels"
         :item-id="itemId"
         :current-frame="currentFrame"
         :current-style="style[2]"
-        :layers="imageMetadata.channels"
-        :layer-map="imageMetadata.channelmap"
+        :layers="metadata.channels"
+        :layer-map="metadata.channelmap"
         :active="currentModeId === 2"
         :class="currentModeId === 2 ? '' : 'invisible'"
         @updateStyle="(style) => updateStyle(2, style)"
       />
       <composite-layers
-        v-if="imageMetadata.bands && modesShown[3]"
+        v-if="metadata.bands && modesShown[3]"
         key="bands"
         :item-id="itemId"
         :current-frame="currentFrame"
         :current-style="style[3]"
-        :layers="imageMetadata.bands"
+        :layers="metadata.bands"
         :layer-map="undefined"
         :active="currentModeId === 3"
         :class="currentModeId === 3 ? '' : 'invisible'"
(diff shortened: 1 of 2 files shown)


girder/girder_large_image/web_client/vue/components/CompositeLayers.vue
```diff
@@ -108,36 +108,39 @@ export default {
                     enabled: true,
                     min: undefined,
                     max: undefined,
-                    autoRange: undefined
+                    autoRange: undefined,
+                    palette: '#ffffff'  // default color is white
```
manthey, line 112: Won't this make ALL bands white?

annehaley: My intention was to make white the default which is overridden in the case that `this.layers.length > 1`. But I didn't notice the `if (!this.compositeLayerInfo[layerName].palette)` statement which would prevent color assignment. [9ab1a65](https://github.com/girder/large_image/pull/1300/commits/9ab1a65daa8a85267af6319fe2e4f86b05b49376) removes that line. Now colors are properly assigned when there is more than one layer.

manthey: What if a layer should have white?  Like a brightfield channel?

manthey: In a quick test, the default colors are no longer being kept (white for brightfield, blue for dapi, etc.)

annehaley: Ok, thanks for catching that. It's been a while since I've looked at the color assignment code, but I updated it in [742af06](https://github.com/girder/large_image/pull/1300/commits/742af068f98a912c8a1598f716ba4592328e600b) to account for the white default and still do color assignment like normal when there is more than one layer. Could you test again?

manthey: It all appears to work in my quick tests.  Would it have been better to have just tried to use the band names for the color lookup, since in a single band is called "gray"?

manthey: In thinking that through, we have some geospatial sources that could expose named bands with some missing (e.g., a three-band image with interpretations of red, 0, 0 (where the other two are near-ir bands).  I'll have to dig around to find such a file.

annehaley: I can do that, too. I would just need to add a new mapping to `colors.js`. In the case you mentioned, though, wouldn't it be better to assign colors based on index rather than name?

manthey: Superficially, it looks like if you reverted to a649a88 and added 
```
    '^red$': '#FF0000',
    '^green$': '#00FF00',
    '^blue$': '#0000FF',
    '^gr[ae]y(|scale)$': '#FFFFFF',
```
to the color table, it would do what we want.

As far as I can tell, we don't get useful band interpretations unless they are red/green/blue/gray/alpha; otherwise it is either "palette" or "other" (also called 0).  If we have the useful band name, then the color table works.

manthey: In the red/near-ir example, they could theoretically be stored in a different order (so you'd get [0, 0, red] as the interpretations), so we would still want to look up the color by interpretation name.

annehaley: Ok, I rewrote the history with that solution instead: [476288d](https://github.com/girder/large_image/pull/1300/commits/476288dfb54d8c6f9f40af3f10d9ecb0a587aaaa) Thanks!

manthey (approved):

Issue #1204, opened 2023-06-12
Read WSI through DICOM web


manthey:
Either via a modification of the existing DICOM source or via a new tile source, we want to read WSI from a DICOM web server.

Related, but not really part of this issue: Since finding WSI from a DICOM web server may involve searching for what is present, it may be useful to also expose a method to do that search.  This means that besides a tile source, this might have some sort of browsing added to Girder.

PR #1303, opened 2023-09-15
Add initial DICOMweb support

psavery:
This creates a new DICOMweb assetstore that is used to track resources on a DICOMweb server. The user provides the DICOMweb server URL and optionally QIDO and WADO prefixes (if the server requires them). The user is then able to import references to the objects on the DICOMweb server.

The files that are created via this assetstore are able to be viewed with the dicom tile source, which has been extended to handle files on a DICOMweb server.

A test was also added that utilizes the public DICOMweb server located [here](https://imagingdatacommons.github.io/slim/).

There is still more work to do in future PRs, including:

* Adding a search/filter capability
* Adding authentication for non-public servers
* Query and store metadata from the DICOMweb objects

Fixes: #1204

### DIFF
--- a/sources/dicom/large_image_source_dicom/__init__.py
+++ b/sources/dicom/large_image_source_dicom/__init__.py
@@ -7,11 +7,14 @@ from importlib.metadata import version as _importlib_version
 
 import numpy as np
 
+from large_image import config
 from large_image.cache_util import LruCacheMetaclass, methodcache
 from large_image.constants import TILE_FORMAT_PIL, SourcePriority
 from large_image.exceptions import TileSourceError, TileSourceFileNotFoundError
 from large_image.tilesource import FileTileSource
 from large_image.tilesource.utilities import _imageToNumpy, _imageToPIL
+
+from .dicom_tags import dicom_key_to_tag
 
 pydicom = None
 wsidicom = None
@@ -113,11 +116,14 @@ class DICOMFileTileSource(FileTileSource, metaclass=LruCacheMetaclass):
         """
         super().__init__(path, **kwargs)
 
+        self.logger = config.getConfig('logger')
+
         # We want to make a list of paths of files in this item, if multiple,
         # or adjacent items in the folder if the item is a single file.  We
         # filter files with names that have a preferred extension.
+        # If the path is a dict, that likely means it is a DICOMweb asset.
         path = self._getLargeImagePath()
-        if not isinstance(path, list):
+        if not isinstance(path, (dict, list)):
             path = str(path)
             if not os.path.isfile(path):
                 raise TileSourceFileNotFoundError(path) from None
@@ -132,10 +138,11 @@ class DICOMFileTileSource(FileTileSource, metaclass=LruCacheMetaclass):
             self._largeImagePath = path
         _lazyImport()
         try:
-            self._dicom = wsidicom.WsiDicom.open(self._largeImagePath)
-        except Exception:
-            msg = 'File cannot be opened via dicom tile source.'
+            self._dicom = self._open_wsi_dicom(self._largeImagePath)
+        except Exception as exc:
+            msg = f'File cannot be opened via dicom tile source ({exc}).'
             raise TileSourceError(msg)
+
         self.sizeX = int(self._dicom.size.width)
         self.sizeY = int(self._dicom.size.height)
         self.tileWidth = int(self._dicom.tile_size.width)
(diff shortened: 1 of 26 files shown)


sources/dicom/large_image_source_dicom/__init__.py
```diff
@@ -129,10 +134,12 @@ def __init__(self, path, **kwargs):
             self._largeImagePath = path
         _lazyImport()
         try:
-            self._dicom = wsidicom.WsiDicom.open(self._largeImagePath)
+            self._dicom = self._open_wsi_dicom(self._largeImagePath)
         except Exception:
-            msg = 'File cannot be opened via dicom tile source.'
+            full_e = traceback.format_exc()
```
manthey, line 139: Rather than importing traceback, it is probably sufficient to do
```
except Exception as exc:
  msg = f'File cannot be opened via dicom tile source ({exc}).'
```

sources/dicom/large_image_source_dicom/__init__.py
```diff
@@ -110,11 +112,14 @@ def __init__(self, path, **kwargs):
...
 
+        self.logger = config.getConfig('logger')
+
         # We want to make a list of paths of files in this item, if multiple,
         # or adjacent items in the folder if the item is a single file.  We
         # filter files with names that have a preferred extension.
+        # If the path is a dict, that likely means it is a DICOMweb asset.
         path = self._getLargeImagePath()
-        if not isinstance(path, list):
+        if not isinstance(path, (dict, list)):
```
manthey, line 122: In the case that we have what looks like a string that isn't a path, should we see if the string is a json or yaml encoded dict (not on this PR)?  I think this might be needed if we are using a dicom web source as an entry in a multi-source source.

sources/dicom/large_image_source_dicom/assetstore/dicomweb_assetstore_adapter.py
```diff
@@ -0,0 +1,220 @@
...
+
+        for field in convert_empty_fields_to_none:
+            if isinstance(info.get(field), str) and not info[field].strip():
+                info[field] = None
+
+        # Now, if there is no authentication, verify that we can connect to the server.
+        # If there is authentication, we may need to prompt the user for their
+        # username and password sometime before here.
+        if info['auth_type'] is None:
+            study_instance_uid_tag = '0020000D'
```
manthey, line 50: We have pydicom as a soft dependency anyway.  Should we use `study_instance_uid_tag = f'{pydicom.datadict.tag_for_keyword("StudyID"):08x}'`

sources/dicom/large_image_source_dicom/assetstore/dicomweb_assetstore_adapter.py
```diff
@@ -0,0 +1,220 @@
...
+        msg = 'DICOMweb assetstores are import only.'
+        raise NotImplementedError(msg)
+
+    def deleteFile(self, file):
+        # We don't actually need to do anything special
+        pass
+
+    def downloadFile(self, file, offset=0, headers=True, endByte=None,
+                     contentDisposition=None, extraParameters=None, **kwargs):
+        # FIXME: do we want to support downloading files? We probably
```
manthey, line 95: We eventually _do_ want to support this (in some future PR).  When we pass files to workers, if this supports offset, then there are some transfer efficiencies if we don't need a complete file.

Trying to run a HistomicsTK task (like nuclei detection) on one of the imported files will tell us what we need to do for basic functionality and that will determine how valuable this is.

sources/dicom/large_image_source_dicom/assetstore/dicomweb_assetstore_adapter.py

sources/dicom/large_image_source_dicom/__init__.py
psavery: Fixed!

sources/dicom/test_dicom/test_web_client.py
```diff
@@ -0,0 +1,18 @@
...
+import sys
+
+import pytest
+
+pytestmark = [
+    pytest.mark.skipif(sys.version_info < (3, 8), reason='requires python3.8 or higher'),
+]
+
+
+@pytest.mark.girder_client
```
manthey, line 11: I think we need both the `@pytest.mark.girder` and `@pytest.mark.girder_client` to be consistent with other marks.

manthey: And, I think the linter will demand these are `@pytest.mark.girder()` and `@pytest.mark.girder_client()`.

sources/dicom/large_image_source_dicom/web_client/templates/assetstoreImport.pug
```diff
@@ -0,0 +1,27 @@
...
+
+form.g-dwas-import-form
+  .form-group
+    label(for="g-dwas-import-dest-type") Destination type
+    select#g-dwas-import-dest-type.form-control
+      option(value="folder", selected=true) Folder
+      option(value="user") User
+      option(value="collection") Collection
+  .form-group
+    label(for="g-dwas-import-dest-id") Parent ID
```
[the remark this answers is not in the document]
manthey: And the placeholder text could be `"Existing folder, user, or collection ID to use as the destination"` to match the others.

sources/dicom/large_image_source_dicom/__init__.py
```diff
@@ -143,6 +148,87 @@ def __init__(self, path, **kwargs):
...
+        # This is how we select the JPEG type to return
+        # The available transfer syntaxes used by wsidicom may be found here:
+        # https://github.com/imi-bigpicture/wsidicom/blob/a2716cd6a443f4102e66e35bbce32b0e2ae72dab/wsidicom/web/wsidicom_web_client.py#L97-L109
+        # (we may need to update this if they add more options)
+        # FIXME: maybe this function better belongs upstream in `wsidicom`?
+        from pydicom.uid import JPEG2000, JPEG2000Lossless, JPEGBaseline8Bit, JPEGExtended12Bit
+
+        # Prefer the transfer syntaxes in this order.
+        transfer_syntax_preferred_order = [
+            JPEGBaseline8Bit,
```
manthey, line 199: I notice that some of the dicoms on the webserver we were looking at have different transfer syntaxes (like JPEG-LS variants).  What happens if we do `transfer_syntax_preferred_order = pydicom.uid.AllTransferSyntaxes`?  Do not all of them work, or is the order bad?  This could be investigated in a future PR.

psavery: Well, our dependency `wsidicom` only supports a subset of those right now. For example see [here](https://github.com/imi-bigpicture/wsidicom/blob/9a7a33714823be42b830405a955383f90cb68d1e/wsidicom/web/wsidicom_web_client.py#L97-L109). I think we would encounter an error in `wsidicom` if we did not use one of those. Also see this [note in their README](https://github.com/imi-bigpicture/wsidicom#limitations).

I'm not sure if they would be able to support JPEG-LS, but we could perhaps ask!

Also, this function might belong better in `wsidicom` rather than here. We are simply making a REST request to see what type of format it supports, and then asking for that format.

sources/dicom/test_dicom/test_web_client.py
psavery: Fixed

sources/dicom/large_image_source_dicom/web_client/templates/assetstoreImport.pug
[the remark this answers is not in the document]
psavery: Fixed.

sources/dicom/large_image_source_dicom/__init__.py
```diff
@@ -143,6 +148,87 @@ def __init__(self, path, **kwargs):
...
+            JPEGExtended12Bit,
+            JPEG2000,
+            JPEG2000Lossless,
+        ]
+        available_transfer_syntax_tag = '00083002'
```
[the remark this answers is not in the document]
psavery: Fixed.

sources/dicom/large_image_source_dicom/__init__.py
manthey: Great.  I'd like to make an issue for more transfer format support on our repo; we can figure out how to contribute upstream or handle it ourselves when we get to it.

manthey (approved):
This is a fantastic start.

sources/dicom/large_image_source_dicom/assetstore/dicomweb_assetstore_adapter.py
psavery: This is now tracked in #1309.

sources/dicom/large_image_source_dicom/__init__.py
psavery: This is tracked in #1310.

Issue #1118, opened 2023-04-18
max-merge option on axis in frame selector


manthey:
When in axis mode, we could have a check box next to each axis that would show a max-merge on that axis.  When enabled, the slider would be disabled.

manthey:
Also called Max Intensity Projection

manthey:
Description:

Suppose a image with 3 z values and 4 time values.  We currently have two sliders, one for z and one for t.  For this example, suppose our image is white.  If you check "max-merge" next to an axis, the slider is disabled.  If you did this on z (and the IndexStride of Z is 1, The style is now `{"bands": [{"band": 1, "framedelta": 0, "palette": "white"}, {"band": 1, "framedelta": 1, "palette": "white"}, {"band": 1, "framedelta": 2, "palette": "white"}]}`.  Basically, we ask for all z-stack frames -- the default compositing is lighten, so you get the brightest pixel from those frames.  If you did max-merge of t (IndexStride of 3), you'd have four entries with the framedeltas of 0, 3, 6, 9.  If you did max-merge of both, you have all twelve frames in the style.

If you are in channel compositing mode, you repeat the palette for each channel for each max-merge entry and have the appropriate framedelta for both the channel and max-merge entry.

The net result of this is you see the brightest pixel from any frame that has been max-merged.  Somethings probably make no sense to max-merge (like an XY axis), but I wouldn't prohibit any of them -- if it has a slider, it should have a max-merge checkbox.

manthey:
This was added in #1306.

PR #1306, opened 2023-09-19
Max Merge option in Frame Selector

annehaley:
Addresses #1118

### DIFF
--- a/girder/girder_large_image/web_client/vue/components/CompositeLayers.vue
+++ b/girder/girder_large_image/web_client/vue/components/CompositeLayers.vue
@@ -50,7 +50,7 @@ export default {
             }
         },
         histogramParams() {
-            this.fetchCurrentFrameHistogram()
+            this.fetchCurrentFrameHistogram();
         }
     },
     mounted() {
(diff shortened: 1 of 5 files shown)


manthey:
This looks like the right approach.  In band compositing mode on the normmedia_8well_col2_livecellgfp.nd2 test file, I see nulls being send for the framedelta.  

And, one further wrinkle: we really should be showing the histogram of the max-merged image, not the un-merged image.  This means that we need to pass the appropriate style to the histogram endpoint when max-merge is checked.

manthey:
This is working, though there is an eccentricity in non-8-bit images.  Specifically, when a style gets applied, the histogram ends up in 8 bits, which then breaks the sliders setting correct values for the min / max (the auto range works fine).  I added a PR (#1326) that allows adding `"dtype": "source"` to the style at the top level (e.g., `{"bands": [...], "dtype": "source"}` that will return the histogram in the original data type.

Not necessarily for this PR:

Minor styling desires: Can we make "Max Merge" fit on the same line to reduce vertical height slightly?

The label `for` reference with max-merge quite do what you'd expect.  Specifically, if `for` property of a label contains the `id` of another element, then clicking on the label acts on that element.  Since this references the `name` property, it doesn't allow clicking on the words to activate the checkbox.

Otherwise, the only issue I see is one of performance -- we ask for the histograms faster than we can handle them resulting in having stale requests when scrubbing quickly.  Can we debounce that in any way?

PR #1338, opened 2023-10-16
Standardize DICOMweb assetstore parameter names/style

willdunklin:
These changes standardize parameter names in accordance with the girder /assetstore/ api

### DIFF
--- a/sources/dicom/large_image_source_dicom/assetstore/rest.py
+++ b/sources/dicom/large_image_source_dicom/assetstore/rest.py
@@ -2,9 +2,10 @@ import json
 
 from girder.api import access
 from girder.api.describe import Description, describeRoute
-from girder.api.rest import Resource, loadmodel
+from girder.api.rest import Resource, boundHandler
 from girder.constants import TokenScope
 from girder.exceptions import RestException
+from girder.models.assetstore import Assetstore
 from girder.utility import assetstore_utilities
 from girder.utility.model_importer import ModelImporter
 from girder.utility.progress import ProgressContext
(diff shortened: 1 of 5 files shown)


sources/dicom/large_image_source_dicom/assetstore/rest.py
```diff
@@ -51,20 +52,19 @@ def _importData(self, assetstore, params):
             )
 
     @access.admin
-    @loadmodel(model='assetstore')
+    @boundHandler
```
manthey, line 55: I don't think we want the `@boundHandler` decorator here.   Rather, with the changes made, I think changing the `@describeRoute` to `@autoDescribeRoute` is needed.

And, as long as we are in here, we should add a scope to the access decorator (e.g., change `@access.admin` to `@access.admin(scope=TokenScope.DATA_WRITE)`).

sources/dicom/test_dicom/web_client_specs/dicomWebSpec.js
```diff
@@ -70,15 +70,15 @@ describe('DICOMWeb assetstore', function () {
                 type: 'GET',
                 async: false,
                 data: {
-                    parentType: 'user',
-                    parentId: userId,
+                    destinationType: 'user',
```
manthey, line 73: These need to still be `parentType` and `parentId`, since it is the folder endpoint and that hasn't changed.

sources/dicom/test_dicom/web_client_specs/dicomWebSpec.js
```diff
@@ -103,8 +103,8 @@ describe('DICOMWeb assetstore', function () {
 
         runs(function () {
             // Set dest type and dest id
-            $('#g-dwas-import-dest-type').val(parentType);
-            $('#g-dwas-import-dest-id').val(parentId);
+            $('#g-dwas-import-dest-type').val(destinationType);
+            $('#g-dwas-import-dest-id').val(destinationId);
```
manthey, line 107: Because we are switching to autoDescribeRoute, some of the tests beyond here will fail.  Specifically, when checking that a float value produces an invalid limit (line 115 if it hasn't changed), we should probably check `return $('.g-validation-failed-message').html().includes('Invalid value');` rather than the `=== 'Invalid limit'`.

willdunklin: Would this also apply to line 128 as well?

manthey: No -- the autoDescribeRoute validates that it is an integer (so its error message applies), but not that it is positive (so the `Invalid Limit` message applies.

sources/dicom/test_dicom/web_client_specs/dicomWebSpec.js
```diff
@@ -103,16 +103,16 @@ describe('DICOMWeb assetstore', function () {
...
 
             // Test error for an invalid limit
             $('#g-dwas-import-limit').val('1.3');
             $('.g-submit-assetstore-import').trigger('click');
         });
 
         waitsFor(function () {
-            return $('.g-validation-failed-message').html() === 'Invalid limit';
+            return $('.g-validation-failed-message').html().includes('Invalid limit');
```
manthey, line 115: The autoDescribeRoute error is `Invalid value` (not `Invalid limit`).

manthey (approved):

Issue #1313, opened 2023-09-20
Add authentication options for DICOMweb servers


psavery:
For non-public DICOMweb servers, authentication will likely be required.

Since we and our dependency `wsidicom` are both using `dicomweb-client`, we are able to provide the server with [an arbitrary authentication handler derived from `requests.auth.AuthBase`](https://dicomweb-client.readthedocs.io/en/latest/usage.html#authentication-and-authorization). This provides us some flexibility. We should determine what types of authentication we need to support and how to expose them on the client.

We currently perform [some parameter validation when a DICOMweb assetstore is created for public servers](https://github.com/girder/large_image/blob/9cc82cdff75ddaf8943877d03129a67cf42c5073/sources/dicom/large_image_source_dicom/assetstore/dicomweb_assetstore_adapter.py#L51-L81), and we should consider whether we are able to support this validation for servers requiring authentication as well.

PR #1349, opened 2023-10-26
Add token authentication option for DICOMweb

psavery:
When the user is creating a DICOMweb assetstore, if they select the "token" authentication type, an input box will appear where the user must enter the token.

This token is saved in MongoDB, and is used for making all requests for this DICOMweb assetstore. That includes requests for importing the data (which may only be done by an admin), and requests for viewing the data (which may be done by anyone with access to the folder).

If an admin wishes to restrict users from viewing DICOMweb assets that were imported using this token, the admin must restrict access to the imported items via girder's folder access controls.

Fixes: #1313

Depends on: imi-bigpicture/wsidicom#117

### DIFF
--- a/sources/dicom/large_image_source_dicom/__init__.py
+++ b/sources/dicom/large_image_source_dicom/__init__.py
@@ -195,7 +195,7 @@ class DICOMFileTileSource(FileTileSource, metaclass=LruCacheMetaclass):
                                           requested_transfer_syntax=transfer_syntax)
 
     def _identify_dicomweb_transfer_syntax(self, client, study_uid, series_uid):
-        # "client" is a DICOMwebClient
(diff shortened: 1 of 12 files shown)


psavery:
I converted this back to a draft because I realized we are not actually passing this token-containing session to `wsidicom`. We need to fix that.

However, `wsidicom` [wants us to provide it](https://github.com/imi-bigpicture/wsidicom/blob/774fe3ca00096e3fd2556fc956520dbfabc81311/wsidicom/web/wsidicom_web_client.py#L39-L45) an object derived from `AuthBase`, and then they create a session out of it. They should also accept a session in case we want to create the session ourselves, so we might need to modify that upstream.

manthey (approved):

PR #1353, opened 2023-11-01
Add progress context to dicomWEB assetstore _importData

willdunklin:


sources/dicom/large_image_source_dicom/assetstore/rest.py
```diff
@@ -51,7 +51,7 @@ def _importData(self, assetstore, params):
             msg = f'Invalid filters: {e}'
             raise RestException(msg)
 
-        progress = self.boolParam('progress', params, default=False)
+        progress = progress or self.boolParam('progress', params, default=False)
```
manthey, line 54: Do you always want the `progress` arg to override the `params` entry if it is specified?  If so, this should probably be `progress = progress if progress is not None else self.boolParam('progress', params, default=False)`

willdunklin: I just pushed a change that reworks the progress context system to be more compliant with the standards for other girder importData APIs. Hopefully this resolves the lack of clarity with the prior expression here

manthey (approved):

PR #1359, opened 2023-11-02
Default to "None" for the limit

psavery:
On the master branch, if I try to import with no limit set, I get the following error:

```
Invalid value for integer parameter limit: .
```

We should remove the datatype restriction and set the default to `None`. That way, it will work properly.

### DIFF
--- a/sources/dicom/large_image_source_dicom/assetstore/rest.py
+++ b/sources/dicom/large_image_source_dicom/assetstore/rest.py
@@ -79,7 +79,7 @@ class DICOMwebAssetstoreResource(Resource):
                enum=('folder', 'user', 'collection'),
                required=True)
         .param('limit', 'The maximum number of results to import.',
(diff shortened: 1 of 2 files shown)


manthey:
I think with this change, we have to change line https://github.com/girder/large_image/blame/master/sources/dicom/test_dicom/web_client_specs/dicomWebSpec.js#L115 back to `return $('.g-validation-failed-message').html() === 'Invalid limit';` for the tests to pass.

psavery:
> I think with this change, we have to change line https://github.com/girder/large_image/blame/master/sources/dicom/test_dicom/web_client_specs/dicomWebSpec.js#L115 back to `return $('.g-validation-failed-message').html() === 'Invalid limit';` for the tests to pass.

Done. Looks like the CI is passing now.

manthey (approved):

PR #1372, opened 2023-11-14
Rename DICOMweb's AssetstoreImportView for clarity

willdunklin:
Changed ambiguous DICOMweb assetstore classes to include references to DICOMweb for clarity.


willdunklin:
Sounds good! The problem I was running into is that most assetstore import views have their scoped names included, for example an equivalent class would be `FilesystemImportView`. Makes more sense to me to not have the dicomweb assetstore be the definitive "`AssetstoreImportView`" across the codebase :)

manthey (approved):

Issue #1314, opened 2023-09-20
Add support for metadata to DICOMweb assets


psavery:
When we query for studies/series, we have the ability to request specific DICOM fields. This is already done, for instance, when [we perform the import](https://github.com/girder/large_image/blob/9cc82cdff75ddaf8943877d03129a67cf42c5073/sources/dicom/large_image_source_dicom/assetstore/dicomweb_assetstore_adapter.py#L156-L159).

We can also request for specific fields we intend to store as metadata, such as the `PatientID`.

We should request for and store certain fields as metadata, and potentially allow the users to edit the metadata as well (such as adding their own metadata).

PR #1378, opened 2023-11-17
Read and set several metadata items, if available

psavery:
When DICOMweb series are imported, several metadata entries from each series are now extracted and set on the item metadata.

We are extracting most of the metadata that is displayed in the SLIM viewer: https://imagingdatacommons.github.io/slim/

Most of the metadata entries are available as series-level metadata items. However, some of the metadata entries are only available on the instances themselves (in particular, the "Specimen" information, which is nested several dataset layers deep).

The SLIM viewer sets most of its metadata by locating the first WSI Volume dataset, and using that (it appears that the metadata for WSI volume datasets in a series are almost completely identical to other WSI volume datasets in the series). We are also using that approach here.

We are using a slightly different approach for displaying the Specimen metadata, but that is something we can always improve upon.

Additionally, the SLIM viewer is finding the first WSI Volume dataset in a whole study, and using that for the metadata for the whole study. We are currently doing it on a series-level, though (we find the first WSI Volume dataset for a series, and use that to set the metadata on the girder item that represents that series).

Here are some examples:

TCGA-05-4244

C3N-01016

HTA9_1

Fixes: #1314

### DIFF
--- a/sources/dicom/large_image_source_dicom/assetstore/dicomweb_assetstore_adapter.py
+++ b/sources/dicom/large_image_source_dicom/assetstore/dicomweb_assetstore_adapter.py
@@ -192,7 +192,7 @@ class DICOMwebAssetstoreAdapter(AbstractAssetstoreAdapter):
 
             # Set the DICOMweb metadata
             item['dicomweb_meta'] = get_dicomweb_metadata(client, study_uid, series_uid)
-            Item().save(item)
+            item = Item().save(item)
 
             # Create a placeholder file with the same name
             file = File().createFile(
--- a/sources/dicom/large_image_source_dicom/assetstore/dicomweb_assetstore_adapter.py
+++ b/sources/dicom/large_image_source_dicom/assetstore/dicomweb_assetstore_adapter.py
@@ -190,13 +190,9 @@ class DICOMwebAssetstoreAdapter(AbstractAssetstoreAdapter):
             item = Item().createItem(name=series_uid, creator=user, folder=folder,
                                      reuseExisting=True)
 
-            # Set the DICOMweb metadata, if available
-            metadata = get_dicomweb_metadata(client, study_uid, series_uid)
-            if metadata:
-                Item().setMetadata(item, metadata)
-                # Also save a copy that the user cannot modify, for _getDicomMetadata()
-                item['dicomweb_meta'] = metadata
-                Item().save(item)
+            # Set the DICOMweb metadata
+            item['dicomweb_meta'] = get_dicomweb_metadata(client, study_uid, series_uid)
+            Item().save(item)
 
             # Create a placeholder file with the same name
             file = File().createFile(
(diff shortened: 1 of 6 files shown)


manthey:
This is great.  In other tile sources, we have an internal source dictionary (read when the source is opened) that is exposed via the tile source's `getInternalMetadata` method.  This data can optionally be shown on the item page (see the settings on the large_image plugin).  I'm wondering if this would be better served to be exposed that way, as that data would be available when large_image is used without girder and the dicom image is opened just via url.

psavery:
> This is great. In other tile sources, we have an internal source dictionary (read when the source is opened) that is exposed via the tile source's `getInternalMetadata` method. This data can optionally be shown on the item page (see the settings on the large_image plugin). I'm wondering if this would be better served to be exposed that way, as that data would be available when large_image is used without girder and the dicom image is opened just via url.

Yeah, we can do this! Would we move the REST calls that request for the DICOMweb metadata somewhere else, like inside `getInternalMetadata()`? Or, for girder, keep them in the database, and add a new method that `DICOMGirderTileSource` will override that gets the metadata from the database?

manthey:
> > This is great. In other tile sources, we have an internal source dictionary (read when the source is opened) that is exposed via the tile source's `getInternalMetadata` method. This data can optionally be shown on the item page (see the settings on the large_image plugin). I'm wondering if this would be better served to be exposed that way, as that data would be available when large_image is used without girder and the dicom image is opened just via url.
> 
> Yeah, we can do this! Would we move the REST calls that request for the DICOMweb metadata somewhere else, like inside `getInternalMetadata()`? Or, for girder, keep them in the database, and add a new method that `DICOMGirderTileSource` will override that gets the metadata from the database?

Yes, inside `getInternalMetadata` is fine, but we should cache the results so if `getInternalMetadata` is called multiple times we only fetch it once.  If that cache is done on a method (e.g., `getInternalMetadata()` calls `_getDicomMetadata()` or some better name), then we should be able to just override that method in the girder tile source; we don't have to do a database lookup there, since the `meta` tags will be located in `self.item['meta']` (but, bear in mind that other metadata might have been added, too).

psavery:
@manthey Here is a summary of changes from the latest commit. Please feel free to ask for any changes:

1. A `_getDicomMetadata()` function was added, that `getInternalMetadata()` now calls. `getInternalMetadata()` now returns two forms of the metadata: one in the `dicom` key (the old way, using `dicom_to_dict`), and one in the `dicom_meta` key, using the new `_getDicomMetadata()`.
2. The `_getDicomMetadata()` is cached, and it works for both girder and non-girder tile sources (the girder version overrides the `_getDicomMetadata()` method so it can get the results from the database). It also works for both file tile sources and DICOMweb tile sources.
3. The girder DICOMweb version stores this metadata in the database in two places: one is on `meta` so the user can see it (and modify it), and the other is `dicom_meta`, which is uneditable by the user and used by `_getDicomMetadata()`.

sources/dicom/large_image_source_dicom/assetstore/dicomweb_assetstore_adapter.py
```diff
@@ -190,6 +190,10 @@ def importData(self, parent, parentType, params, progress, user, **kwargs):
             item = Item().createItem(name=series_uid, creator=user, folder=folder,
                                      reuseExisting=True)
 
+            # Set the DICOMweb metadata
+            item['dicomweb_meta'] = get_dicomweb_metadata(client, study_uid, series_uid)
+            Item().save(item)
```
manthey, line 195: We should probably use `item = Item().save(item)`.  The item _could_ be altered via some girder event, so it is considered good form to do this to ensure that the returned item reflects this if it happened.

manthey (approved):

PR #1387, opened 2023-11-28
Frame Selector component: ensure that no color can be assigned twice

annehaley:
A user found that two channels called "DAPI" and "DAPI 1" are both assigned the same color. This is because they both match the following regex: `'^DAPI.*$'`. Only the first match should receive the mapped color for that regex.

This PR adds a check to the `getChannelColor` function that the color does not already exist in a list of used colors.


manthey:
A recent change to bugbear causes our linting to fail.  I have a fix in https://github.com/girder/large_image/pull/1388.

manthey (approved):

Issue #1308, opened 2023-09-20
Add support for more DICOMweb transfer syntaxes


psavery:
See discussion [here](https://github.com/girder/large_image/pull/1303#discussion_r1330152068).

This may require making changes to [wsidicom](https://github.com/imi-bigpicture/wsidicom).

psavery:
If [this issue](https://github.com/imi-bigpicture/wsidicom/issues/119) is fixed, it will largely fix this issue as well.

psavery:
If we add the optional dependencies `imagecodecs` and `pylibjpeg-rle`, several more transfer syntaxes will be supported. The full list can be found [here](https://github.com/imi-bigpicture/wsidicom#limitations).

manthey:
We can add `imagecodecs` as a dependency directly, since it has good cross-platform support.

For `pylibjpeg-rle`, it doesn't have binary wheels for python >= 3.11, which means that we can't install it on modern python without a rust compiler.  We can either add this with a version restriction (< 3.11, which means we'll need to update it when they release a newer wheel), or we add it as an extra requires and build our own linux wheels for it.  For the nonce, maybe just add it with a version restriction?

psavery:
Pydicom's RLE decoder [only requires numpy to be present](https://github.com/pydicom/pydicom/blob/c5f756b70899bcbc0351809dab24bff5ea8c00d3/src/pydicom/pixel_data_handlers/rle_handler.py#L42-L48). I asked Erik if he would consider using that instead.

PR #1393, opened 2023-11-30
Remove logic for determining transfer syntax

psavery:
wsidicom now determines the transfer syntax automatically, as of version 0.14.0, so we don't need to do this anymore.

wsidicom also supports more transfer syntaxes as well, so the list of transfer syntaxes we had are not complete now either. Just remove this logic and let wsidicom take care of it.

wsidicom now has optional dependencies: imagecodecs and pylibjpeg-rle. If we install those optional dependencies, it adds support for several more transfer syntaxes. Those are found [here](https://github.com/imi-bigpicture/wsidicom#limitations).

Fixes: #1308


psavery:
Since wsidicom no longer supports Python3.8, we'll need to only support DICOMweb for Python > 3.8. We'll make that change in this PR too.

manthey (approved):

PR #1396, opened 2023-12-05
Fix an issue when tifffile reports no key frame

manthey:


manthey:
Closes https://github.com/Kitware/UPennContrast/issues/609

PR #1418, opened 2024-01-02
Add DCM4CHEE server for DICOMweb tests in CircleCI

psavery:
In CircleCI, this PR spins up a local DCM4CHEE server and uploads example DICOM files to it. This server is then used for the DICOMweb tests.

~~This is almost working. The DICOMweb tests are still skipped, and I need to get the `DICOMWEB_TEST_URL` environment variable to propagate to the pytests...~~

~~The `DICOMWEB_TEST_URL` is now being passed via tox's `passenv`. However, I still need to access the environment variable in the JavaScript test. I tried to do so via `process.env`, but it is apparently not there...~~

We made the JavaScript spec file into a simple template and inserted the `DICOMWEB_TEST_URL` into it before running it.

This is ready!

### DIFF
--- a/.circleci/dcm4chee/upload_example_data.py
+++ b/.circleci/dcm4chee/upload_example_data.py
@@ -27,5 +27,10 @@ def upload_example_data(server_url):
 
 
 if __name__ == '__main__':
-    url = 'http://localhost:8008/dcm4chee-arc/aets/DCM4CHEE/rs'
+    import os
+
+    url = os.getenv('DICOMWEB_TEST_URL')
+    if url is None:
+        raise Exception('DICOMWEB_TEST_URL must be set')
+
     upload_example_data(url)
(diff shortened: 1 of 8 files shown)


.circleci/dcm4chee/upload_example_data.py
```diff
@@ -0,0 +1,36 @@
...
+
+from dicomweb_client import DICOMwebClient
+from pydicom import dcmread
+
+
+def upload_example_data(server_url):
+
+    # This is TCGA-AA-3697
+    download_urls = [
+        'https://data.kitware.com/api/v1/file/6594790e9c30d6f4e17c9d0d/download',
```
manthey, line 14: I'd prefer to use the hashsum download rather than the file id download, that way if the files are moved they can still be accessed (that is, use `https://data.kitware.com/api/v1/file/hashsum/{algo}/{hashvalue}/download`).

psavery: Done!

manthey (approved):

PR #1426, opened 2024-01-08
Limit amount of RAM that dcm4chee can take up

psavery:
The `arc` container tends to take up around 650 MB of RAM. Limit it to 500 MB. We can increase it if it seems unstable.


psavery:
This is no longer needed due to #1425. Closing.

Issue #1309, opened 2023-09-20
Add support for downloading files from a DICOMweb server


psavery:
This will require implementing [this function](https://github.com/girder/large_image/blob/9cc82cdff75ddaf8943877d03129a67cf42c5073/sources/dicom/large_image_source_dicom/assetstore/dicomweb_assetstore_adapter.py#L95-L109).

It would be beneficial to support offsetting, if possible.

Note the discussion [here](https://github.com/girder/large_image/pull/1303#discussion_r1329076538).

psavery:
One thing that complicates this is that currently each item is associated with a series, and there is a single file per item. However, each series typically contains more than one DICOM file (which is often referred to as an "instance"). 

We could, for instance, use [this function](https://dicomweb-client.readthedocs.io/en/latest/package.html#dicomweb_client.api.DICOMClient.retrieve_series) to retrieve all instances in a series. However, this is likely to return multiple DICOM datasets (each of which can be written to a file).

Alternatively, [this function](https://dicomweb-client.readthedocs.io/en/latest/package.html#dicomweb_client.api.DICOMClient.retrieve_instance) can be used to retrieve a single instance (a.k.a., a single DICOM dataset).

Perhaps we should modify our assetstore model to make separate files for each instance, so we can more easily support downloading files?

For actually downloading the files themselves (as raw data in chunks): the `DICOMwebClient` API doesn't provide us an easy way to do this. However, I think we can take inspiration from their code. For instance, [this function](https://github.com/ImagingDataCommons/dicomweb-client/blob/39dc8a799fac508756d9df57cf16fea51d54a9c3/src/dicomweb_client/web.py#L1026-L1033) is a way to obtain a resource in chunks. We could obtain a DICOM file this way (using the `application/dicom` content type), as long as we can figure out the URL as well.

manthey:
If we used the `retrieve_series` to get all of the files conceptually, and downloaded them and then created a zip of those, does that zip file open with our dicom source reader?  If so, perhaps we should expose that conceptually (range requests will be tricky, but if the individual files don't have further compression inside the zip, this could work).

PR #1429, opened 2024-01-09
Add support for downloading DICOMweb files

psavery:
Each girder item represents a series. Previously, each item would contain a single file with the same name.

Now, each instance (DICOM file) within a series is represented by separate girder files within the item.

This PR includes download support, so individual DICOM files can be downloaded, or all DICOM files in a series can be downloaded together by downloading the girder item.

Range requests for the DICOM files are attempted if requested. However, the DICOMweb standard doesn't guarantee that they will be supported, and none of the DICOMweb servers we have tried have supported range requests for DICOM files.

To do:

- [x] Figure out why downloading the file has the wrong mimetype (which results in the web browser displaying the contents rather than downloading)

Fixes: #1309

### DIFF
--- a/sources/dicom/large_image_source_dicom/assetstore/dicomweb_assetstore_adapter.py
+++ b/sources/dicom/large_image_source_dicom/assetstore/dicomweb_assetstore_adapter.py
@@ -104,19 +104,39 @@ class DICOMwebAssetstoreAdapter(AbstractAssetstoreAdapter):
 
     def downloadFile(self, file, offset=0, headers=True, endByte=None,
                      contentDisposition=None, extraParameters=None, **kwargs):
-        # FIXME: do we want to support downloading files? We probably
-        # wouldn't download them the regular way, but we could instead
-        # use a dicomweb-client like so:
-        # instance = client.retrieve_instance(
-        #     study_instance_uid=...,
-        #     series_instance_uid=...,
-        #     sop_instance_uid=...,
-        # )
-        # pydicom.filewriter.write_file('output_name.dcm', instance)
-        msg = 'Download support not yet implemented for DICOMweb files.'
-        raise NotImplementedError(
-            msg,
-        )
+
+        dicomweb_meta = file['dicomweb_meta']
+        study_uid = dicomweb_meta['study_uid']
+        series_uid = dicomweb_meta['series_uid']
+        instance_uid = dicomweb_meta['instance_uid']
+
+        client = _create_dicomweb_client(self.assetstore_meta)
+
+        def stream():
+            from dicomweb_client.web import _Transaction
+
+            url = client._get_instances_url(
+                _Transaction.RETRIEVE,
+                study_uid,
+                series_uid,
+                instance_uid
+            )
+
+            transfer_syntax = '*'
+            accept_parts = [
+                'multipart/related',
+                'type="application/dicom"',
+                f'transfer-syntax={transfer_syntax}',
+            ]
+            headers = {
+                'Accept': '; '.join(accept_parts),
+            }
+
+            response = client._http_get(url, headers=headers)
+            for part in client._decode_multipart_message(response, False):
+                yield part
+
+        return stream
 
     def importData(self, parent, parentType, params, progress, user, **kwargs):
         """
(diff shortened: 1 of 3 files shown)


sources/dicom/large_image_source_dicom/girder_source.py
```diff
@@ -59,15 +59,14 @@ def _getFilesystemLargeImagePath(self):
 
     def _getDICOMwebLargeImagePath(self, assetstore):
         meta = assetstore[DICOMWEB_META_KEY]
-        file = Item().childFiles(self.item, limit=1)[0]
-        file_meta = file['dicomweb_meta']
```
psavery, line 63: This PR is mostly backward-compatible with existing imports. However, we previously stored the study/series UIDs in the file (within the item) and it was in the `dicomweb_meta` key. Now, the study/series UIDs are stored in the item, and the key is `dicom_uids`.

Do we want to keep this PR backward-compatible with existing imports? If so, I can first check if `dicom_uids` is present in the item. If not, then I can look for `dicomweb_meta` in the file.

manthey: We don't need to keep it backward-compatible.

psavery:
This is ready for review! I added a test for downloading an item (which downloads each file).

During local testing, I downloaded some DICOM files, uploaded them back to girder, and confirmed I could view them with large image (so this confirms that the downloaded DICOM files work correctly).

sources/dicom/large_image_source_dicom/assetstore/dicomweb_assetstore_adapter.py
```diff
@@ -104,20 +107,161 @@ def deleteFile(self, file):
...
+        # since it is before the message body. An end boundary might look like this:
+        # \r\n--50d7ccd118978542c422543a7156abfce929e7615bc024e533c85801cd77--
+        if boundary is None:
+            content_type = response.headers['content-type']
+            msg = f'Failed to locate boundary in content-type: {content_type}'
+            raise ValueError(msg)
+
+        # Both dicomweb-client and requests-toolbelt check for
+        # the ending boundary exactly like so:
+        ending = b'\r\n--' + boundary
```
psavery, line 200: For reference, [here](https://github.com/requests/toolbelt/blob/c62d9774a3b1469f609bdc5a213f86c714e059bf/requests_toolbelt/multipart/decoder.py#L137-L149) is where `requests-toolbelt` does this. Note the lines:

```python
boundary = b''.join((b'--', self.boundary))
```

and 

```python
parts = content.split(b''.join((b'\r\n', boundary)))
```

So it is essentially `b'\r\n--' + self.boundary`.

And [here](https://github.com/ImagingDataCommons/dicomweb-client/blob/39dc8a799fac508756d9df57cf16fea51d54a9c3/src/dicomweb_client/web.py#L674-L675) is the same kind of thing in `dicomweb_client`.

sources/dicom/large_image_source_dicom/assetstore/dicomweb_assetstore_adapter.py
```diff
@@ -104,20 +107,161 @@ def deleteFile(self, file):
...
+
+        # Sometimes, there are a few extra bytes after the ending, such
+        # as '--' and '\r\n'. Imaging Data Commons has '--\r\n' at the end.
+        # We will make this number large enough to capture possible extra bytes.
+        max_ending_size = len(ending) + 6
```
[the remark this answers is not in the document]
psavery: Okay, this is fixed!

psavery:
Ready for review...

sources/dicom/large_image_source_dicom/assetstore/dicomweb_assetstore_adapter.py
```diff
@@ -192,25 +339,33 @@ def importData(self, parent, parentType, params, progress, user, **kwargs):
...
+                instance_uid = instance[instance_uid_key]['Value'][0]
+
+                file = File().createFile(
+                    name=f'{instance_uid}.dcm',
+                    creator=user,
+                    item=item,
+                    reuseExisting=True,
+                    assetstore=self.assetstore,
+                    mimeType='application/dicom',
+                    size=0,
```
manthey, line 359: If we set the size to `None` rather than `0`, the Girder UI won't print any size, rather than the misleading `0`.

manthey (approved):

PR #1440, opened 2024-01-19
Add authentication to DICOMweb testing

psavery:
This adds an additional docker-compose yaml, `auth-docker-compose.yml`, which spins up a dcm4chee server that requires authentication via keycloak. This authenticated server is now being used for the CircleCI testing.

This PR also adds a couple of scripts to work with this server, including an `update_access_token_lifespan.py` script (with the default access token lifespan of 10 minutes, the token would expire before it would be tested in the CI), and a `create_keycloak_token.py` script, which is used to create the access token that we provide to the assetstore.

Since we may want to perform local testing without authentication, the non-authenticated dcm4chee docker-compose is still there. The testing code checks for a `DICOMWEB_TEST_TOKEN` environment variable, where it expects to find a token. If that environment variable is set, then it uses token authentication with that token. If it is not set, then it assumes authentication is not required and attempts to run the tests without authentication.

### DIFF
--- a/.circleci/dcm4chee/update_access_token_lifespan.py
+++ b/.circleci/dcm4chee/update_access_token_lifespan.py
@@ -1,0 +1,40 @@
+#!/usr/bin/env python3
+
+# Change the access token life span in the realm settings
+
+import requests
+from keycloak import KeycloakOpenIDConnection
+
+
+def create_openid_admin_token():
+    # Get an admin OpenID access token. This expires after 60 seconds.
+    return KeycloakOpenIDConnection(
+        server_url='https://localhost:8843',
+        username='admin',
+        password='changeit',
+        realm_name='master',
+        verify=False,
+    ).token['access_token']
+
+
+def set_access_token_life_span(token, lifespan):
+    # curl command looks like this:
+    # curl 'https://localhost:8843/admin/realms/dcm4che' \
+    #   -X 'PUT' \
+    #   -H 'Content-Type: application/json' \
+    #   -H 'authorization: Bearer $TOKEN' \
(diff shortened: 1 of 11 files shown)


.circleci/dcm4chee/update_access_token_lifespan.py
```diff
@@ -0,0 +1,40 @@
...
+    url = 'https://localhost:8843/admin/realms/dcm4che'
+    r = session.put(url, json={'accessTokenLifespan': lifespan}, verify=False)
+    r.raise_for_status()
+
+
+if __name__ == '__main__':
+    token = create_openid_admin_token()
+
+    # Set default timetout to be 1 hour
+    set_access_token_life_span(token, 3600)
```
psavery, line 40: I hope 1 hour is long enough...

manthey (approved):

manthey:
@psavery Should we do anything to keep the various docker image tags up to date?  (Don't let this question delay merging).

psavery:
> @psavery Should we do anything to keep the various docker image tags up to date? (Don't let this question delay merging).

I'm not sure. [Here](https://github.com/dcm4che/dcm4chee-arc-light/wiki/Running-on-Docker) are all of the docker images, along with explanations of their versioning. Some of the versions between images have to match (eg., the slapd image has a X.X.X-Y.Y version format, where the Y.Y has to match the minor/patch of the archive image).

I just used the versions shown in the example `docker-compose` files, such as [here](https://github.com/dcm4che/dcm4chee-arc-light/wiki/Run-secured-archive-services-on-a-single-host#use-docker-compose).

PR #1446, opened 2024-01-29
Zarr Tile Sink

annehaley:
This PR makes changes to the Zarr tile source to allow the user to write new N-dimensional files. This includes the addition of `new`, `addTile`, and `write` methods.

This PR also rewrites `test/test_sink` to copy data from `TestFileTileSource` rather than randomly generate data.

TODOs:
- [x] Improve band bookkeeping
- [x] Allow user to crop data upon write

Future Work:
- [ ] #1454 
- [ ] Generate lower resolution tiles
- [ ] Multiprocessing support (https://github.com/girder/large_image/pull/1446#discussion_r1528412665)
- [ ] Add support for setting mm_x, mm_y, possibly minHeight and minWidth. It would be nice to also include in the list of properties we write to output files magnification and some form of image description (which could also just be an additional json record of metadata -- when generating data via algorithms, it is useful to store the parameters that were used).
- [ ] Add a method of prioritizing sinks so `large_image.new()` picks the most preferred one installed


sources/zarr/large_image_source_zarr/__init__.py
```diff
@@ -438,6 +488,204 @@ def getTile(self, x, y, z, pilImageAllowed=False, numpyAllowed=False, **kwargs):
...
+            zarr.copy_store(data_store, dir_store)
+            dir_store.close()
+
+        else:
+            from large_image_converter import convert
+
+            convert(data_file.name, path, overwrite=overwriteAllowed)
+
+        if self.crop:
+            data_file.close()
```
manthey, line 687: I can imagine a scenario: create a sink, add some data, save, add some more data, save again.  With this `close`, does that work using crop?

manthey:
I was playing with this a little, trying to replace a custom data collector with this branch.  I was expecting the class properties `sizeX` and `sizeY` properties (and `frames`, `bandCount`, `dtype`) to update based on the data collected.  Also, adding data is *slow*, because (a) we are using `np.pad` and (b) we are calling `create_dataset` on every change.  If `create_dataset` is being called just to rechunk things, we can pick a chunking of 512 in x and y, a largish value in samples (bands) like 16, and 1 for all other axes.

sources/zarr/large_image_source_zarr/__init__.py
```diff
@@ -79,7 +90,43 @@ def __init__(self, path, **kwargs):
...
+        self._output = None
+        self._editable = True
+        self._bandRanges = None
+        self._addLock = threading.RLock()
+        self._framecount = 0
+        self._mm_x = 0
+        self._mm_y = 0
+
+    def __del__(self):
+        try:
```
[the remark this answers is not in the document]
annehaley: Added the `if` guard and separated `try` blocks in [f36f1a8](https://github.com/girder/large_image/pull/1446/commits/f36f1a886d677234e70f7bf67caf6c9d1e03316d)

sources/zarr/large_image_source_zarr/__init__.py
```diff
@@ -396,6 +447,9 @@ def _getAssociatedImage(self, imageKey):
 
     @methodcache()
     def getTile(self, x, y, z, pilImageAllowed=False, numpyAllowed=False, **kwargs):
+        if self._levels is None:
```
manthey, line 450: I don't see anywhere where `self._levels` is set to None.  Maybe this needs to be done on `addTile`?

annehaley: This was previously done in `addTile`, but I recently changed it so that `addTile` sets `self._levels` to a valid value. I think I should take away this check and call to `_validateZarr`, since it was meant to fill those values when `addTile` didn't.

manthey: You are setting `self.levels` in `addTile`, not `self._levels`.  `self._levels` is a list of lists indicating which dataset is used for different resolutions (clearly a longer, more descriptive variable name would have been better).

As a simple test to add, after calling `addTile` with some data, call `getTile(0, 0, 0)` and see that it returns something.

annehaley: Oh I see the distinction now. I can put the line back in `addTile` where we set `self._levels = None`.

sources/zarr/large_image_source_zarr/__init__.py
```diff
@@ -438,6 +492,203 @@ def getTile(self, x, y, z, pilImageAllowed=False, numpyAllowed=False, **kwargs):
...
+            for axis, length in new_dims.items()
+            if axis in axes[:-3]
+        ])
+        self._levels = None
+        self.levels = int(max(1, math.ceil(math.log(max(
+            self.sizeX / self.tileWidth, self.sizeY / self.tileHeight)) / math.log(2)) + 1))
+
+        if not mask:
+            mask = np.full(tile.shape, True)
+        full_mask = np.full(tuple(new_dims.values()), False)
```
manthey, line 564: There really is no need to create this full mask.  It will not scale.  I want to build an image that won't fit in memory; this mask will be the size of the output and therefore won't fit in memory either.  I also wouldn't create a mask if none is specified.

Instead, `np.place(data, full_mask, tile)` can probably be 
```
if mask is not None:
    np.place(root[mask_placement_slices], mask, tile)
else:
   root[mask_placement_slices] = tile
```

annehaley: Changed in [cd0c715](https://github.com/girder/large_image/pull/1446/commits/cd0c715d61812c4e3e2eefddcf92dd659a65fe89). I had to do a slight variation because `np.place` won't edit the zarr data in place, so I used a placeholder variable "data" which contains the sliced section of the zarr array.

manthey: Maybe the `np.where` idiom would be clearer so you don't need the `data` placeholder variable:
```
if mask:
    root[mask_placement_slices] = np.where(mask, tile, root[mask_placement_slices])
else:
    root[mask_placement_slices] = tile
```

sources/zarr/large_image_source_zarr/__init__.py
```diff
@@ -438,6 +492,205 @@ def getTile(self, x, y, z, pilImageAllowed=False, numpyAllowed=False, **kwargs):
...
+                self._zarr.create_dataset('root', data=tile, chunks=chunking)
+            else:
+                root = current_arrays['root']
+                root.resize(*tuple(new_dims.values()))
+                data = root[mask_placement_slices]
+                if mask is not None:
+                    np.place(data, mask, tile)
+                else:
+                    data = tile
+                np.place(data, full_mask, tile)
```
manthey, line 586: This line is no longer required.

sources/zarr/large_image_source_zarr/__init__.py
```diff
@@ -438,6 +492,205 @@ def getTile(self, x, y, z, pilImageAllowed=False, numpyAllowed=False, **kwargs):
...
+                root[mask_placement_slices] = data
+
+            # Edit OME metadata
+            self._zarr.attrs.update({
+                'multiscales': [{
+                    'version': '0.5-dev',
```
manthey, line 592: I haven't checked ome zarr docs for version information.  Are we more conformant to a dev version than the latest official version?

sources/zarr/large_image_source_zarr/__init__.py
annehaley: Removed in [41ea6b5](https://github.com/girder/large_image/pull/1446/commits/41ea6b5e7938dc867122ad6d3da5836b58a47aed)

sources/zarr/large_image_source_zarr/__init__.py
annehaley: Changed in [41ea6b5](https://github.com/girder/large_image/pull/1446/commits/41ea6b5e7938dc867122ad6d3da5836b58a47aed)

sources/zarr/large_image_source_zarr/__init__.py
annehaley: I was just following what I found here: https://ngff.openmicroscopy.org/latest/#multiscale-md, but I'm not entirely confident that this is all correct.

sources/zarr/large_image_source_zarr/__init__.py
```diff
@@ -438,6 +492,202 @@ def getTile(self, x, y, z, pilImageAllowed=False, numpyAllowed=False, **kwargs):
...
+                chunking = tuple([
+                    self._tileSize if a in ['x', 'y'] else
+                    32 if a == 's' else 1
+                    for a in axes
+                ])
+                self._zarr.create_dataset('root', data=tile, chunks=chunking)
+            else:
+                root = current_arrays['root']
+                root.resize(*tuple(new_dims.values()))
+                if mask:
```
manthey, line 581: This needs to be `if mask is not None` since otherwise numpy complains `ValueError: The truth value of an array with more than one element is ambiguous. Use a.any() or a.all()`.

sources/zarr/large_image_source_zarr/__init__.py
```diff
@@ -438,6 +492,202 @@ def getTile(self, x, y, z, pilImageAllowed=False, numpyAllowed=False, **kwargs):
...
+        self._framecount = np.prod([
+            length
+            for axis, length in new_dims.items()
+            if axis in axes[:-3]
+        ])
+        self._levels = None
+        self.levels = int(max(1, math.ceil(math.log(max(
+            self.sizeX / self.tileWidth, self.sizeY / self.tileHeight)) / math.log(2)) + 1))
+
+        if not mask:
```
manthey, line 562: We don't need to create a mask if none was specified.  The `mask_placement_slices` could also be renamed to `placement_slices` and use `tile.shape` instead of `mask.shape`

annehaley: Changed in [3e62aae](https://github.com/girder/large_image/pull/1446/commits/3e62aaeafa9338e9fa7559d6e7c4f234df4919e6)

sources/zarr/large_image_source_zarr/__init__.py
annehaley: Changed in [3e62aae](https://github.com/girder/large_image/pull/1446/commits/3e62aaeafa9338e9fa7559d6e7c4f234df4919e6)

manthey:
Awesome.  This is now working with and without masks in my test case.

One interface thing with masks, I have a case where I have a multi band image but a single mask that applies to the whole thing.  I wonder if we we should do something like:
`if mask is not None and len(mask.shape) + 1 == len(tile.shape):  mask = mask[:, :, np.newaxis]`

I'd still like to test some of the axes options.

manthey:
I get an error with 2D tiles.  Example:
```
import numpy as np
s = large_image_source_zarr.new()
s.addTile(np.random.random((10, 10)), 0, 0)
s.addTile(np.random.random((10, 10)), 10, 0)
```
throws
```
Traceback (most recent call last):
  File "<stdin>", line 1, in <module>
  File "/home/manthey/misc/large_image/sources/zarr/large_image_source_zarr/__init__.py", line 580, in addTile
    root.resize(*tuple(new_dims.values()))
  File "/home/manthey/env/lib/python3.11/site-packages/zarr/core.py", line 2615, in resize
    return self._write_op(self._resize_nosync, *args)
           ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/home/manthey/env/lib/python3.11/site-packages/zarr/core.py", line 2583, in _write_op
    return self._synchronized_op(f, *args, **kwargs)
           ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/home/manthey/env/lib/python3.11/site-packages/zarr/core.py", line 2574, in _synchronized_op
    result = f(*args, **kwargs)
             ^^^^^^^^^^^^^^^^^^
  File "/home/manthey/env/lib/python3.11/site-packages/zarr/core.py", line 2620, in _resize_nosync
    new_shape = normalize_resize_args(old_shape, *args)
                ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/home/manthey/env/lib/python3.11/site-packages/zarr/util.py", line 258, in normalize_resize_args
    raise ValueError("new shape must have same number of dimensions")
ValueError: new shape must have same number of dimensions

```

annehaley:
> I get an error with 2D tiles.

Nice catch. I didn't realize that the call to `_imageToNumpy(tile)` would add a samples axis to the tile. With your example, that line changes the tile shape from `(10, 10)` to `(10, 10, 1)`.  We have some options for a fix depending on what the desired behavior is. If we don't want this samples axis added, I can just replace this line (or remove it entirely if it doesn't add value). If we do, I can just move it to the beginning of the `addTile` method so that `axes` and `new_dims` will be updated accordingly. What do you recommend?

manthey:
> > I get an error with 2D tiles.
> 
> Nice catch. I didn't realize that the call to `_imageToNumpy(tile)` would add a samples axis to the tile. With your example, that line changes the tile shape from `(10, 10)` to `(10, 10, 1)`. We have some options for a fix depending on what the desired behavior is. If we don't want this samples axis added, I can just replace this line (or remove it entirely if it doesn't add value). If we do, I can just move it to the beginning of the `addTile` method so that `axes` and `new_dims` will be updated accordingly. What do you recommend?

Elsewhere in large image, we _always_ have a samples axis.  I suppose for consistency, if we are handed a tile it MUST have x and y axes; if it doesn't have a samples axis, we would add it.  _imageToNumpy will almost certainly do the wrong thing if the source data isn't just YX or YXS (but it is useful to convert non-numpy arrays to numpy arrays).  The logic should probably account for that (e.g., if not a numpy array or axes are unspecified, pass it through _imageToNumpy; if a numpy array and axes don't include S, add it at the end).

manthey:
Playing with this, the chunking has a profound effect on speed.  Currently we are using (as I had recommended) a chunk size of _tileSize for x and y, 32 for s, and 1 for all other axes.

For s we definitely want to make this the dimension of the first tile's s axis.  There will be cases where we later grow s -- in the ideal, we would probably want to rechunk the dataset when this occurs.  In my test with a s dimension of 1 or 2, having it the actual value was fastest; having it 4 was only slightly slower, having it 32 was deeply slower (added 1.5 seconds per 2.5k square tile addition).

For x and y, I get a slight speed up with a larger chunk size, but it seems to be in the noise.

manthey:
After this PR is merged, we also want to add support for setting `mm_x`, `mm_y`, possibly `minHeight` and `minWidth`.  It would be nice to also include in the list of properties we write to output files `magnification` and some form of image description (which could also just be an additional json record of metadata -- when generating data via algorithms, it is useful to store the parameters that were used).

sources/zarr/large_image_source_zarr/__init__.py
```diff
@@ -438,6 +492,208 @@ def getTile(self, x, y, z, pilImageAllowed=False, numpyAllowed=False, **kwargs):
...
+
+        placement_slices = tuple([
+            slice(placement.get(a, 0), placement.get(a, 0) + tile.shape[i], 1)
+            for i, a in enumerate(axes)
+        ])
+        if mask is not None and len(mask.shape) + 1 == len(tile.shape):
+            mask = mask[:, :, np.newaxis]
+
+        current_arrays = dict(self._zarr.arrays())
+        with self._addLock:
```
manthey, line 567: We need to make sure any access to `self.<attr>` that could change is inside of the `_addLock`.  Moving it above `self._axes = ...` would work, but maybe some other lines could be moved outside of the lock.

manthey:
I was expecting that `addTile(<tile with shape y=256, x=256, s=4>, z=1)` would create a zarr file with four dimensions (z=[0, 1], y=[0, 255], x=[0, 255], s=[0, 3]), but it doesn't add the additional axis from the kwargs.

manthey (changes requested):
I modified the algorithm progression example to try things out.  Changing the storage type really makes a big difference in speed.  For a small image with 6250 parameter combinations, the SQLite store took ~21 minutes, the ZipStore 6.5 minutes, and the DirectoryStore 3.5 minutes (all outputting to a zip file).  Looking at the cpu utilization, with Zip and Directory stores, around 13-15 cores gets saturated; with SQLite, 2-3 cores get saturated.

sources/zarr/large_image_source_zarr/__init__.py
```diff
@@ -438,6 +492,208 @@ def getTile(self, x, y, z, pilImageAllowed=False, numpyAllowed=False, **kwargs):
...
+
+        elif suffix == '.zarr':
+            dir_store = zarr.storage.DirectoryStore(path)
+            zarr.copy_store(data_store, dir_store)
+            dir_store.close()
+
+        else:
+            from large_image_converter import convert
+
+            convert(data_file.name, path, overwrite=overwriteAllowed)
```
manthey, line 692: If we change to a DirectoryStore for the temp collector, this would change to something line `convert(Path(data_file.name) / '.zattrs', ...)`

sources/zarr/large_image_source_zarr/__init__.py
annehaley: I moved all the changes to the internal attributes to the end of `addTile`, within the `addLock` context, here: [6706f69](https://github.com/girder/large_image/pull/1446/commits/6706f6988764ffcbc02886e9ebd5e4d5333c7496). Is this what you mean?

sources/zarr/large_image_source_zarr/__init__.py
```diff
@@ -79,7 +90,47 @@ def __init__(self, path, **kwargs):
...
             msg = 'File cannot be opened -- not an OME NGFF file or understandable zarr file.'
             raise TileSourceError(msg)
-        self._tileLock = threading.RLock()
+
+    def _initNew(self, path, **kwargs):
+        """
+        Initialize the tile class for creating a new image.
+        """
+        self._tempfile = tempfile.NamedTemporaryFile(suffix=path)
+        self._zarr_store = zarr.SQLiteStore(self._tempfile.name)
```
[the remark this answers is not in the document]
annehaley: I switched to using `DirectoryStore` in [5118fbc](https://github.com/girder/large_image/pull/1446/commits/5118fbc7121f470065f846a66b7ad49858c292a5). This will need further testing and refinement, though.

sources/zarr/large_image_source_zarr/__init__.py
```diff
@@ -438,6 +492,208 @@ def getTile(self, x, y, z, pilImageAllowed=False, numpyAllowed=False, **kwargs):
...
+        ])
+        self._levels = None
+        self.levels = int(max(1, math.ceil(math.log(max(
+            self.sizeX / self.tileWidth, self.sizeY / self.tileHeight)) / math.log(2)) + 1))
+
+        placement_slices = tuple([
+            slice(placement.get(a, 0), placement.get(a, 0) + tile.shape[i], 1)
+            for i, a in enumerate(axes)
+        ])
+        if mask is not None and len(mask.shape) + 1 == len(tile.shape):
```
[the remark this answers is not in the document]
annehaley: Changed in [756502b](https://github.com/girder/large_image/pull/1446/commits/756502b977233628d11f8239cc6e3fbbb99bd258)

sources/zarr/large_image_source_zarr/__init__.py
```diff
@@ -438,6 +492,208 @@ def getTile(self, x, y, z, pilImageAllowed=False, numpyAllowed=False, **kwargs):
...
+                cropped_zarr.create_dataset(arr_name, data=cropped_arr, overwrite=True)
+                cropped_zarr.attrs.update(self._zarr.attrs)
+
+        data_file.flush()
+
+        if suffix in ['.db', '.sqlite']:
+            shutil.copy2(data_file.name, path)
+
+        elif suffix == '.zip':
+            zip_store = zarr.storage.ZipStore(path)
```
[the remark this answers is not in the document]
annehaley: Changed in [5118fbc](https://github.com/girder/large_image/pull/1446/commits/5118fbc7121f470065f846a66b7ad49858c292a5)

sources/zarr/large_image_source_zarr/__init__.py
```diff
@@ -438,6 +492,208 @@ def getTile(self, x, y, z, pilImageAllowed=False, numpyAllowed=False, **kwargs):
...
+            zarr.copy_store(data_store, zip_store)
```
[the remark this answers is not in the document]
annehaley: Changed in [5118fbc](https://github.com/girder/large_image/pull/1446/commits/5118fbc7121f470065f846a66b7ad49858c292a5)

sources/zarr/large_image_source_zarr/__init__.py
```diff
@@ -438,6 +492,208 @@ def getTile(self, x, y, z, pilImageAllowed=False, numpyAllowed=False, **kwargs):
...
+                    indices=range(y, y + h),
+                    axis=self._axes.get('y'),
+                )
+                cropped_zarr.create_dataset(arr_name, data=cropped_arr, overwrite=True)
+                cropped_zarr.attrs.update(self._zarr.attrs)
+
+        data_file.flush()
+
+        if suffix in ['.db', '.sqlite']:
+            shutil.copy2(data_file.name, path)
```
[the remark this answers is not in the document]
annehaley: I'm not sure. What does the error look like? I can search for if anyone has experienced it and found a solution.

sources/zarr/large_image_source_zarr/__init__.py
annehaley: Changed in [5118fbc](https://github.com/girder/large_image/pull/1446/commits/5118fbc7121f470065f846a66b7ad49858c292a5)

sources/zarr/large_image_source_zarr/__init__.py
```diff
@@ -438,6 +492,208 @@ def getTile(self, x, y, z, pilImageAllowed=False, numpyAllowed=False, **kwargs):
...
+
+        :param path: output path.
+        :param lossy: if false, emit a lossless file.
+        :param alpha: True if an alpha channel is allowed.
+        :param overwriteAllowed: if False, raise an exception if the output
+            path exists.
+        """
+        # TODO: compute half, quarter, etc. resolutions
+        if not overwriteAllowed and os.path.exists(path):
+            raise TileSourceError('Output path exists (%s).' % str(path))
```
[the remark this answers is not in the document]
annehaley: Changed in [5118fbc](https://github.com/girder/large_image/pull/1446/commits/5118fbc7121f470065f846a66b7ad49858c292a5)

manthey:
With #1477, there are now some tests that can be enabled.  Uncomment the two parametered tests on lines 
https://github.com/girder/large_image/blob/master/test/test_examples.py#L90-L91 ; they should pass for this branch before we consider it mergeable.

sources/zarr/large_image_source_zarr/__init__.py
```diff
@@ -52,12 +56,19 @@ def __init__(self, path, **kwargs):
...
+        self._tileLock = threading.RLock()
+
+    def _initOpen(self, **kwargs):
         self._largeImagePath = str(self._getLargeImagePath())
         self._zarr = None
         if not os.path.isfile(self._largeImagePath) and '//:' not in self._largeImagePath:
             raise TileSourceFileNotFoundError(self._largeImagePath) from None
         try:
-            self._zarr = zarr.open(zarr.SQLiteStore(self._largeImagePath), mode='r')
+            self._zarr = zarr.open(zarr.DirectoryStore(self._largeImagePath), mode='r')
```
manthey, line 71: I think we need to leave this line as trying to open an SQLiteStore.  `zarr.open(path)` will work with Directory and Zip stores (and being pointed to the zarr files (.zattrs, at least) within a directory), but won't open an sqlite store.

sources/zarr/large_image_source_zarr/__init__.py
manthey: Mostly.  The `self._axes = ` (above) needs to be within the lock context, too.

sources/zarr/large_image_source_zarr/__init__.py
annehaley: Changed back in [5317daa](https://github.com/girder/large_image/pull/1446/commits/5317daaa3ab2d8b89734fe14470df3f5d9902c1d)

sources/zarr/large_image_source_zarr/__init__.py
```diff
@@ -438,6 +492,221 @@ def getTile(self, x, y, z, pilImageAllowed=False, numpyAllowed=False, **kwargs):
...
+                mask = mask[:, :, np.newaxis]
+        if 'x' not in axes or 'y' not in axes:
+            err = 'Invalid value for axes. Must contain "y" and "x".'
+            raise ValueError(err)
+        for k in placement:
+            if k not in axes:
+                axes[0:0] = [k]
+        self._axes = {k: i for i, k in enumerate(axes)}
+        while len(tile.shape) < len(axes):
+            tile = np.expand_dims(tile, axis=0)
```
manthey, line 540: It occurred to me this could be done without an explicit while loop.  I think the equivalent would be:
```
if len(tile.shape) < len(axes):
    tile = np.expand_dims(tile, axis=tuple(range(len(axes) - len(tile.shape))))
if mask is not None and len(mask.shape) < len(axes):
    mask = np.expand_dims(mask, axis=tuple(range(len(axes) - len(mask.shape))))
```
I'm not sure that is clearer, though.

sources/zarr/large_image_source_zarr/__init__.py
annehaley: Changed in [32c9c58](https://github.com/girder/large_image/pull/1446/commits/32c9c5819c60276749f3beebf0c5d36ef848eeab)

sources/zarr/large_image_source_zarr/__init__.py
```diff
@@ -79,7 +90,47 @@ def __init__(self, path, **kwargs):
...
+        self.sizeX = self.sizeY = self.levels = 0
+        self.tileWidth = self.tileHeight = self._tileSize
+        self._cacheValue = str(uuid.uuid4())
+        self._output = None
+        self._editable = True
+        self._bandRanges = None
+        self._addLock = threading.RLock()
+        self._framecount = 0
+        self._mm_x = 0
+        self._mm_y = 0
```
manthey, line 114: We should probably add `self._levels = []` here and a test that just does
```
assert sink.metadata['levels'] == 0
assert sink.getRegion(format=TILE_FORMAT_NUMPY)[0].shape[:2] == (0, 0)
```

sources/zarr/large_image_source_zarr/__init__.py
```diff
@@ -438,6 +496,220 @@ def getTile(self, x, y, z, pilImageAllowed=False, numpyAllowed=False, **kwargs):
...
+        if not isinstance(tile, np.ndarray) or axes is None:
+            axes = 'yxs'
+            tile, mode = _imageToNumpy(tile)
+        elif not isinstance(axes, str) and not isinstance(axes, list):
+            err = 'Invalid type for axes. Must be str or list[str].'
+            raise ValueError(err)
+        axes = [x.lower() for x in axes]
+        if axes[-1] != 's':
+            axes.append('s')
+            if mask is not None and len(axes) - 1 == len(mask.shape):
```
manthey, line 534: This needs to be outdented (not in the `if axes[-1]` clause).

manthey (changes requested):

manthey:
Let's add a mask test:
```
def testAddTileWithMask():
    tile0 = np.random.random((10, 10))
    sink.addTile(tile0, 0, 0)
    orig = sink.getRegion(format='numpy')[0]
    tile1 = np.random.random((10, 10))
    sink.addTile(tile1, 0, 0, mask=np.random.random((10, 10)) > 0.5)
    cur = sink.getRegion(format='numpy')[0]
    assert (tile0 == orig[:, :, 0]).all()
    assert not (tile1 == orig[:, :, 0]).all()
    assert not (tile0 == cur[:, :, 0]).all()
    assert not (tile1 == cur[:, :, 0]).all()
```

This will fail because of caching.  In the vips source we have
https://github.com/girder/large_image/blob/master/sources/vips/large_image_source_vips/__init__.py#L125

and do `self._cacheValue = str(uuid.uuid4())` as part of `_initNew` and when we invalidate the image (when we set `self._levels = None`).

manthey (approved):

PR #1448, opened 2024-01-29
Infer DICOM file size, when possible

psavery:
Google Healthcare API supports single-part requests for DICOM files. Single-part requests are not officially supported by the DICOMweb standard, but they are easier to work with. Google Healthcare API also provides the Content-Length. We can utilize this information to infer the DICOM file size without having to stream the whole file.

This PR adds logic to automatically infer the DICOM file size and set it, when possible, without streaming the whole file.

Unfortunately, dcm4chee does not support single-part requests, and they do not return a Content-Length either. So we must still stream the whole file to get the file size for dcm4chee servers. But if we determine a new way to infer the file size, we can add it as well.


manthey (approved):

psavery:
In a follow-up PR, I will store whether the server supports singlepart requests in the assetstore.

PR #1450, opened 2024-01-31
Add support for Imaging Data Commons

psavery:
The NCI's [Imaging Data Commons](https://datacommons.cancer.gov/repository/imaging-data-commons) is a big repository (>38k studies) for cancer research. For the DICOMweb server, it uses Google's Cloud Healthcare API behind a proxy. This DICOMweb server sometimes behaves differently than the dcm4chee server we have been testing with.

This PR fixes a couple of issues we encountered. One of which is that we cannot use the `SOPClassUID` as a search filter (even though the [DICOMweb standard](https://dicom.nema.org/medical/dicom/current/output/chtml/part18/sect_10.6.html) indicates that it should be supported when searching for instances). We can perform manual filtering instead, however.

This PR also adds a default limit to the import page (which is required for importing from IDC), and a default search filter that specifies `SM`. Without the search filter, we end up importing a lot of non-WSI datasets. With the `SM` search filter, most of the imported datasets look correct, and all are viewable.

Also needed to support the IDC: imi-bigpicture/wsidicom#149

### DIFF
--- a/sources/dicom/large_image_source_dicom/assetstore/dicomweb_assetstore_adapter.py
+++ b/sources/dicom/large_image_source_dicom/assetstore/dicomweb_assetstore_adapter.py
@@ -438,7 +438,6 @@ class DICOMwebAssetstoreAdapter(AbstractAssetstoreAdapter):
         if progress:
             progress.update(message='Searching for series...')
 
-        # FIXME: might need to search in chunks for larger web servers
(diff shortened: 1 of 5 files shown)


psavery:
https://github.com/imi-bigpicture/wsidicom/pull/149 was merged, so this is definitely ready!

manthey (approved):

PR #1469, opened 2024-02-25
Add redis as cache backend option

nipeone:


.circleci/config.yml
```diff
@@ -103,6 +103,10 @@ commands:
           name: start memcached
           command: |
             docker run --rm -d -p 11211:11211 memcached -m 64
+      - run:
+          name: start redis
+          command: |
+            docker run --rm -d -p 6379:6379 redis -m 64
```
manthey, line 109: I think CI will pass if you remove the `-m 64` from the end of this line.

manthey:
This is great.  With this, to pass tests locally one needs memcached and redis running.  I'm thinking that after this is merged, I'll add some environment checks so that the memcached and redis tests are only run if the appropriate service exists.

.circleci/config.yml
nipeone: I'm very sorry it's taken so long to deal with this issue, I'll fix it and resubmit the PR

manthey:
Thanks for working on this.  You should be able to fix the linting errors by doing `tox -e format`.

nipeone:
i get this error like async call, but i'm sure i used sync function of redis. due to i'm not very familiar with CI.  maybe i need some help.
```
large_image/cache_util/rediscache.py:60: error: Argument 1 to "len" has incompatible type "Union[Awaitable[Any], Any]"; expected "Sized"  [arg-type]
large_image/cache_util/rediscache.py:65: error: Incompatible return value type (got "Union[Awaitable[Any], Any]", expected "bool")  [return-value]
large_image/cache_util/rediscache.py:79: error: Argument 1 to "loads" has incompatible type "Union[Awaitable[Any], Any]"; expected "Buffer"  [arg-type]
large_image/cache_util/rediscache.py:114: error: Incompatible return value type (got "Union[Awaitable[Any], Any]", expected "int")  [return-value]
large_image/cache_util/rediscache.py:143: error: Value of type "Union[Awaitable[Any], Any]" is not indexable  [index]
large_image/cache_util/rediscache.py:151: error: List or tuple expected as variadic arguments  [misc]
```

manthey:
I think redis doesn't type response from the client aside from a very general type, so I'd use some judicious casts.  From your current checkin, I think the following diff would clean up the type and lint issues:
[diff.txt](https://github.com/girder/large_image/files/15420579/diff.txt)

manthey (approved):

manthey:
@nipeone Thank you so much for adding this.  I'll make a follow on PR to make it so the redis tests are run conditionally to make it easier to test locally if redis isn't available.

PR #1488, opened 2024-03-26
Collect and report implicit tile source extensions and mime types

manthey:


manthey:
See https://girder.github.io/large_image/formats.html

PR #1490, opened 2024-03-29
Zarr Tile Sink: Generate Downsampled Levels

### LINKED ISSUE #1466: Tile sink output options (lower resolution tiles)
This could be considered details related to #1002.

When we write an image, we want to be able to control how downsampling is done for lower resolution layers.  In vips, we have (https://www.libvips.org/API/current/VipsRegion.html)
```
VIPS_REGION_SHRINK_MEAN | use the average 
VIPS_REGION_SHRINK_MEDIAN | use the median 
VIPS_REGION_SHRINK_MODE | use the mode 
VIPS_REGION_SHRINK_MAX | use the maximum 
VIPS_REGION_SHRINK_MIN | use the minimum   
VIPS_REGION_SHRINK_NEAREST | use the top-left pixel 
```
and, in PIL we have https://pillow.readthedocs.io/en/stable/reference/Image.html#resampling-filters.  PIL properly does this (see https://zuru.tech/blog/the-dangers-behind-image-resizing).

We want to add an optional kwarg: `resample` that would an enum that would include the PIL Resample modes and the vips options.  For any of the PIL modes, we should just use PIL for downsampling.  For other modes, we probably need to implement this with numpy.

Also, when downsampling, to reduce tile edge effects caused by filtering (e.g., when using Lanczos or hamming), we probably want to pick a larger area, downsample that, and crop.

annehaley:
Resolves #1466 

This PR adds a new file to `large_image.tilesource` called `resample.py`. This new file contains an enum called `ResampleMethod` and the implementations for using each resample method to downsample a tile to half resolution. This file includes a function `downsampleTileHalfRes` to perform this downsampling given a tile and a resample method.

This PR updates the `ZarrFileTileSource` to use the `downsampleTileHalfRes` function in a new class function `_generateDownsampledLevels`. These updates include changes to `addTile` and the cropping method used in `write`.

This PR also adds tests to confirm the shapes of the generated downsampled levels.


large_image/tilesource/resample.py
```diff
@@ -0,0 +1,124 @@
...
+    tile: np.ndarray,
+    new_shape: Dict,
+    resample_method: ResampleMethod,
+) -> np.ndarray:
+    if resample_method == ResampleMethod.NP_NEAREST:
+        return tile[::2, ::2]
+    else:
+        pixel_selection = None
+        subarrays = np.asarray(
+            [
```
manthey, line 51: What happens if the tile's shape is not a multiple of 2?

manthey: We probably could replicate the last row and/or column in those instances.

large_image/tilesource/resample.py
```diff
@@ -0,0 +1,124 @@
...
+                (
+                    (subarrays[1] == subarrays[2]).all(axis=2) |
+                    (subarrays[1] == subarrays[3]).all(axis=2)
+                ),
+                1, np.where(
+                    (subarrays[2] == subarrays[3]).all(axis=2),
+                    2, 0,
+                ),
+            )
+        elif resample_method == ResampleMethod.NP_MAX:
```
manthey, line 76: I think I was expecting NP_MAX to to be the max of each band independently and NP_MAX_CROSSBAND to be one of the existing pixels.  By my expectation, these are reversed.

large_image/tilesource/resample.py
```diff
@@ -0,0 +1,124 @@
...
+            msg = f'Unknown resample method {resample_method}.'
+            raise ValueError(msg)
+
+
+def downsampleTileHalfRes(
+    tile: np.ndarray,
+    resample_method: ResampleMethod,
+) -> np.ndarray:
+    new_shape = {
+        'height': int(tile.shape[0] / 2),
```
manthey, line 102: Why not use `//` ?
And, I think we want to round up (e.g., `(tile.shape[0] + 1) // 2`).

large_image/tilesource/resample.py
annehaley: Oh ok, maybe we can name these better, because I thought "crossband" would mean that the result is a combination of the maxs from each band (and may not be an existing color). 
It does make sense to call the max of each band independently NP_MAX, since it is the simpler version. Maybe we can call the color-preserving version NP_MAX_COLOR or something similar?

sources/zarr/large_image_source_zarr/__init__.py
```diff
@@ -659,12 +677,70 @@ def crop(self, value):
...
+        current_arrays = dict(self._zarr.arrays())
+        if '0' not in current_arrays:
+            msg = 'No root data found, cannot generate lower resolution levels.'
+            raise TileSourceError(msg)
+        if 'x' not in self._axes or 'y' not in self._axes:
+            msg = 'Data must have an X axis and Y axis to generate lower resolution levels.'
+            raise TileSourceError(msg)
+
+        metadata = self.getMetadata()
+        tile_size = dict(width=4096, height=4096)
```
manthey, line 691: For efficiency reasons, we might want this to be 4096+(overlap size).  More on overlap below.

sources/zarr/large_image_source_zarr/__init__.py
```diff
@@ -659,12 +677,70 @@ def crop(self, value):
...
+            )
+            for frame in metadata.get('frames', [{'Index': 0}]):
+                frame_position = {
+                    k.replace('Index', '').lower(): v
+                    for k, v in frame.items()
+                    if k.replace('Index', '').lower() in self._axes
+                }
+                for tile in self.tileIterator(
+                    tile_size=tile_size,
+                    tile_overlap=dict(x=4, y=4),
```
manthey, line 706: I would use `tile_overlap=dict(x=4, y=4, edges=True)`.

sources/zarr/large_image_source_zarr/__init__.py
```diff
@@ -659,12 +677,70 @@ def crop(self, value):
...
+                    resample=False,  # TODO: incorporate resampling in core
+                ):
+                    new_tile = downsampleTileHalfRes(tile['tile'], resample_method)
+                    overlap = {k: int(v / 2) for k, v in tile['tile_overlap'].items()}
+                    new_tile = new_tile[
+                        slice(overlap['top'], new_tile.shape[0] - overlap['bottom']),
+                        slice(overlap['left'], new_tile.shape[1] - overlap['right']),
+                    ]
+
+                    x = int(
```
manthey, line 718: You should use `(tile['x'] + tile['overlap']['left']) / 2`.

Run the samples with a tile size that is smaller than the images to make everything looks right -- the tile size includes the overlap (see https://github.com/girder/large_image/blob/master/large_image/tilesource/base.py#L2258-L2272)

large_image/tilesource/resample.py
annehaley: Ok, changed in a472576

large_image/tilesource/resample.py
manthey: > Oh ok, maybe we can name these better, because I thought "crossband" would mean that the result is a combination of the maxs from each band (and may not be an existing color). It does make sense to call the max of each band independently NP_MAX, since it is the simpler version. Maybe we can call the color-preserving version NP_MAX_COLOR or something similar?

Other libraries have MAX / MIN and mean the max / min of each band, so we want the simple name to be that.  Do you think NP_MAX_SAMPLE is clearer than NP_MAX_COLOR (pick whichever you like better).

sources/zarr/large_image_source_zarr/__init__.py
manthey: As a petty refinement, if the downsampling technique is one of our NP versions or PIL_NEAREST_NEIGHBOR, the overlap doesn't gain any benefit (since there won't be edge effects as in the interpolated modes), so we could have a conditional to either use 4 or 0 depending (and then I'd chose that first and then make tile_size be 4096 plus the chosen overlap).

sources/zarr/large_image_source_zarr/__init__.py
annehaley: Thanks, changed in [9a74c0a](https://github.com/girder/large_image/pull/1490/commits/9a74c0ac85f7c6394c25d0a29e395f0c371f337f)

sources/zarr/large_image_source_zarr/__init__.py
annehaley: Good idea. I added this conditional in [9a74c0a](https://github.com/girder/large_image/pull/1490/commits/9a74c0ac85f7c6394c25d0a29e395f0c371f337f)

sources/zarr/large_image_source_zarr/__init__.py
annehaley: Changed in [9a74c0a](https://github.com/girder/large_image/pull/1490/commits/9a74c0ac85f7c6394c25d0a29e395f0c371f337f)

sources/zarr/large_image_source_zarr/__init__.py
```diff
@@ -659,12 +677,75 @@ def crop(self, value):
...
+                    resample=False,  # TODO: incorporate resampling in core
+                ):
+                    new_tile = downsampleTileHalfRes(tile['tile'], resample_method)
+                    overlap = {k: int(v / 2) for k, v in tile['tile_overlap'].items()}
+                    new_tile = new_tile[
+                        slice(overlap['top'], new_tile.shape[0] - overlap['bottom']),
+                        slice(overlap['left'], new_tile.shape[1] - overlap['right']),
+                    ]
+
+                    x = int((tile['x'] + overlap['left']) / 2)
```
manthey, line 729: This should either be `(tile['x'] + tile['overlap']['left']) / 2` or `tile['x'] / 2 + overlap['left']`

large_image/tilesource/resample.py
annehaley: We've discovered that "sample" can mean a lot of things, and I feel that "color" implies the combination of bands while "sample" could imply a single band. I think the version that uses the sums to compare pixels is intended to preserve color. I'm leaning towards using `NP_MAX_COLOR` unless you have a preference.

sources/zarr/large_image_source_zarr/__init__.py
annehaley: I see, I forgot `overlap` was already halved. Fixed in [e7234b2](https://github.com/girder/large_image/pull/1490/commits/e7234b2e1a6dc310959c1fbaef8ee8d91a9377ee)

large_image/tilesource/resample.py
annehaley: Added in [dbcb661](https://github.com/girder/large_image/pull/1490/commits/dbcb661c038b87058b7d2975d5952b43b38b46d3)

large_image/tilesource/resample.py
annehaley: Renamed in [ff67307](https://github.com/girder/large_image/pull/1490/commits/ff67307955fbf421b3ab44357dfb5dc1721e25f8)

sources/zarr/large_image_source_zarr/__init__.py
```diff
@@ -684,8 +765,14 @@ def write(
...
-        # TODO: compute half, quarter, etc. resolutions
-        self._validateZarr()
+        if resample is None:
+            resample = (
+                ResampleMethod.PIL_LANCZOS
+                if lossy and self.dtype == np.uint8
+                else ResampleMethod.NP_NEAREST
+            )
+
+        self._generateDownsampledLevels(resample)
```
manthey, line 775: We probably want to do this after crop.

We also don't need to do it if we are saving via the large_image_converter, since it does its own downsampling.

annehaley: In order to generate downsampled levels after cropping, we will need to change the way we do cropping.

Currently, cropped data is kept in a separate zarr store which we create at the time of `write`. If we generate downsampled levels after this crop, we would need to propagate the cropped store to the `generateDownsampledLevels` and `addTile` methods as the target for changes. This will also change how we access `self.getMetadata()` for frames, `self.levels`, `self.sizeX`, `self.sizeY`, etc. in these methods. Since this approach would complicate other methods, I figured it would be better to change how we crop. 

Instead of managing a temporary zarr store directly, we can use a new instance of the `ZarrFileTileSource`. We can copy tiles from the original to the new sink in a similar fashion to `generateDownsampledLevels`, with a `tileIterator`. We can use the `region` arg to perform the crop. Then we can call `generateDownsampledLevels` on this instance if the output type doesn't call for the large_image_converter.

I made these changes and added a new test for this behavior in [4ef0c58](https://github.com/girder/large_image/pull/1490/commits/4ef0c584f47119f2e7c4f350255a8361430b82a0).

test/test_sink.py
```diff
@@ -176,3 +198,63 @@ def testImageCopySmallMultiband(file_type, tmp_path):
...
+    assert written_arrays.get('2') is not None
+    assert written_arrays.get('2').shape == (2, 3, 1024, 512, 3)
+    assert written_arrays.get('3') is not None
+    assert written_arrays.get('3').shape == (2, 3, 512, 256, 3)
+
+    # TODO: Can the content of the downsampled data be compared back to the test source?
+
+
+@pytest.mark.parametrize('resample_method', RESAMPLE_METHODS)
+def testImageCopyLargeDownsamplingMultiband(resample_method, tmp_path):
```
manthey, line 234: These tests are very slow compared to the overall CI time.  There are two different things we are trying to test: that downsampling methods are correct, and that our multi-downsampled levels are correct.  We aren't currently testing all of the second, because the size of these tests aren't big enough to see if we have the overlap effects right.

Probably we should keep these tests, but do them on small samples (just big enough to need downsampling).  Because the fractal is regular, for the numpy tests we can check if we get the expected answer in one or two spots.  The multiband tests could also just be done on a subset of downsampling methods (lanczos and np_nearest_neighbor, maybe) just to test the different code paths.

As a test of the effects of overlap we want to test at just lanczos and np_nearest_neighbor on something more than twice as big as the operating tile size (the 4096 we used), and check (at least visually) that it looks right.  In this case, I probably would only test the single band variant.

The previous test will also test the recursion on the downsampling.

annehaley: Ok, I can decrease the image size of these tests. I will also reduce the set of resample methods sent through the 6-band test.  I will need your help to write the correct assertion for the content of the downsampled levels.
Am I understanding correctly that a new test should use a larger size (but only one band) to test the overlap effects on two resample methods? And this will be the only test that has more than 2 levels?

annehaley: Also, since we are using the small image size for the main downsampling tests, they can be combined with `testImageCopySmall` and `testImageCopySmallMultiband`. I can just move the assertions.

manthey: That is what I was thinking; I'm trying to balance code coverage and test time.  CI currently takes 20+ minutes on master and is currently failing on this branch because it sometimes exceeds the hour limit.

annehaley: Reorganized the tests in [f7876a3](https://github.com/girder/large_image/pull/1490/commits/f7876a3e4512dc3c2349536a88a5204dca91f13d), including the new assertion for the content of the downsampled levels depending on the resample method.

test/test_sink.py
```diff
@@ -158,21 +183,128 @@ def testImageCopySmallMultiband(file_type, tmp_path):
...
+@pytest.mark.parametrize('resample_method', list(ResampleMethod))
+def testImageCopySmallDownsampling(resample_method, tmp_path):
+    output_file = tmp_path / f'test_{resample_method}.db'
+    sink = large_image_source_zarr.new()
+    source = large_image_source_test.TestTileSource(
+        fractal=True,
+        tileWidth=128,
+        tileHeight=128,
+        sizeX=512,
+        sizeY=1024,
```
manthey, line 219: We have pretty good coverage, but aren't exercising the odd-sized-tiles code.  We could use odd dimensions to exercise it; making the sizes slightly smaller would make the tests take no more time than at present; there is no reason they need to be powers of 2.  Obviously the asserts would have to change, too.

manthey (approved):
I made one comment about testing odd sized tiles, but otherwise this looks good.

test/test_sink.py
annehaley: Ok, I changed the dimension sizes by 1 in [2e10ec4](https://github.com/girder/large_image/pull/1490/commits/2e10ec443da02cb4f454a64867090f71034826cd)

manthey:
> I made one comment about testing odd sized tiles, but otherwise this looks good.

Great.  Merge away.

PR #1508, opened 2024-04-17
Zarr sink metadata management

### LINKED ISSUE #1482: In zarr sink, set mm_x, mm_y
There are stored in the `datasets` `coordinateTransformations` `scale` metadata.

### LINKED ISSUE #1483: In zarr sink, set image description

### LINKED ISSUE #1484: In zarr sink, set channel names
Each value along the `c` axis can have a name (e.g., `DAPI`, `Red`)

### LINKED ISSUE #1485: In zarr sink, support adding associated images
Associated images are small enough they are not tiled.

annehaley:
- [x] Resolves #1482
- [x] Resolves #1483 
- [x] Resolves #1484 
- [x] Resolves #1485


sources/zarr/large_image_source_zarr/__init__.py
```diff
@@ -653,6 +648,111 @@ def addTile(self, tile, x=0, y=0, mask=None, axes=None, **kwargs):
                 self.levels = int(max(1, math.ceil(math.log(max(
                     self.sizeX / self.tileWidth, self.sizeY / self.tileHeight)) / math.log(2)) + 1))
 
+    def addAssociatedImage(self, image):
```
manthey, line 651: We should change this method signature to `def addAssociatedImage(self, image, imageKey)`, where imageKey is a string that we store and can use to identify what this associated image is used for.

manthey: Note that you couldn't set the imageKey in the vips sink (we could go back and add it eventually).

manthey: We don't currently read in the imageKey when finding associated images in zarr files.  If we used the key as the zarr group OR used associated_<num>, then we could use the group name on read (this means changing https://github.com/girder/large_image/blob/master/sources/zarr/large_image_source_zarr/__init__.py#L323-L324 and associated areas).

annehaley: Ok, I made this change in [8976f3d](https://github.com/girder/large_image/pull/1508/commits/8976f3df1790217e1b45ca6466907488ff7c3cc0). We now store `self._associatedImages` as a dict.

manthey (approved):

PR #1547, opened 2024-06-14
Tifffile _biggestSeries() bug for large images

Dany546:
The _biggestSeries() method used to read tifffile sources rely on np.prod(shape) to find the biggest series in the file (macro image > image). This leads to an overflow for big images and the macro image is wrongly selected as the main content of the file. Simple fix is done with np.sqrt(), though a better solution could be found.


manthey:
Do you have an example image where it fails?  If it can't be shared, can the output of `tifftools dump <image>` be shared?

I don't understand why the sqrt of the shape product would be desirable.  The intent is that the main series would be the series with the most total pixels.  If it is overflowing do to data types, a better choice might be to `samples = np.prod(float(x) for x in s.shape)` to cast the values to float values before multiplication.

PR #1551, opened 2024-06-18
Zarr Sink Multiprocessing

annehaley:
In addition to the zarr sink being compatible with multithreading, this PR makes the zarr sink compatible with multiprocessing. This is accomplished by the following:

-  splitting the `addLock` into `threadLock` and `processLock` (each a different class)
-  changing how a zarr sink is initialized from a path; now made more predictable by using `os.tempdir` as the root rather than a generated `tempdir.TemporaryDirectory()` and opened with "append" mode via zarr
- removing the clause to set `self.unpickleable = True`
- adding a pytest to check the behavior of both multithreading and multiprocessing cases

With this addition, there continues to be the caveat that the size of the whole image must be known before any multiprocessing or multithreading can work. The image size cannot be changed within a thread or process. The test reflects this; the last tile in the tileset (at the maximum extents) is added first, then all other tiles are added within a thread or process.

### DIFF
--- a/sources/zarr/large_image_source_zarr/__init__.py
+++ b/sources/zarr/large_image_source_zarr/__init__.py
@@ -3,6 +3,7 @@ import os
 import shutil
 import tempfile
 import threading
+import multiprocessing
 import uuid
 import warnings
 from importlib.metadata import PackageNotFoundError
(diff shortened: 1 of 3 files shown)


manthey (approved):
Excellent.  @annehaley I know you are working on it, but the next step is making sure we document how to use this.  I'd like to see an example either in the example_usage document (or somewhere similar) or as a jupyter notebook example.

PR #1557, opened 2024-06-25
Zarr Sink Channel Metadata Fix

### LINKED ISSUE #1554: Zarr sink channels and samples issue
This code:
```
import large_image, numpy as np

s = large_image.new()
for c in range(5):
    s.addTile(np.random.random((4, 4, 3)), c=c)

s.channelNames = ['a', 'b', 'c', 'd', 'e']
s.write('example.zip')
t = large_image.open('example.zip')
assert len(t.metadata['channels'] ) == 5
```
fails.  The channel names aren't written because of the code on https://github.com/girder/large_image/blob/master/sources/zarr/large_image_source_zarr/__init__.py#L677
which should probably be `channel_axis = self._axes.get('c', self._axes.get('s'))`, and then https://github.com/girder/large_image/blob/master/sources/zarr/large_image_source_zarr/__init__.py#L708 should check that `channel_axis` is not None rather than not falsy.  However, these changes then causes the code in https://github.com/girder/large_image/blob/master/sources/zarr/large_image_source_zarr/__init__.py#L712-L720 to be incorrect as bands (samples) and channels are conflated.

annehaley:
Resolves #1554

### DIFF
--- a/sources/zarr/large_image_source_zarr/__init__.py
+++ b/sources/zarr/large_image_source_zarr/__init__.py
@@ -674,7 +674,7 @@ class ZarrFileTileSource(FileTileSource, metaclass=LruCacheMetaclass):
         with self._addLock:
             name = str(self._tempdir.name).split('/')[-1]
             arrays = dict(self._zarr.arrays())
-            channel_axis = self._axes.get('s') or self._axes.get('c')
+            channel_axis = self._axes.get('c', self._axes.get('s'))
             datasets = []
             axes = []
             channels = []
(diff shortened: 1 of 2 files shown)


sources/zarr/large_image_source_zarr/__init__.py
```diff
@@ -717,7 +717,13 @@ def _writeInternalMetadata(self):
                         'inverted': False,
                         'label': f'Band {c + 1}',
                     }
-                    channel_data = base_array[..., c]
+                    slicing = tuple(
+                        slice(None)
+                        if k != ('C' if 'C' in self._axes else 'S')
```
manthey, line 722: On line 677, self._axes is lowercase; here we are checking upper case.  I think this should be lowercase.

annehaley: Good point. Changed in [5e33da7](https://github.com/girder/large_image/pull/1557/commits/5e33da7ab9ab71ceb3e502208809f5e18744a9b6).

manthey (approved):

Issue #1139, opened 2023-05-02
Add an endpoint to recheck an entire folder for large images


manthey:
This would check if any files could be used as large images that aren't marked as such, and would check if existing large_images are still readable, and, if not, unmark and try to remark them.  It should have a recurse option.

manthey:
This is related to #873.

manthey:
I think the basic process would be to iterate through all the items in a folder or folder tree.  Pseudocode:
```
from girder_large_image.models.image_item import ImageItem

for item in Folder().childItems():
    if item.get('largeImage'):
        if item['largeImage']['expected']:
            # we asked to create a large image and it hasn't finished.  Either the process is in the works or failed;
            # maybe a flag on the endpoint would optionally `ImageItem().delete(item)` in this case and then fall through
            # to the condition is largeImage isn't present in item.
            pass
        else:
            try:
                ItemItem().getMetadata(item)
                # all well, continue
                continue
            except Exception:
                 # we failed to open the item and get its image metadata, therefore:
                 previousFileId = item['largeImage']['fileId']
                 ImageItem().delete(item)
                 # also allow creating jobs based on an option
                 ImageItem().createImageItem(item, File().load(previousFileId))
    else:
        # item was never a large image, can we ask to make sure it can't be?
        # get file id the same way as the createTiles method in the girder large_image rest endpoint
        # also allow creating jobs based on an option
        ImageItem().createImageItem(item, File().load(previousFileId))
```

Options we could want for this endpoint (this can be multiple PRs):
- [ ] recurse the specified folder
- [ ] delete creation jobs (default False)
- [ ] use creation jobs if we can't read files as images directly (default False)
- [ ] optional criteria to force undoing the largeImage and redoing it (the two examples I have are a regex for the item name or a source name)
- [ ] always use a job when creating the largeImage (the same action as when we ctrl-click the large image button in the UI)

manthey:
There are a few motivations for this endpoint.:
- It is possible that the format support was added or improved since the items were uploaded
- Previous versions might have picked a poor source because of a partial upload.  Example, mrxs files have a folder of data and an item that can be interpreted as a JPEG.  If the item is selected as a large image before the folder is fully uploaded, we read it with the pil source as a jpeg; redoing it will read it with the openslide source as a mrxs file (and have much more detail).
- Some files may have gone offline; unmarking them reduces the attempt to use them (this is probably indicative of bad asset management, but that should be addressed differently).
[...]

PR #1568, opened 2024-07-09
Create large images for all items in a folder

gabrielle6249:
This adds an endpoint that creates large images for an entire folder, as mentioned [here](https://github.com/girder/large_image/issues/1139). It iterates through the folder's items and creates a large image item for each if it isn't one already, with the option to recurse child folders. It also has an option to create a job for each large image item, and another to make the jobs local.

This does not yet finish creating large images for items with a large image in progress, but seemed to be a good start.


girder/girder_large_image/rest/__init__.py
```diff
@@ -17,6 +19,7 @@ def addSystemEndpoints(apiRoot):
     """
     apiRoot.folder.route('GET', (':id', 'yaml_config', ':name'), getYAMLConfigFile)
     apiRoot.folder.route('PUT', (':id', 'yaml_config', ':name'), putYAMLConfigFile)
+    apiRoot.folder.route('PUT', (':id', 'items'), createLargeImages)
```
manthey, line 22: This adds the endpoint `folder/:id/items`.  Other endpoints that are explicitly for large_images are either in `item/:id/tiles/...` or `large_image/...`.  I think, conceptually, this is a utility function so it would make more sense to move this to rest/large_image_resource.py endpoint and rename the route to `('folder', ':id', 'create')` (though I could also see using `tiles` rather than `create` for consistency).

gabrielle6249: That route and location make more sense. Function moved and route renamed [here](https://github.com/girder/large_image/commit/588d03056eeedf6c4f06037d81e360a769e41140).

gabrielle6249:
Ran into some trouble on this branch, so I created a new [branch](https://github.com/girder/large_image/tree/create-large-images) & [PR](https://github.com/girder/large_image/pull/1572).

PR #1572, opened 2024-07-11
Create new large images for each item in a folder

gabrielle6249:
This adds an endpoint that creates large images for an entire folder, as mentioned [here](https://github.com/girder/large_image/issues/1139). It iterates through the folder's items, skips over items with a large image already and creates a large image for the items without one. It has several options, one that recurses child folders, a second that creates a job for each large image item, and another that makes the jobs local.

This does not yet retry creating large images for items with an unfinished large image, but seemed to be a good start.

### DIFF
--- a/girder/girder_large_image/rest/large_image_resource.py
+++ b/girder/girder_large_image/rest/large_image_resource.py
@@ -42,7 +42,7 @@ from girder.models.folder import Folder
 from girder.models.item import Item
 from girder.models.setting import Setting
 from large_image import cache_util
-from large_image.exceptions import TileGeneralError
+from large_image.exceptions import TileGeneralError, TileSourceError
 
 from .. import constants, girder_tilesource
 from ..models.image_item import ImageItem
@@ -450,8 +450,8 @@ class LargeImageResource(Resource):
     @access.user(scope=TokenScope.DATA_WRITE)
     @autoDescribeRoute(
         Description('Create new large images for all items within a folder.')
-        .notes('Does not work for items with multiple files and removes existing '
-               'large images from items.')
+        .notes('Does not work for items with multiple files and skips over items with '
+               'existing or unfinished large images.')
         .modelParam('id', 'The ID of the folder.', model=Folder, level=AccessType.WRITE,
                     required=True)
         .param('createJobs', 'Whether job(s) should be created for each large image.',
@@ -474,6 +474,7 @@ class LargeImageResource(Resource):
 
     def createImagesRecurseOption(self, folder, createJobs, user, recurse, localJobs):
         result = {'childFoldersRecursed': 0,
+                  'itemsSkipped': 0,
                   'largeImagesCreated': 0,
                   'largeImagesRemovedAndRecreated': 0,
                   'totalItems': 0}
@@ -489,14 +490,15 @@ class LargeImageResource(Resource):
             result['totalItems'] += 1
             if item.get('largeImage'):
                 if item['largeImage'].get('expected'):
-                    pass
+                    result['itemsSkipped'] += 1
                 else:
                     try:
-                        Item().getMetadata(item)
+                        ImageItem().getMetadata(item)
+                        result['itemsSkipped'] += 1
                         continue
-                    except Exception:
-                        previousFileId = item['largeImage'].get(
-                            'originalId', item['largeImage']['fileId'])
+                    except TileSourceError:
+                        previousFileId = item['largeImage'].get('originalId',
+                                                                item['largeImage']['fileId'])
                         ImageItem().delete(item)
                         ImageItem().createImageItem(item, File().load(user=user, id=previousFileId),
                                                     createJob=createJobs, localJob=localJobs)
@@ -507,6 +509,8 @@ class LargeImageResource(Resource):
                     ImageItem().createImageItem(item, files[0], createJob=createJobs,
                                                 localJob=localJobs)
                     result['largeImagesCreated'] += 1
+                else:
+                    result['itemsSkipped'] += 1
         return result
 
     @describeRoute(
--- a/girder/girder_large_image/rest/large_image_resource.py
+++ b/girder/girder_large_image/rest/large_image_resource.py
@@ -449,12 +449,12 @@ class LargeImageResource(Resource):
 
     @access.user(scope=TokenScope.DATA_WRITE)
     @autoDescribeRoute(
-        Description('Create large images for all items within a folder.')
-        .notes('Does not work for items with multiple files.')
+        Description('Create new large images for all items within a folder.')
+        .notes('Does not work for items with multiple files and removes existing large images from items.')
         .modelParam('id', 'The ID of the folder.', model=Folder, level=AccessType.WRITE, required=True)
-        .param('createJobs','Whether a job should be created for each large image.', required=False, 
+        .param('createJobs','Whether job(s) should be created for each large image.', required=False, 
             default=False, dataType='boolean')
-        .param('localJobs', 'Whether a the jobs created should be local.', required=False, 
+        .param('localJobs', 'Whether the job(s) created should be local.', required=False, 
             default=False, dataType='boolean')
         .param('recurse', 'Whether child folders should be recursed', required=False, default=False, 
             dataType='boolean')
@@ -466,16 +466,22 @@ class LargeImageResource(Resource):
         createJobs = params.get('createJobs')
         if createJobs:
             createJobs = 'always' or True
-        self.createImagesRecurseOption(folder=folder, createJobs=createJobs, user=user, 
+        return self.createImagesRecurseOption(folder=folder, createJobs=createJobs, user=user, 
                                 recurse=params.get('recurse'), localJobs=params.get('localJobs'))
                     
     def createImagesRecurseOption(self, folder, createJobs, user, recurse, localJobs):
-        count = {'large images created': 0}
+        result = {'childFoldersRecursed': 0,
+                  'largeImagesCreated': 0,
+                  'largeImagesRemovedAndRecreated': 0,
+                  'totalItems': 0}
         if recurse:
             for childFolder in Folder().childFolders(parent=folder, parentType='folder'):
-                self.createImagesRecurseOption(folder=childFolder, createJobs=createJobs, user=user, 
+                result['childFoldersRecursed']+=1
+                childResult = self.createImagesRecurseOption(folder=childFolder, createJobs=createJobs, user=user, 
                                         recurse=recurse, localJobs=localJobs)
+                for key in childResult: result[key]+= childResult[key]
         for item in Folder().childItems(folder=folder):
+            result['totalItems'] +=1
             if item.get('largeImage'):
                 if item['largeImage'].get('expected'):
                     pass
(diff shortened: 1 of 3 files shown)


girder/girder_large_image/rest/large_image_resource.py
```diff
@@ -445,6 +447,55 @@ def _deleteCachedImages(self, spec, associatedImages=False, imageKey=None):
...
+                        Item().getMetadata(item)
+                        continue 
+                    except Exception:
+                        previousFileId = item['largeImage'].get('originalId', item['largeImage']['fileId'])
+                        ImageItem().delete(item)
+                        ImageItem().createImageItem(item, File().load(user=user, id=previousFileId), 
+                                                    createJob=createJobs, localJob=localJobs)
+            else:
+                largeImageFileId = None
+                files = list(Item().childFiles(item=item, limit=0))
```
annehaley, line 493: Did you ever find out if we can document the behavior of `limit=0` somewhere?

girder/girder_large_image/rest/large_image_resource.py
```diff
@@ -445,6 +447,55 @@ def _deleteCachedImages(self, spec, associatedImages=False, imageKey=None):
...
+        self.createImagesRecurseOption(folder=folder, createJobs=createJobs, user=user, 
+                                recurse=params.get('recurse'), localJobs=params.get('localJobs'))
+                    
+    def createImagesRecurseOption(self, folder, createJobs, user, recurse, localJobs):
+        count = {'large images created': 0}
+        if recurse:
+            for childFolder in Folder().childFolders(parent=folder, parentType='folder'):
+                self.createImagesRecurseOption(folder=childFolder, createJobs=createJobs, user=user, 
+                                        recurse=recurse, localJobs=localJobs)
+        for item in Folder().childItems(folder=folder):
```
annehaley, line 478: Does this call to `childItems` also need `limit=0`? And the call to `childFolders` a few lines up?

girder/girder_large_image/rest/large_image_resource.py
```diff
@@ -445,6 +447,55 @@ def _deleteCachedImages(self, spec, associatedImages=False, imageKey=None):
...
+    def createLargeImages(self, folder, params):
+        user=self.getCurrentUser()
+        createJobs = params.get('createJobs')
+        if createJobs:
+            createJobs = 'always' or True
+        self.createImagesRecurseOption(folder=folder, createJobs=createJobs, user=user, 
+                                recurse=params.get('recurse'), localJobs=params.get('localJobs'))
+                    
+    def createImagesRecurseOption(self, folder, createJobs, user, recurse, localJobs):
+        count = {'large images created': 0}
```
annehaley, line 473: I don't see where `count` is used. Is this intended to provide summary messages back to the user? We should include things like files skipped, folders recursed, and other info about the behavior.

gabrielle6249: While working I had printed count and just missed this line when cleaning up, but returning a summary message is a great idea!

girder/girder_large_image/rest/large_image_resource.py
```diff
@@ -445,6 +447,55 @@ def _deleteCachedImages(self, spec, associatedImages=False, imageKey=None):
...
         return removed
 
+    @access.user(scope=TokenScope.DATA_WRITE)
+    @autoDescribeRoute(
+        Description('Create large images for all items within a folder.')
+        .notes('Does not work for items with multiple files.')
+        .modelParam('id', 'The ID of the folder.', model=Folder, level=AccessType.WRITE, required=True)
+        .param('createJobs','Whether a job should be created for each large image.', required=False, 
+            default=False, dataType='boolean')
+        .param('localJobs', 'Whether a the jobs created should be local.', required=False,
```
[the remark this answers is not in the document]
gabrielle6249: Good idea, I'll change that.

girder/girder_large_image/rest/large_image_resource.py
gabrielle6249: It could probably be added in the method's docstring, but it's defined in Girder.

girder/girder_large_image/rest/large_image_resource.py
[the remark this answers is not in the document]
gabrielle6249: Changed in [this commit](https://github.com/girder/large_image/pull/1572/commits/ef19b736766a8f3410552291b78805fc62bc4563)

girder/girder_large_image/rest/large_image_resource.py
gabrielle6249: Removed that [here](https://github.com/girder/large_image/pull/1572/commits/ef19b736766a8f3410552291b78805fc62bc4563).

girder/girder_large_image/rest/large_image_resource.py
gabrielle6249: Added in [this commit](https://github.com/girder/large_image/pull/1572/commits/ef19b736766a8f3410552291b78805fc62bc4563)

girder/girder_large_image/rest/large_image_resource.py
annehaley: Ok, thanks. It's unrelated to this PR, so I'll mark as resolved

girder/girder_large_image/rest/large_image_resource.py
```diff
@@ -445,6 +447,70 @@ def _deleteCachedImages(self, spec, associatedImages=False, imageKey=None):
...
+                    except Exception:
+                        previousFileId = item['largeImage'].get(
+                            'originalId', item['largeImage']['fileId'])
+                        ImageItem().delete(item)
+                        ImageItem().createImageItem(item, File().load(user=user, id=previousFileId),
+                                                    createJob=createJobs, localJob=localJobs)
+                        result['largeImagesRemovedAndRecreated'] += 1
+            else:
+                largeImageFileId = None
+                files = list(Item().childFiles(item=item))
```
manthey, line 506: Add a `limit=2` to the `childFiles` call; we don't want to retrieve any more data than needed for this.

girder/girder_large_image/rest/large_image_resource.py
```diff
@@ -445,6 +447,70 @@ def _deleteCachedImages(self, spec, associatedImages=False, imageKey=None):
...
+                if len(files) == 1:
+                    largeImageFileId = str(files[0]['_id'])
+                    ImageItem().createImageItem(item, File().load(user=user, id=largeImageFileId),
```
manthey, line 509: These lines can be simplied -- there is no need to load the file; we already have it.  That ism you can get rid of `largeImageFileId = str(files[0]['_id'])` and just do `ImageItem().createImageItem(item, files[0], createJob=createJobs, localJob=localJobs)`

manthey (changes requested):

manthey:
I'd like to add a basic test of this.  Something like
```
@pytest.mark.usefixtures('unbindLargeImage')
@pytest.mark.plugin('large_image')
def testFolderCreateImages(server, admin, user, fsAssetstore):
    file = utilities.uploadExternalFile('sample_image.ptif', admin, fsAssetstore)
    itemId = file['itemId']
    item = Item().load(itemId, user=admin)
    folderId = str(item['folderId'])
    # Remove the large image from this item
    ImageItem().delete(item)
    # Ask to make all items in this folder large images
    resp = server.request(
        method='PUT', path=f'/large_image/folder/{folderId}/tiles', user=admin)
    assert utilities.respStatus(resp) == 200
    assert resp.json['largeImagesCreated'] == 1
    item = Item().load(itemId, user=admin)
    # Check that this item became a large image again
    assert 'largeImage' in item
```
could be added to the `girder/test_girder/test_large_image.py` file.

A better test set would also create a subfolder and put a file there to test recursion, and would test running the process a second time doesn't create any large images since they already exist and are fine.

girder/girder_large_image/rest/large_image_resource.py
gabrielle6249: That's much cleaner. Change made [here](https://github.com/girder/large_image/pull/1572/commits/71107ba9aa59ce3d162431c3bee4b1144444c44e).

girder/girder_large_image/rest/large_image_resource.py
gabrielle6249: Agreed. Changed that in [this commit](https://github.com/girder/large_image/pull/1572/commits/71107ba9aa59ce3d162431c3bee4b1144444c44e).

girder/girder_large_image/rest/large_image_resource.py
```diff
@@ -445,6 +447,72 @@ def _deleteCachedImages(self, spec, associatedImages=False, imageKey=None):
...
+        .param('recurse', 'Whether child folders should be recursed.', required=False,
+               default=False, dataType='boolean')
+        .errorResponse('ID was invalid.')
+        .errorResponse('Write access was denied for the folder.', 403),
+    )
+    def createLargeImages(self, folder, params):
+        user = self.getCurrentUser()
+        createJobs = params.get('createJobs')
+        if createJobs:
+            createJobs = 'always' or True
```
manthey, line 470: This seems to be a fragment from the item tiles endpoint with the force flag.  If this is a function renaming from the force flag, then it should go back to being called force for consistency and then internally, this goes back to `createJob='always' if self.boolParam('force', params, default=False) else True`.  The third state (don't use a job no matter what) would require the endpoint's parameter to accept three states (functionally these are `'always'`, never (False), and if-needed (True)).

gabrielle6249: In order for the parameter to accept the three states, do you think it's best to make a string parameter and change createJobs accordingly? Should the flag still be called force?

girder/test_girder/test_large_image.py
```diff
@@ -237,6 +237,25 @@ def testThumbnailFileJob(server, admin, user, fsAssetstore):
...
+    # Remove the large image from this item
+    ImageItem().delete(item)
+    # Ask to make all items in this folder large images
+    resp = server.request(
+        method='PUT', path=f'/large_image/folder/{folderId}/tiles', user=admin)
+    assert utilities.respStatus(resp) == 200
+    assert resp.json['largeImagesCreated'] == 1
+    item = Item().load(itemId, user=admin)
+    # Check that this item became a large image again
+    assert 'largeImage' in item
```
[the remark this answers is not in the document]
gabrielle6249: Function modified [here](https://github.com/girder/large_image/pull/1572/commits/5ae4647dec4600d5aced0a51286e02ec1958e64d).

girder/girder_large_image/rest/large_image_resource.py
```diff
@@ -445,6 +447,72 @@ def _deleteCachedImages(self, spec, associatedImages=False, imageKey=None):
...
+            result['totalItems'] += 1
+            if item.get('largeImage'):
+                if item['largeImage'].get('expected'):
+                    result['itemsSkipped'] += 1
+                else:
+                    try:
+                        ImageItem().getMetadata(item)
+                        result['itemsSkipped'] += 1
+                        continue
+                    except TileSourceError:
```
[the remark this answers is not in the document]
gabrielle6249: Good point. Change made [here](https://github.com/girder/large_image/pull/1572/commits/5ae4647dec4600d5aced0a51286e02ec1958e64d).

manthey (approved):
Great,  There are two minor tasks in #1139 that can be in future PRs (the ability to delete pending creation jobs and the ability to force un/redoing existing large images based on a filter).

PR #1586, opened 2024-07-24
Add options to cancel jobs, redo existing large images, and not force jobs

gabrielle6249:
This adds two more options from #1139 to the PUT large_image/folder/{id}/tiles endpoint, and allows the createJobs filter to accept three states. The first flag allows cancelling unfinished large image jobs related to items in the folder, so the large image can then be removed and a new one created. The second gives the option to remove and redo existing large images in the folder. The three states createJobs now accepts can either never use a job, use one only if needed, or force one for every large image.

### DIFF
--- a/girder/girder_large_image/rest/large_image_resource.py
+++ b/girder/girder_large_image/rest/large_image_resource.py
@@ -453,22 +453,22 @@ class LargeImageResource(Resource):
         .notes('Does not work for items with multiple files.')
         .modelParam('id', 'The ID of the folder.', model=Folder,
                     level=AccessType.WRITE, required=True)
-        .param('createJobs', 'Whether job(s) should be used to create each '
-               'large image. Select true to use a job if needed, false to '
-               'never use a job, and always to always use a job.',
-               required=False, default='false', dataType='string',
-               enum=['true', 'false', 'always'])
-        .param('localJobs', 'Whether the job(s) created should be local.',
-               required=False, default=False, dataType='boolean')
-        .param('recurse', 'Whether child folders should be recursed.',
-               required=False, default=False, dataType='boolean')
-        .param('cancelJobs', 'Whether unfinished large image job(s) associated '
-               'with items in the folder should be canceled, then a new large '
-               'image created. If false, items with an unfinished large image '
-               'will be skipped.', required=False, default=False,
-               dataType='boolean')
-        .param('redoExisting', 'Whether existing large images should be removed and '
-               'recreated.', required=False, default=False, dataType='boolean')
+        .param('createJobs', 'If true, a job will be used to create the image '
+               'when needed; if always, a job will always be used; if false, a '
+               'job will never be used.', dataType='string', default='false',
+               required=False, enum=['true', 'false', 'always'])
+        .param('localJobs', 'If true, run each creation job locally; if false, '
+               'run via the remote worker.', dataType='boolean', default='false',
+               required=False)
+        .param('recurse', 'If true, items in child folders will also be checked.',
+               dataType='boolean', default=False, required=False)
+        .param('cancelJobs', 'If true, unfinished large image job(s) associated '
+               'with items in the folder will be canceled, then a new large '
+               'image created; if false, items with an unfinished large image '
+               'will be skipped.', dataType='boolean', default=False, required=False)
+        .param('redoExisting', 'If true, existing large images should be removed and '
+               'recreated. Otherwise they will be skipped.', dataType='boolean',
+               default=False, required=False)
         .errorResponse('ID was invalid.')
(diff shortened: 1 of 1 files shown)


girder/girder_large_image/rest/large_image_resource.py
```diff
@@ -449,56 +449,100 @@ def _deleteCachedImages(self, spec, associatedImages=False, imageKey=None):
...
-        .param('force', 'Whether creation job(s) should be forced for each large image.',
+        Description('Create large images for all items within a folder.')
+        .notes('Does not work for items with multiple files.')
+        .modelParam('id', 'The ID of the folder.', model=Folder,
+                    level=AccessType.WRITE, required=True)
+        .param('createJobs', 'Whether job(s) should be used to create each '
+               'large image. Select true to use a job if needed, false to '
+               'never use a job, and always to always use a job.',
+               required=False, default='false', dataType='string',
+               enum=['true', 'false', 'always'])
```
marySalvi, line 460: Based on @manthey 's description in #1139 I get the impression this should be split into two options.
ie 'force' for always and 'localJob' making both boolean as seen here:
https://github.com/girder/large_image/blob/56193ce5784cf60c2591bd570a09aebaa3bef1fa/girder/girder_large_image/rest/tiles.py#L199-L205

girder/girder_large_image/rest/large_image_resource.py
```diff
@@ -449,56 +449,100 @@ def _deleteCachedImages(self, spec, associatedImages=False, imageKey=None):
...
+        .param('localJobs', 'Whether the job(s) created should be local.',
                required=False, default=False, dataType='boolean')
-        .param('localJobs', 'Whether the job(s) created should be local.', required=False,
-               default=False, dataType='boolean')
-        .param('recurse', 'Whether child folders should be recursed.', required=False,
-               default=False, dataType='boolean')
+        .param('recurse', 'Whether child folders should be recursed.',
```
marySalvi, line 463: I generally prefer definitions not to include the option unless _very_ self-explanatory. Perhaps something like: 
'If true, images in child folders will also be checked'

Also as a personal preference I would change the descriptions from 'Whether...' to 'If...'

girder/girder_large_image/rest/large_image_resource.py
```diff
@@ -449,56 +449,100 @@ def _deleteCachedImages(self, spec, associatedImages=False, imageKey=None):
...
-
-    def createImagesRecurseOption(self, folder, createJobs, user, recurse, localJobs):
+        createJobs = params.get('createJobs')
+        if createJobs in ['true', 'false']:
+            createJobs = self.boolParam('createJobs', params, default=False)
+        return self.createLargeImagesRecurse(folder=folder, user=user,
+                                             recurse=params.get('recurse'), createJobs=createJobs,
+                                             localJobs=params.get('localJobs'),
+                                             cancelJobs=params.get('cancelJobs'),
+                                             redo=params.get('redoExisting'))
```
marySalvi, line 484: I think breaking the options apart as suggested above would allow you to simplify this.

girder/girder_large_image/rest/large_image_resource.py
```diff
@@ -449,56 +449,100 @@ def _deleteCachedImages(self, spec, associatedImages=False, imageKey=None):
...
         if recurse:
             for childFolder in Folder().childFolders(parent=folder, parentType='folder'):
                 result['childFoldersRecursed'] += 1
-                childResult = self.createImagesRecurseOption(folder=childFolder,
-                                                             createJobs=createJobs, user=user,
-                                                             recurse=recurse, localJobs=localJobs)
+                childResult = self.createLargeImagesRecurse(folder=childFolder, user=user,
+                                                            recurse=recurse, createJobs=createJobs,
+                                                            localJobs=localJobs,
+                                                            cancelJobs=cancelJobs, redo=redo)
```
marySalvi, line 501: This feels logically backwards to me. I think I would expect:
```
def createLargeImage(self, folder):
    for item in folder:
        ***createImage logic here***
 
def folderRecurse(self, params):
     if recurse:
        for childFolder in folder:
           createLargeImage()
```
additionally I am not sure  I would make a new function to check the recurse param. That is a design choice for you and David to decide though.

marySalvi (commented):
Overall looks good. I have a few thoughts on simplifying things but will defer to @manthey

girder/girder_large_image/rest/large_image_resource.py
gabrielle6249: That reads clearer and explaining what happens depending on what the user selects is a good idea. Change made [here](https://github.com/girder/large_image/commit/e22df9496ed745dc055801bb6328d64c75afcc15).

manthey (approved):

PR #1597, opened 2024-08-01
Documentation update

annehaley:
This PR makes some minor improvements to the documentation:

- Reformat "Example Usages" page as a "Getting Started" page
- Add a favicon
- Change appearance of section links
- Group Girder documentation
- Group API documentation
- Format the Config Options page as a table, with a link to each row
- Add more links to provide context to mentioned topics/libraries
- Add a Zarr Sink example notebook
- Miscellaneous wording clarifications


docs/config_options.rst
```diff
@@ -3,43 +3,143 @@ Configuration Options
...
+     - Type
+     - Required
+     - Default
+
+       .. _config_logger:
+   * - ``logger`` :ref:`🔗 <config_logger>`
+     - Most log messages are sent here.
+     - ``logging.Logger``
+     -
+     -
```
manthey, line 21: This defaults to the standard python logger using the name ``large_image``.  When using Girder, this default to Girder's logger.

docs/config_options.rst
```diff
@@ -3,43 +3,143 @@ Configuration Options
...
+
+       .. _config_logprint:
+   * - ``logprint`` :ref:`🔗 <config_logprint>`
+     - Messages about available tilesources are sent here.
+     - ``logging.Logger``
+     -
+     -
```
manthey, line 28: This defaults to the standard python logger using the name ``large_image``.  When using Girder, this default to Girder's logger which allows colored console output.

docs/config_options.rst
```diff
@@ -3,43 +3,143 @@ Configuration Options
...
+
+       .. _config_cache_backend:
+   * - ``cache_backend`` :ref:`🔗 <config_cache_backend>`
+     - String specifying how tiles are cached.  If memcached is not available for any reason, the python cache is used instead.
+     - ``str: "python" | "memcached"``  # What about redis?
+     - ``False``
+     - ``"python"``
```
manthey, line 35: This can also be ``None``, which is the default.  When ``None``, the first cache available in the order memcached, redis, python is used.  Otherwise, the specified cache is used if available, falling back to python if not.

docs/config_options.rst
```diff
@@ -3,43 +3,143 @@ Configuration Options
...
+
+       .. _config_cache_python_memory_portion:
+   * - ``cache_python_memory_portion`` :ref:`🔗 <config_cache_python_memory_portion>`
+     - If tiles are cached with python, the cache is sized so that it is expected to use less than 1 / (``cache_python_memory_portion``) of the available memory.
+     - ``int``
+     -
+     -
```
manthey, line 42: Default is 16.

docs/config_options.rst
```diff
@@ -3,43 +3,143 @@ Configuration Options
...
+
+       .. _config_cache_tilesource_maximum:
+   * - ``cache_tilesource_maximum`` :ref:`🔗 <config_cache_tilesource_maximum>`
+     - If this is non-zero, this further limits the number of tilesources than can be cached to this value.
+     -
+     -
+     -
```
manthey, line 98: The minimum is always 3 (unless cache_sources is False), regardless of either this or the memory portion setting.

docs/config_options.rst
```diff
@@ -3,43 +3,143 @@ Configuration Options
...
+     - ``None``
+
+       .. _config_cache_tilesource_memory_portion:
+   * - ``cache_tilesource_memory_portion`` :ref:`🔗 <config_cache_tilesource_memory_portion>`
+     - Tilesources are cached on open so that subsequent accesses can be faster.  These use file handles and memory.  This limits the maximum based on a memory estimation and using no more than 1 / (``cache_tilesource_memory_portion``) of the available memory.
+     - ``int``
+     -
+     -
```
manthey, line 91: Memory usage by tile source is necessarily a rough estimate, since it can vary due to a wide variety of image-specific and deployment-specific details; this is intended to be conservative.  Default is 32.

docs/config_options.rst
```diff
@@ -3,43 +3,143 @@ Configuration Options
...
+     -
+
+       .. _config_cache_sources:
+   * - ``cache_sources`` :ref:`🔗 <config_cache_sources>`
+     - If set to False, the default will be to not cache tile sources.  This has substantial performance penalties if sources are used multiple times, so should only be set in singular dynamic environments such as experimental notebooks.
+     -
+     -
+     -
```
manthey, line 105: Default is True

docs/config_options.rst
```diff
@@ -3,43 +3,143 @@ Configuration Options
...
+
+       .. _config_max_small_image_size:
+   * - ``max_small_image_size`` :ref:`🔗 <config_max_small_image_size>`
+     - The PIL tilesource is used for small images if they are no more than this many pixels along their maximum dimension.
+     - ``int``
+     -
+     -
```
manthey, line 112: Default is 4096.  Specifying values greater than this _could_ reduce compatibility with tile use on some browsers.  In general 8192 is safe for all modern systems, and values greater than 16384 should not be specified if the image will be viewed in any browser.

docs/config_options.rst
```diff
@@ -3,43 +3,143 @@ Configuration Options
...
+
+       .. _config_source_ignored_names:
+   * - ``source_bioformats_ignored_names``,
+       ``source_pil_ignored_names``,
+       ``source_vips_ignored_names`` :ref:`🔗 <config_source_ignored_names>`
+     - Some tile sources can read some files that are better read by other tilesources.  Since reading these files is suboptimal, these tile sources have a setting that, by default, ignores files without extensions or with particular extensions.
+     - ``str`` (regular expression)
+     -
+     - ``r'(^[!.]*|\.(jpg|jpeg|jpe|png|tif|tiff|ndpi))$'`` (for bioformats)
```
manthey, line 121: Bioformats actually has a longer default.  Maybe just add a note that "see each source for its default value".

docs/config_options.rst
```diff
@@ -3,43 +3,143 @@ Configuration Options
...
+
+       .. _config_all_sources_ignored_names:
+   * - ``all_sources_ignored_names`` :ref:`🔗 <config_all_sources_ignored_names>`
+     - If a file matches the regular expression in this setting, it will only be opened by sources that explicitly match the extension or mimetype.  Some formats are composed of multiple files that can be read as either a small image or as a large image depending on the source; this prohibits all sources that don't explicitly support the format.
+     -
+     -
+     -
```
manthey, line 128: Default is `'(\.mrxs|\.vsi)$'`.  Typically, these should be formats that required multiple files to work but could also be opened as single files lacking details.

docs/config_options.rst
```diff
@@ -3,43 +3,143 @@ Configuration Options
...
+
+       .. _config_icc_correction:
+   * - ``icc_correction`` :ref:`🔗 <config_icc_correction>`
+     -  If this is True or undefined, ICC color correction will be applied for tile sources that have ICC profile information.  If False, correction will not be applied.  If the style used to open a tilesource specifies ICC correction explicitly (on or off), then this setting is not used.  This may also be a string with one of the intents defined by the PIL.ImageCms.Intents enum.  ``True`` is the same as ``perceptual``.
+     - ``bool | str``
+     -
+     -
```
manthey, line 135: Default is True.

docs/config_options.rst
annehaley: Could you clarify this one? What is the default value for `cache_tilesource_maximum`?

manthey: The default is `0` which is the signifier that `cache_tilesource_memory_portion` determines the number of sources that will be cached.  If >0 the cache will be the smaller of the value computed for the memory portion and this value (but always at least 3).

annehaley: If the user specifies `cache_tilesource_maximum=2`, is the behavior the same as specifying 1 or 0? Does this value only work if greater than 3?

manthey: Specifying 1 or 2 is the same as specifying 3.

annehaley: Thanks. Added in [4465d0c](https://github.com/girder/large_image/pull/1597/commits/4465d0c81b67d58e243766b6441257327a2cfcdf)

manthey: The primary reason to set this is to limit memory use.  In the (expectedly) rare cases where you don't want caching, you'd set `cache_sources` to False.

manthey: In some sense the minimum of 3 should be considered an implementation detail.  I could imagine a refactor where a value of 1 or 2 could be honored.

Internally, we treat styled sources as a separate cache entry to unstyled sources, which means that, currently for the cache to actually work for styled sources we need to keep both the styled and unstyled source in the cache (hence a minimum of 2).  Further, when opening a multi-source source that uses styled base sources, a value of 2 can cause the cache to cycle and act like it is functionally not there (and thus we get the minimum of 3).  In practice, you could have a multi-source source that could be itself styled and could have many styled base sources, then you'd really want this number to be 1 + (1 if multi-source source is styled else 0) + number of distinct based sources + number of distinct style/base source combinations, , but this isn't always practicable (e.g., we've seen multi-source sources with 50,000 base sources).

We pick 3 as a minimum with a reasonable compromise for honoring the intent and not breaking caching in common situations.

docs/tilesource_options.rst
```diff
@@ -59,15 +59,15 @@ A band definition is an object which can contain the following keys:
 
 - ``nodata``: the value to use for missing data.  null or unset to not use a nodata value.
 
-- ``composite``: either 'lighten' or 'multiply'.  Defaults to 'lighten' for all except the alpha band.
+- ``composite``: either 'lighten' or 'multiply'.  Defaults to 'lighten' for all except the alpha band. Read more about blend modes and see examples `here <blend_modes_>`_.
```
manthey, line 62: We don't use the python library you've linked to.  Maybe the css definitions are better: `https://developer.mozilla.org/en-US/docs/Web/CSS/blend-mode`.

docs/tilesource_options.rst
```diff
@@ -59,15 +59,15 @@ A band definition is an object which can contain the following keys:
...
+- ``icc``: by default, sources that expose ICC color profiles (PIL, OpenJPEG, TIFF, TiffFile) will apply those profiles to the image data, converting the results to the sRGB profile.  To use the raw image data without ICC profile adjustments, specify an ``icc`` value of ``false``.  If the entire style is ``{"icc": false}``, the results will be the same as the default bands with only the adjustment being skipped.  Similarly, if the entire style is ``{"icc": true}``, this is the same as the default style with where the adjustment is applied.  Besides a boolean, this may also be a string with one of the intents defined by the PIL.ImageCms.Intents enum.  ``true`` is the same as ``perceptual``.   Note that not all tile sources expose ICC color profile information, even if the base file format contains it.
```
manthey, line 70: This source list isn't complete (also OpenSlide and OMETiff), plus there could be other sources not in our monorepo (e.g., iSyntax) that expose this, too.  I fear that listing these explicitly will become out of date at some point without us noticing.

docs/zarr_sink_example.ipynb
```diff
@@ -0,0 +1,545 @@
+{
+ "cells": [
+  {
+   "cell_type": "markdown",
+   "id": "9190dbbd",
+   "metadata": {},
+   "source": [
+    "# Using the Zarr Tile Sink\n",
+    "\n",
+    "The `ZarrFileTileSource` class has file-writing capabilities; a user may start with an empty Zarr store, add data tiles, then save the data to a file of any format. \n",
```
manthey, line 10: This is along the lines of how much do we need to explain about external packages.  Someone new to zarr won't know what store is and doesn't need to know to use our library.  And, although we add data via `addTile`, we really are adding arbitrary regions, so perhaps a better wording would be "an empty image can be created, image data can be added as tiles or arbitrary regions, and the image can be saved to a file in any of several formats".

docs/zarr_sink_example.ipynb
```diff
@@ -0,0 +1,545 @@
...
+    "# computing the processed image takes about 1 minute for each value\n",
+    "footprint_sizes = [1, 10, 50]\n",
+    "print(f'Processing image for {len(footprint_sizes)} frames.')\n",
+    "\n",
+    "# create a frame for each processed result\n",
+    "for i, footprint_size in enumerate(footprint_sizes):\n",
+    "    print('Processing image with footprint_size = ', footprint_size)\n",
+    "    # iterate through tiles, getting numpy arrays for each tile\n",
+    "    for tile in source.tileIterator(format='numpy'):\n",
+    "            # for each tile, run some processing algorithm\n",
```
manthey, line 296: Can we make the indent of this section just 8 spaces rather than 12?

docs/config_options.rst
```diff
@@ -56,6 +137,8 @@ Configuration from Environment
 
 All configuration parameters can be specified as environment parameters by prefixing their uppercase names with ``LARGE_IMAGE_``.  For instance, ``LARGE_IMAGE_CACHE_BACKEND=python`` specifies the cache backend.  If the values can be decoded as json, they will be parsed as such.  That is, numerical values will be parsed as numbers; to parse them as strings, surround them with double quotes.
 
+.. The mention of numbers needing quotes to be considered strings could be confusing. Do the above variables need numbers as strings?
```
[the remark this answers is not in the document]
annehaley: This makes sense, thanks for explaining. It looks like I forgot to ask about this note and resolve it sooner. I don't think this needs additional explanation, so I'll just remove my comment.

docs/tilesource_options.rst
annehaley: I'll add OpenSlide and OMETiff for now, but I do think there needs to be a list somewhere. I had searched the code to find mentions of "icc", and the fact that I missed some demonstrates that it wouldn't be easy for a user to discover this list on their own. The `OMETiffTileSource` code doesn't mention "icc" at all, but I now see that it inherits that functionality from `TiffFileTileSource`.

annehaley:
@manthey I have addressed the above comments all in one commit. See changes here: [0afcfb5](https://github.com/girder/large_image/pull/1597/commits/0afcfb5f9e142c1c1d91206c525d797d92c13114).

manthey (approved):

manthey (approved):

PR #1599, opened 2024-08-02
Use linkchecker to verify all docs links

manthey:
@annehaley It occurred to me that we should test the links that the documentation generates to ensure that they go to actual destinations.  I added a test as part of building the docs with tox.


manthey:
@annehaley I made one change to the docs to make this pass.

PR #1610, opened 2024-08-20
Fix the README.rst so we can deploy.

manthey:
Because we use the README as part of the built docs, we added references into it.  These won't deploy, as the description uploaded to pypi.org can't use them directly.


manthey:
@annehaley I'll merge this when it passes; if we want the references I removed to be put back, we'll have to figure out how to add them while still getting the README.rst to pass twine's linter (which seems to be equivalent to the rst-linter in https://github.com/Lucas-C/pre-commit-hooks-markup)

PR #1625, opened 2024-09-04
Zarr Sink Frame Values

### LINKED ISSUE #1325: Axis conceptual values
Multi-axis images have integer indices along each axis.  We should also be able to specify the conceptual value.  This would be exposed as Value<axis>: <value>, and a summary of the values at the top level.  These values could be numeric with units or text.

The multi-source source would allow overriding these values.  Each format could have a different way of extracting these values.

annehaley:
Resolves #1325. This PR adds frame values to the Zarr sink, which allows a user to specify true values for each index of a frame axis. This can be done with `addTile` or by setting `frameValues` directly. Additionally, the user may set a dictionary of `frameUnits` to describe the units used by each frame axis and/or a list of `frameAxes` to assert an ordering to the frame axes. 

For example:
```
import numpy as np

sink.frameAxes = ['z', 't', 'm']
sink.frameUnits = dict(z='cm', t='ms', m='mm')
for z, z_value in enumerate([2, 4, 6, 8]):
    for t, t_value in enumerate([10.0, 20.0, 30.0]):
        t_value += 0.01 * z
        for m, m_value in enumerate(['r', 'g', 'b']):
            random_tile = np.random.random((300, 400, 3))
            sink.addTile(
                random_tile,
                0, 0,
                z=z, t=t, m=m,
                z_value=z_value, t_value=t_value, m_value=m_value
            )
print(sink.getMetadata())
```

This PR also uses the frame values in the slider labels shown in the Frame Selector component. The following video demonstrates the value labels on axis sliders for a randomly-generated image using the new Zarr sink API. 
C and Z have uniform values, while T has non-uniform values. (This image was generated with the code in `testFrameValues` in `test/test_sink.py`.)
https://github.com/user-attachments/assets/4b41bcd3-8164-46f8-b78b-9d919cecefa0

### DIFF
--- a/docs/getting_started.rst
+++ b/docs/getting_started.rst
@@ -334,25 +334,60 @@ You can also composite a multi-frame image into a false-color output:
 Writing an Image
 ----------------
 
-If you wish to visualize numpy data, large_image can write a tiled tiff.  This requires a tile source that supports writing to be installed.  As of this writing, the ``large-image-source-zarr`` and ``large-image-source-vips`` sources supports this.  If both are installed, the ``large-image-source-zarr`` is the default.
+If you wish to visualize numpy data, ``large_image`` can write a tiled image.
+This requires a tile source that supports writing to be installed.
+As of this writing, the ``large-image-source-zarr`` and ``large-image-source-vips`` sources both support this.
+If both are installed, the ``large-image-source-zarr`` is the default.
+Some of the API options available for ``large-image-source-zarr`` are not available for ``large-image-source-vips``.
 
 .. code-block:: python
 
     import large_image
+
     source = large_image.new()
     for nparray, x, y in fancy_algorithm():
         # We could optionally add a mask to limit the output
         source.addTile(nparray, x, y)
     source.write('/tmp/sample.tiff', lossy=False)
 
-The ``large-image-source-zarr`` can be used to store multiple frame data with arbitrary axes.
+Multiple Frames
+~~~~~~~~~~~~~~~
+
+``large-image-source-zarr`` can be used to store multiframe data with arbitrary axes.
+The example below demonstrates the creation of an image with five axes: T, Z, Y, X, S.
 
 .. code-block:: python
 
     import large_image
+
+    time_values = [0.5, 1.5, 2.5, 3.5]
+    z_values = [3, 6, 9]
+    tile_pos_values = [0, 1024, 2048, 3072, 4096]
+
     source = large_image.new()
-    for nparray, x, y, time, param1 in fancy_algorithm():
-        source.addTile(nparray, x, y, time=time, p1=param1)
+    for t_index, t_value in enumerate(time_values):
+        for z_index, z_value in enumerate(z_values):
+            for y_value in tile_pos_values:
+                for x_value in tile_pos_values:
+
+                    # tile is a numpy array with shape (1024, 1024, 3)
+                    # this shape corresponds to the following axes, respectively: (Y, X, S)
+                    tile = get_my_data_tile(x_value, y_value, z_value, t_value)
+
+                    source.addTile(
+                        tile,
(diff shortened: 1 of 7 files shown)


manthey:
If you have a multi-band tile and specify a c_value, it doesn't work as expected.  That is
```
ts = large_image.new()
ts.addTile(np.zeros((100, 100, 3)), x=0, y=0, c=0, z=0, z_value=1, c_value='DAPI')
ts.addTile(np.zeros((100, 100, 3)), x=0, y=0, c=1, z=0, z_value=1, c_value='CD4')
```
fails, but without the `c_value` succeeds.

manthey:
Another use case I was expecting to work:
```
ts = large_image.new()
ts.addTile(np.zeros((100, 100, 1)), x=0, y=0, z=0, z_value=1)
ts.addTile(np.zeros((100, 100, 1)), x=0, y=0, z=1, z_value=3.2)
ts.addTile(np.zeros((100, 100, 1)), x=0, y=0, z=2, z_value=6.4)
ts.addTile(np.zeros((100, 100, 1)), x=0, y=0, z=2, z_value=6.3, c=1)
```

annehaley:
The first two snippets you provided have been added as a pytest and fixed in [c85ebf0](https://github.com/girder/large_image/pull/1625/commits/c85ebf08d9bc6df034b66a537cb0b87374e7f3d7). 
The third snippet you provide is a use case that is already not supported by the `addTile` function. The current implementation of `addTile` requires that all axes are declared upon the first call, and a user cannot add new axes later. While outside the scope of this PR, this use case is reasonable and should be addressed in another PR. I have created #1674 to track this.

manthey:
I had another example that had been failing during `addTile`, but with the latest changes now fails when getting the metadata:
```
ts = large_image.new()
ts.addTile(np.zeros((100, 100, 3)), x=0, y=0, c=0, z=0, z_value=1, c_value='DAPI')
ts.addTile(np.zeros((100, 100, 3)), x=0, y=0, c=0, z=1, z_value=3.2, c_value='DAPI')
ts.addTile(np.zeros((100, 100, 3)), x=0, y=0, c=0, z=2, z_value=6.3, c_value='DAPI')
ts.addTile(np.zeros((100, 100, 3)), x=0, y=0, c=1, z=2, z_value=6.4, c_value='CD4')
ts.addTile(np.zeros((100, 100, 3)), x=0, y=0, c=1, z=1, z_value=3.1, c_value='CD4')
ts.addTile(np.zeros((100, 100, 3)), x=0, y=0, c=0, z=1, z_value=1.1)
pprint.pprint(ts.metadata)
```
My guess is that this is due to unpopulated values in some combinations.

manthey:
Currently, the metadata entries for `Value<axis>` is in the zarr source code.  I think it would be better to move it to the `_addMetadataFrameInformation` method in the base class so if these values are populated in frames in non-zarr sources, the aggregated values will still be added to the metadata.

manthey:
Should we cast the values stored in each frame to native python `float` or `int` datatypes rather than leave them as numpy data types?  As long as the native numpy data types serialize to json, it probably doesn't matter, but for consistency between sources, we may want to do the cast.

manthey (approved):

manthey:
I'd like to have issues for two follow ups to this PR:

- Setting values in multiple processes doesn't appear in the final result.  That is, I think only the last process to white the zarr has sets values.

- The display precision should probably be limited to some number of significant digits.  For instance, in a test example I get 
image

annehaley:
@manthey Since the value precision for the labels is a quick fix, I added that change in [e41e125](https://github.com/girder/large_image/pull/1625/commits/e41e1256214abfaaa2a37cad9d6c0215678d284a). We can address the multiprocessing case in a follow-up PR as you suggested. I made this issue to track it: #1698.

Issue #1639, opened 2024-09-12
MultiFileTileSource: specify backing source module


banesullivan-kobold:
Could we specify to the `MultiFileTileSource` to use the `GDALFileTileSource` to open each of the images? 

I have a lot of pyramidal tiffs (non-geospatial) and I compose them in YAML to stitch them together for use with `MultiFileTileSource`. However the images are all stored on S3 and I have hundreds of thousands of images. I want to be able to use GDAL's virtual file system and `/vsis3/` for performant reads directly from S3 (no downloading of data locally)

I think this is feasible from my look over the MultiFileTileSource, but would appreciate any clarifying details. Or is it possible to performantly load pyramidal tiffs from cloud storage with another tile source?

manthey:
You can specify the "sourceName" of each source, which I _think_ lets you do this.

banesullivan-kobold:
Ooh! This is a part of the YAML spec itself, I see now! 

However, this is leading to an error with the `gadal` source:

```
TypeError: large_image.tilesource.base.TileSource.tileIterator() got multiple values for keyword argument 'resample'
```

<details>

```py
File [/python/lib/python3.11/site-packages/large_image/cache_util/cache.py:106](https://jupyterhub-production.internal.koboldmetals.com/python/lib/python3.11/site-packages/large_image/cache_util/cache.py#line=105), in methodcache.<locals>.decorator.<locals>.wrapper(self, *args, **kwargs)
    103 except (ValueError, pickle.UnpicklingError):
    104     # this can happen if a different version of python wrote the record
    105     pass
--> 106 v = func(self, *args, **kwargs)
    107 try:
    108     if lock:

File [/python/lib/python3.11/site-packages/large_image/tilesource/base.py:1599](https://jupyterhub-production.internal.koboldmetals.com/python/lib/python3.11/site-packages/large_image/tilesource/base.py#line=1598), in TileSource.getThumbnail(self, width, height, **kwargs)
   1597 params['output'] = {'maxWidth': width, 'maxHeight': height}
   1598 params.pop('region', None)
-> 1599 return self.getRegion(**params)

File [/python/lib/python3.11/site-packages/large_image/tilesource/base.py:1748](https://jupyterhub-production.internal.koboldmetals.com/python/lib/python3.11/site-packages/large_image/tilesource/base.py#line=1747), in TileSource.getRegion(self, format, **kwargs)
   1746     kwargs = kwargs.copy()
[...]

PR #1640, opened 2024-09-12
Fix an issue when asking not to resample a GDAL tile iterator

manthey:
Fixes an issue discussed in #1639


banesullivan-kobold (approved):
This indeed fixes the issue I ran into

PR #1649, opened 2024-09-25
Setting global config file based on settings

jeffbaumes:
The table view is now always active because there is always a config file, which is built from the legacy settings if no actual config file exists in the folder hierarchy.

There is now also a place to store user-specific settings, which override group settings, which override general access level settings, which override top-level settings.

@manthey Note that even though you suggested creating the file with `force=bool(user_context)`, this is not an option when creating a file so that is currently not in there.


manthey:
> @manthey Note that even though you suggested creating the file with `force=bool(user_context)`, this is not an option when creating a file so that is currently not in there.

I think it needs to be on the `createItem` call.

girder/girder_large_image/__init__.py
```diff
@@ -483,15 +483,65 @@ def adjustConfigForUser(config, user):
                     {'_id': {'$in': user['groups']}}, sort=[('name', SortDir.ASCENDING)]):
                 if isinstance(groups.get(group['name']), dict):
                     config = _mergeDictionaries(config, groups[group['name']])
+    # Do this after merging groups, because the group access-level values can
+    # override the base access-level options.  For instance, if the base has
+    # an admin option, and the group has a user option, then doing this before
+    # group application can end up with user options for an admin.
```
manthey, line 489: @jeffbaumes Note the rationale here -- now documented.

manthey:
> > @manthey Note that even though you suggested creating the file with `force=bool(user_context)`, this is not an option when creating a file so that is currently not in there.
> 
> I think it needs to be on the `createItem` call.

I added this in https://github.com/girder/large_image/pull/1649/commits/f2facdbac61c9a7133e8de7d7d7a37352a699a17

manthey (approved):

jeffbaumes:
I updated to incorporate your changes and everything seems to be working as intended with some in-progress config-changing UI.

PR #1691, opened 2024-10-16
Format Examples Pooch Datastore

annehaley:


manthey (approved):
In the ideal world, we wouldn't have two hashes for these files, but that can be put off for when it feels cumbersome when we add more examples.

PR #1723, opened 2024-11-12
Fix an issue searching for annotation metadata

manthey:
If a user has permission to view an annotation but not the item, a 403 was percolated to the UI


manthey:
To replicate, create an item in a public folder with an annotation with metadata that you can find in a search.  Switch the folder to private.  The annotation metadata search throws a 403 that is visible in the UI.  After this change, the item just isn't listed.

PR #1726, opened 2024-11-13
Add a distinct class name to controls in the item list

manthey:


manthey:
@jeffbaumes This adds a class name so if the only control is the download control it is still discoverable by other plugins.

jeffbaumes (approved):
I do not have a system that includes DIVE to fully check but LGTM. Thanks for fixing the typos

PR #1738, opened 2024-12-06
Jupyter Frame Selector component

annehaley:
This PR moves and refactors the component files for the Frame Selector such that the widget can be used in the Girder/Backbone context as well as the Jupyter context, using the same source code.

https://github.com/user-attachments/assets/9fe5f190-87d1-4cd3-8d4d-1dd041c6c988


large_image/tilesource/jupyter.py
```diff
@@ -393,19 +399,55 @@ def from_map(self, coordinate: Union[List[float], Tuple[float, float]]) -> Tuple
...
+            netloc = parsed_url.netloc
+            path = parsed_url.path
+            query = parsed_url.query
+            query = {k: v[0] for k, v in parse_qs(query).items()}
+            query.update(dict(
+                frame=frame,
+                style=json.dumps(style),
+            ))
+            query_string = urlencode(query, quote_via=quote, safe='{}')
+            self._layer.url = f'{scheme}://{netloc}{path}?{query_string}'
```
manthey, line 415: scheme could be `''`, at which point this will fail.  It is better to use urlunparse to make sure it is done right.

annehaley: Ok, changed in 724c7f9

large_image/tilesource/jupyter.py
```diff
@@ -393,19 +398,59 @@ def from_map(self, coordinate: Union[List[float], Tuple[float, float]]) -> Tuple
...
+                parsed_url.scheme, parsed_url.netloc,
+                parsed_url.path, parsed_url.params,
+                query_string, parsed_url.fragment,
+            ))
             self._layer.redraw()
 
+    def get_frame_histogram(self, query):
+        if self._layer is not None:
+            frame = query.get('frame')
+            frame_histograms = self.frame_selector.frameHistograms or {}
```
manthey, line 421: Why do you copy this and modify the copy?  If there are concurrent histogram requests, won't this end up just keeping the last one?

annehaley: The copy is done so that we set `self.frame_selector.frame_histograms` to a new object to trigger the Vue watcher that detects changes to that object. Without the copy statement, only the first histogram will be shown and Vue cannot detect subsequent updates. I haven't encountered a case where concurrent requests overwrite one another, but maybe some extra bookkeeping on this class could resolve that.

annehaley: Extra bookkeeping added in [303ebe3](https://github.com/girder/large_image/pull/1738/commits/303ebe30d932c8c0c69fa020ad811c12cd719b72).

manthey (approved):

PR #1745, opened 2024-12-16
Make pre-release versions of large_image for Girder 5

naglepuff:


naglepuff:
@zachmullen I used https://github.com/girder/girder/pull/3533 as a guide here. 

Am I correct in thinking that since sub-directories here in `large_image`  (such as `large_image/girder`, `large_image/girder/girder_annotation`) have 

```python
version = get_version(root='..', local_scheme=prerelease_local_scheme)
```

a prerelease version will be available for those sub-packages?

My goal here is to make a prerelease version of `girder-large-image-annotation` available for HistomicsUI.

zachmullen (approved):

PR #1752, opened 2025-01-02
When reopening a zarr sink, validate it

manthey:
Also, write the metadata after fundamental changes so that anything that reopens the file after that point will see those changes.


manthey:
@annehaley This fixes the memory issues I found in #1713 with the example of preallocating the axis size.

annehaley (approved):

PR #1776, opened 2025-01-17
Default to not caching sources in notebooks

manthey:
@annehaley This should remove the need to add `nocache=True` to let the frame selector work in notebooks.

@banesullivan This should make your work easier, too.


banesullivan:
Fantastic!

PR #1777, opened 2025-01-17
Build plugin client code on release

naglepuff:
Ship Girder plugin packages with built client packages.

This makes it easier for dependent plugins to use `large_image` Girder plugins.

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -296,7 +296,7 @@ jobs:
           path: ~/wheels
   check_release:
     docker:
-      - image: cimg/python:3.10
+      - image: girder/tox-and-node
     steps:
       - checkout
       - run:
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -297,6 +297,10 @@ jobs:
   check_release:
     docker:
       - image: girder/tox-and-node
+    parameters:
+      node:
+        type: string
+        default: v20
     steps:
       - checkout
       - run:
@@ -307,6 +311,13 @@ jobs:
       - run:
           name: Install python packages
           command: pip install setuptools_scm twine
+      - run:
+          name: Switch node versions
+          command: |
+            nvm install << parameters.node >>
+            nvm alias default << parameters.node >>
+            NODE_DIR=$(dirname $(which node))
+            echo "export PATH=$NODE_DIR:\$PATH" >> $BASH_ENV
       - run:
           name: Check node versions
           command: |
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -322,6 +322,7 @@ jobs:
       - run:
           name: Switch node versions
           command: |
+            . ~/.bashrc
             nvm install << parameters.node >>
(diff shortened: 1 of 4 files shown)


.circleci/config.yml
```diff
@@ -296,7 +296,11 @@ jobs:
           path: ~/wheels
   check_release:
     docker:
-      - image: cimg/python:3.10
+      - image: girder/tox-and-node
+    parameters:
+      node:
+        type: string
+        default: v20
```
naglepuff, line 303: @manthey I'm a bit confused about the failure of the `check_release` [job for this branch](https://app.circleci.com/pipelines/github/girder/large_image/6386/workflows/58cf8ea3-bed1-4cb1-81e8-859b6b832727/jobs/45523). I believe that `girder/tox-and-node` is built from the `test.Dockerfile` file, which [does install nvm](https://github.com/girder/large_image/blob/ec069cb0b1dc9bc113f04b96a2f12703c0a27c6d/test.Dockerfile#L119-L120). Am I misinterpreting the `docker.image` config option?

.circleci/config.yml
```diff
@@ -307,6 +311,27 @@ jobs:
...
+          command: |
+            echo 'export NVM_DIR="/opt/circleci/.nvm"' >> $BASH_ENV
+            echo '[ -s "$NVM_DIR/nvm.sh" ] && \. "$NVM_DIR/nvm.sh"' >> $BASH_ENV
+      - restore_cache:
+          name: Restore nvm cache
+          key: v1-nvm-cache-<< parameters.node >>
+      - run:
+          name: Switch node versions
+          command: |
+            nvm install << parameters.node >>
```
manthey, line 326: Since nvm is just bash scripts, apparently we need to source the bashrc before they will work.  We haven't actually switched node versions in CI in a while, so either something changed or we never fully tested this.

Add
```
            . ~/.bashrc
```
before the `nvm install` line.

naglepuff: Thanks, this seems to have done the trick

.circleci/config.yml
```diff
@@ -296,7 +296,11 @@ jobs:
           path: ~/wheels
   check_release:
     docker:
-      - image: cimg/python:3.10
+      - image: girder/tox-and-node
```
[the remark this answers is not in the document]
naglepuff: I went ahead and made this change.

naglepuff:
@manthey I had to add `MANIFEST.in` files for `girder_large_image` and `girder_large_image_annotation`. Locally this seems to have the intended effect of including built client artifacts in the python distribution. This is what girder plugins in the main `girder` repo do, so I copied that behavior here.

manthey:
As a follow on (I'll make a PR): we also have a make_wheels.sh script that currently runs as CI and pushes wheels as artifacts.  We should get rid of that and instead push the wheels we make in the check_release as artifacts.  We should also adjust the release script to push all wheels to twine in one command rather than each in turn, so if the actual twine command fails, it fails for all components.

PR #1779, opened 2025-01-20
Remove unused type: ignore

naglepuff:


setup.py
```diff
@@ -108,7 +108,10 @@ def prerelease_local_scheme(version):
         'cachetools',
         'palettable',
         'Pillow',
-        'numpy',
+        # Newer versions of numpy include some typing changes that may need to be addressed in
+        # large_image.
+        # https://github.com/numpy/numpy/issues/28076
+        'numpy<2.1.0',
```
manthey, line 114: @naglepuff See #1780; it addresses these.

PR #1786, opened 2025-01-23
Combine the wheels and check-release CI jobs

manthey:
Publish wheel artifacts as part of the check release job.  Publish all wheels using a single twine command so partial failures are better handled.


manthey:
I have to fix how we build the slim.Dockerfile and other other docker images before the github actions will pass.

manthey:
@naglepuff This should be good for review now.

naglepuff (approved):

PR #1792, opened 2025-01-28
Check that aiohttp is present in Jupyter environments

manthey:


manthey:
@annehaley I should have caught this in #1738.

annehaley (approved):
Thanks for finding this!

PR #1803, opened 2025-02-03
Histogram AutoRange True Values

### LINKED ISSUE #1030: Add a method or endpoint to get histogram max/min:<value> direct values
If you use a style min/max of min:<value>/max:<value>, it would be useful to know what value are computed for those parameters so that a histogram display can show where those are located.

annehaley:
When using the Histogram Editor within the Frame Selector component, this PR keeps the min and max input boxes visible even when autoRange is enabled (though they are disabled in that case). The min and max inputs show the real values when specifying range with a distribution percentage.

Resolves #1030.

image

### DIFF
--- a/large_image/widgets/HistogramEditor.vue
+++ b/large_image/widgets/HistogramEditor.vue
@@ -3,7 +3,7 @@ function clamp(num, min, max) {
     return Math.min(Math.max(num, min), max);
 }
 
-function makeDraggableSVG(svg, validateDrag, callback, xRange) {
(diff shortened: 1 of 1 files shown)


manthey:
I like this a lot, but in the HistomicsUI instance of this we don't have space to show these boxes (since the right side of the area we do have gets used up by the auto-range percent.  We can hide these in that instance, but it'd cool if we can also add tooltips maybe to the grey areas of the histogram (or the drag handles) that would also show these values?

manthey:
I think you need "user-select" none on the tooltip bubbles.

PR #1815, opened 2025-02-10
Ipyleaflet Region Indicator

annehaley:
This PR adds widgets to the `ipyleaflet` image viewer in Jupyter notebooks. The user may draw (and remove) rectangles on the image. Clicking on the rectangle shows a popup with information about the selected region. This PR also adds a fullscreen view button to the options along the side of the viewer. The following screenshots are captured in fullscreen mode.

Geospatial example:
image

Non-geospatial example:
image

### DIFF
--- a/large_image/tilesource/jupyter.py
+++ b/large_image/tilesource/jupyter.py
@@ -350,7 +350,7 @@ class Map:
         return VBox(children)
 
     def add_region_indicator(self):
-        from ipyleaflet import GeomanDrawControl, FullScreenControl, Popup
+        from ipyleaflet import FullScreenControl, GeomanDrawControl, Popup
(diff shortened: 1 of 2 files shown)


manthey:
It seems to have nothing to do with this branch, but I can't zoom out far enough on some images.  We should at least create another issue to auto-scale ipyleaflet viewers so that the image fits the viewer on first load.

manthey:
If I open a geospatial source without a projection, this fails.

And, on geospatial source with a projection, the bounding coordinates are in interface coordinates (lat/lon) and the bounding box appears to be in map projection coordinates (web mercator).  I think they should both be in interface coordinates.

manthey (approved):

PR #1817, opened 2025-02-10
Fix JS bug in metadata editor widget

zachmullen:


manthey:
@zachmullen I think the same change is needed on line 651.

zachmullen:
Good catch

PR #1831, opened 2025-02-20
Add `minWidth` and `minHeight` to Zarr Sink

### LINKED ISSUE #1809: Specifying the whole slide image size beforehand while creating the zarr sink object
Currently, the size of zarr sink image depends on array/tile added to it. This however causes a size mismatch issue between the original image and the processed image that is written to sink. For example, I want to create superpixel mask for a wsi, and generate superpixels for only those tiles that contain tissue. If the right-most/bottom-most tiles don't have any tissue, they won't be processed and written to the sink. Thus, the resultant sink image would have smaller width/height. To tackle this an empty array for non-tissue tiles explicitly need to be written (especially at the right/bottom ends).

annehaley:
Resolves #1809.

### DIFF
--- a/sources/zarr/large_image_source_zarr/__init__.py
+++ b/sources/zarr/large_image_source_zarr/__init__.py
@@ -789,12 +789,7 @@ class ZarrFileTileSource(FileTileSource, metaclass=LruCacheMetaclass):
             else:
                 arr = current_arrays[store_path]
                 new_shape = tuple(
-                    max(
-                        v,
-                        self.minWidth if self.minWidth is not None and k == 'x' else
-                        self.minHeight if self.minHeight is not None and k == 'y' else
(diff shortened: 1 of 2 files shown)


manthey:
Conceptually, we probably only want to apply minWidth and minHeight when we write to a file (or read back), rather then when we add a tile.  For instance, imagine the code path:
```
import large_image
import numpy as np
s = large_image.new()
s.addTile(np.zeros((256, 256, 1), dtype=np.uint8), x=0, y=0)
s.minWidth = 1024
s.addTile(np.zeros((256, 256, 1), dtype=np.uint8), x=256, y=0)
s.minWidth = 768
s.write('sample.tiff')
```
Here I'd expect width of the output to be 768 (not 1024).

manthey (approved):

PR #1833, opened 2025-02-20
Reorder zarr axes in read mode

### LINKED ISSUE #1703: Zarr axis order
Prior to PR #1625, we always preferred to have the C axis have a stride of 1 if it was present.  This changed with writing axes, but I don't think we want this for the general reading case.  There are some users who expect the channel to be the inner mode frame axis.

This specifically is here: https://github.com/girder/large_image/blob/master/sources/zarr/large_image_source_zarr/__init__.py#L449-L452

We could make it conditional on whether we are editing or not, since while using the image as a sink reading back in a different axis order could be surprising.  Specifically, we could do something like
```
        # If we aren't in editable mode, prefer the channel axis to have a
        # stride of 1, then the z axis, then the t axis, then sorted by the
        # axis name.
        axisOrder = ((-'tzc'.index(k) if k in 'tzc' else 1, k)
                     for k in self._axes if k not in 'xys')
        # In editable mode, prefer the order that the axes are being written.
        if self._editable:
            axisOrder = ((-self._axes.get(k, 'tzc'.index(k) if k in 'tzc' else -1), k)
                         for k in self._axes if k not in 'xys')
        for _, k in sorted(axisOrder):
            ...
```

Note that this will cause `testFrameValues` test in test_sink.py to fail.

annehaley:
Resolves #1703.

### DIFF
--- a/test/test_sink.py
+++ b/test/test_sink.py
@@ -668,7 +668,6 @@ def testFrameValuesSmall(use_add_tile_args, tmp_path):
 
 @pytest.mark.parametrize('use_add_tile_args', [True, False])
 def testFrameValues(use_add_tile_args, tmp_path):
-    output_file = tmp_path / 'test.db'
     sink = large_image_source_zarr.new()
 
     frame_shape = (300, 400, 3)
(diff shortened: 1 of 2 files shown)


test/test_sink.py
```diff
@@ -741,10 +740,6 @@ def testFrameValues(use_add_tile_args, tmp_path):
     }
     compare_metadata(dict(sink.getMetadata()), expected_metadata)
 
-    sink.write(output_file)
-    written = large_image_source_zarr.open(output_file)
-    compare_metadata(dict(written.getMetadata()), expected_metadata)
```
manthey, line 746: Rather than getting rid of this part of the test, we could replace the `compare_metadata` line with
```
    assert written.metadata['IndexRange'] == expected_metadata['IndexRange']
    assert written.metadata['IndexStride']['IndexC'] == 1
```

manthey (approved):

PR #1838, opened 2025-02-24
Reformat docs extensions list as a table

annehaley:
This PR adds a table to the "Image Formats" page of the documentation. This table shows extensions and mimetypes for each tilesource. The resulting documentation page can be viewed here: https://output.circle-artifacts.com/output/job/a09981e2-a5f6-4872-865b-e447681b9331/artifacts/0/build/docs/formats.html

### DIFF
--- a/docs/generate_format_table.py
+++ b/docs/generate_format_table.py
@@ -110,30 +110,47 @@ def get_mimetypes_list():
 
 
 def get_extensions_list():
+    extensions = large_image.listExtensions()
+    return [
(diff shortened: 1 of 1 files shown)


manthey:
This looks good, but I wonder if we still want the complete list, too?  And, do we want to add the mimetypes into the table?  Some sources (gdal, notably) doesn't list many mimetypes so I also wonder how distinct some of the extensions are and if we could use them to glean mimetypes.

manthey (approved):

PR #1839, opened 2025-02-26
Histogram Editor: Allow inverted range

### LINKED ISSUE #1835: Allow histogram editor to specify inverted values
We set the minimum and maximum values allowed in the histogram editor.  But you might want to invert the pixel values (e.g., on a greyscale image have 0 be white and a high number be black).  We should handle this case, including allowing the drag handles to invert and still be accessible.  We may want to change how the drag handles area gets shaded when it is inverted.

Auto-range shouldn't allow inversion, though.

annehaley:
Resolves #1835


large_image/widgets/HistogramEditor.vue
```diff
@@ -1,33 +1,33 @@
...
-}
+    int8: [-128, 127],
+    uint8: [0, 255],
+    int16: [-32768, 32767],
+    uint16: [0, 65535],
+    int32: [-2147483648, 2147483647],
+    uint32: [0, 4294967295],
+    int64: [-9e18, 9e18],
+    uint64: [0, 1e19],
+    float16: [-65504, 65504],
```
manthey, line 27: I'm curious where 65504 came from, since when I ask numpy I get 
```
>>> np.finfo(np.dtype('float16')).min
np.float16(-65500.0)
>>> np.finfo(np.dtype('float16')).max
np.float16(65500.0)
```

annehaley: My fault for taking AI overview at face value: 
image
Let's go with what's reported by numpy. Changed in [4308f2e](https://github.com/girder/large_image/pull/1839/commits/4308f2ec93e55816d4b2543401149caf37e93712)

manthey: And, we might want 2 significant figures for int64 and uint64 (or round to a less zero value), since with uint64 with lost nearly half the range.

manthey:
Notice how the left edge of the shaded area jumps further left on inversion:

https://github.com/user-attachments/assets/a2c458a1-09de-4407-994f-c9e9861e5a06

This is more obvious in the HistomicsUI interface.

And, as a possibly separate issue, can we make it so when dragging we don't select text.

annehaley:
@manthey Could you elaborate? I don't see the change in the shaded area on the left. I do notice it on the right, though, so I fixed that behavior in [7033314](https://github.com/girder/large_image/pull/1839/commits/70333148dea4c6ebb98ede5c1d04ea3537523f95). I also disabled user-select while dragging in [b430e4a](https://github.com/girder/large_image/pull/1839/commits/b430e4a1697cad654eadac420a27bf6335a3d7d5).

manthey:
> @manthey Could you elaborate? I don't see the change in the shaded area on the left. I do notice it on the right, though, so I fixed that behavior in [7033314](https://github.com/girder/large_image/pull/1839/commits/70333148dea4c6ebb98ede5c1d04ea3537523f95). I also disabled user-select while dragging in [b430e4a](https://github.com/girder/large_image/pull/1839/commits/b430e4a1697cad654eadac420a27bf6335a3d7d5).

I meant right.  I don't know what I typed left.  I blame inversion.

manthey (approved):

PR #1844, opened 2025-03-04
Frame Selector: Queue/Debounce histogram requests

annehaley:


manthey:
Scrubbing around, I can still get to a state where there are a lot of histogram requests pending.
image

Issue #1106, opened 2023-04-12
Should geospatial sources open with a default projection?


banesullivan:
The current default projection is `None` for geospatial sources -- this has a specific meaning (to use image coordinates).

I'm wondering if geospatial sources should be opened with `EPSG:3857` by default, as this is intuitive and what most users probably do anyways.

Changing this default will likely have ramifications in downstream packages (I know it will affect RGD and `django-large-image` at least).

Similarly to #1105, this will improve useability of #1065 but is less critical

manthey:
Although having a default project would make sense in some contexts, I think this would be surprising in others.  From a tile serving endpoint, this makes sense.  From an algorithm endpoint, this doesn't (at least, I think many algorithms want to work on a pixel-wise basis and not a projected space basis).  Maybe the tile serving endpoints that don't specify a project could default to none for non-geospatial and EPSG:3857 for geospatial, but the pythonic interfaces would stay with none as the default.

banesullivan:
This makes sense. I need to look into if `getTile()` takes/supports `projection` as a keyword argument or could easily support it instead of using `self.projection`.

Perhaps for geospatial sources (at the class-level), `getTile()` would have a default `projection='EPSG:3857'` keyword argument while still opening with `projection=None` by default

manthey:
I think we currently open different projections as different sources (unlike with styles), so this _isn't_ a `getTile` parameter.

banesullivan:
Hm, then this might be too much effort to handle at this time

PR #1846, opened 2025-03-05
Add `default_encoding` and `default_projection` config values

### LINKED ISSUE #1105: Should `encoding="PNG"` be default?
I noticed that JPEG seems to be the default encoding when opening a tile source. With the addition of the Jupyter visualizer in #1065, it may benefit users to have `encoding="PNG"` be the default to avoid opaque/black tiles outside of the image bounds:

| Default | `encoding='PNG'` |
|---|---|
| Screen Shot 2023-04-12 at 11 06 04 AM | Screen Shot 2023-04-12 at 11 06 18 AM |

annehaley:
This PR adds two new configuration values `default_encoding` and `default_projection`. These new config values are also added to the "Configuration Options" documentation page. May resolve #1105 and #1106.

### DIFF
--- a/sources/gdal/large_image_source_gdal/__init__.py
+++ b/sources/gdal/large_image_source_gdal/__init__.py
@@ -26,6 +26,7 @@ from importlib.metadata import version as _importlib_version
 import numpy as np
 import PIL.Image
 
+from large_image import config
 from large_image.cache_util import LruCacheMetaclass, _cacheClearFuncs, methodcache
 from large_image.constants import (TILE_FORMAT_IMAGE, TILE_FORMAT_NUMPY,
                                    TILE_FORMAT_PIL, SourcePriority,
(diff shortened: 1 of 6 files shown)


sources/gdal/large_image_source_gdal/__init__.py
```diff
@@ -140,6 +141,8 @@ def __init__(self, path, projection=None, unitsPerPixel=None, **kwargs):  # noqa
         self.tileSize = 256
         self.tileWidth = self.tileSize
         self.tileHeight = self.tileSize
+        if projection is None:
+            projection = config.getConfig('default_projection')
```
manthey, line 145: We'll only want to do this IF the source is actually geospatial.  We currently have the isGeospatial method; this takes a path, so it should probably be broken into two methods where the current method calls a isDatasetGeospatial or the original function accepts a dataset or a path (and something has to be done equivalently in rasterio), and that if we are apply the default, we only do so IF the dataset is geospatial.

annehaley: Changed in [39e61be](https://github.com/girder/large_image/pull/1846/commits/39e61bea2aecab9268c68b17a1a7a470773c5b2c)

PR #1854, opened 2025-03-13
Follow-up to default encoding and projection

annehaley:


manthey:
I think (not certain) we also need to change the LRUHash code.  This `kwargs.get('projection', args[1] if len(args) >= 2 else None) or config.getConfig('default_projection')` will get the default _projection when it is None, but maybe we need to say 
```
proj = kwargs.get('projection', args[1] if len(args) >= 2 else constants.PROJECTION_SENTINEL)
proj = proj if proj != constants.PROJECTION_SENTINEL else config.getConfig('default_projection')
```

manthey (approved):

PR #1866, opened 2025-04-04
Use `unitsWH` in `getRegion` for geospatial images opened with a projection

### LINKED ISSUE #1863: Fix `getRegion` with `unitsWH` for geospatial images opened with a projection
When a geospatial image is opened with a projection and `getRegion` is called with arguments for `top`, `left`, `units`, `width`, `height`, and `unitsWH`, the result is a blank image.
Image

annehaley:
Resolves #1863 
This PR also expands `TileInputUnits` in `constants.py` to accept kilometers, meters, micrometers, and nanometers.


manthey:
Before, without a projection, I could do this:
```
import large_image
ts = large_image.open('./build/tox/externaldata/TC_NG_SFBay_US_Geo_COG.tif')
ts.getRegion(region=dict(top=37.84, left=-122.49, units='EPSG:4326', width=4000000, height=4000000, unitsWH='mm'), format='numpy')[0].shape
```
and get (289, 289, 4)
Now, I'm getting an empty array.  I'd expect the same result (and also using `width=4000, height=4000, unitsWH='m'`, and `width=4, height=4, unitsWH='km'`).

large_image/tilesource/base.py
```diff
@@ -418,6 +418,9 @@ def _scaleFromUnits(
                       desiredMagnification['mm_x'])
             scaleY = (desiredMagnification['scale'] /
                       desiredMagnification['mm_y'])
+            factors = dict(nm=1000000, um=1000, mm=1, m=0.001, km=0.000001)
```
manthey, line 421: If we define this dictionary higher up, then the `elif units in [...]:` would become `elif units in mmConversionDict:`

annehaley: Good point. Moved and renamed in [f39a716](https://github.com/girder/large_image/pull/1866/commits/f39a716d9d434c29e48fa14a827f216d313d02ec)

PR #1867, opened 2025-04-07
Pin numcodecs since we require zarr<3

manthey:


manthey:
For zarr 3, we either need to abandon the sqlite store (and thus any files written with it), or wait until someone writes a zarr 3 sqlite store.

PR #1876, opened 2025-04-14
Use a local task to ingest annotations

manthey:
Adjust the function so it can be reused by HistomicsUI


manthey:
This also merged master for the girder-5 branch.

PR #1878, opened 2025-04-14
Add an example notebook to show off the frame selector widget

manthey:
@annehaley After talking with @waxlamp, we thought it would be good to have a simple notebook to illustrate getting a frame selector.  The sample image has 24 channels.  Although we have some examples with both channels and z-stacks, I thought this was the right complexity for a demo.


annehaley (approved):
Great, I just tried the notebook locally and everything works as expected. Leaving a screenshot here for future reference: 
image

PR #1881, opened 2025-04-17
Allow specifying output path for `getRegion`

### LINKED ISSUE #682: Override outputPath for GDALFileTileSource.getRegion
Would it be okay to override `outputPath` for the GDAL tile source's `getRegion`?

https://github.com/girder/large_image/blob/d68aec03e303336231c1bce2f723d645c5a6a76a/sources/gdal/large_image_source_gdal/__init__.py#L1137

There are many circumstances where I want to store/use that file immediately and I don't want to have to copy the file out of a temp folder into my working space.

Also, I noticed once on my Mac that nearly immediately after the temp file was created by `getRegion`, it got deleted by the system as there was some sort of automatic cleanup.

annehaley:
When calling `getRegion`, if encoding is "TILED", then the user may also specify an output path for the tiled image. Example:

```
region, _ = ts.getRegion(
   output=dict(maxWidth=1024, maxHeight=1024, path='./path/to/region.tif'),
   encoding='TILED'
)
```

Resolves #682

### DIFF
--- a/sources/gdal/large_image_source_gdal/__init__.py
+++ b/sources/gdal/large_image_source_gdal/__init__.py
@@ -985,7 +985,7 @@ class GDALFileTileSource(GDALBaseFileTileSource, metaclass=LruCacheMetaclass):
             '-ts', str(int(math.floor(outWidth))), str(int(math.floor(outHeight))),
         ]
 
-        outputPath = kwargs.get('output').get('path')
+        outputPath = kwargs.get('output', {}).get('path')
         if outputPath is not None:
             outputPath = pathlib.Path(outputPath)
             outputPath.parent.mkdir(parents=True, exist_ok=True)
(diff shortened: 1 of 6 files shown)


sources/gdal/large_image_source_gdal/__init__.py
```diff
@@ -985,8 +985,13 @@ def getRegion(self, format=(TILE_FORMAT_IMAGE, ), **kwargs):
             '-ts', str(int(math.floor(outWidth))), str(int(math.floor(outHeight))),
         ]
 
-        fd, outputPath = tempfile.mkstemp('.tiff', 'tiledGeoRegion_')
-        os.close(fd)
+        outputPath = kwargs.get('output', {}).get('path')
```
manthey, line 988: Do you think it is worth pulling this out to a function and calling that from the three distinct spots?  If not, merge away.

manthey (approved):

sources/gdal/large_image_source_gdal/__init__.py
annehaley: Given that this snippet is slightly different in each of the three spots, it probably isn't worth the effort to generalize it as its own function

manthey: Then merge away.

Issue #1002, opened 2022-11-29
Improved image creation


manthey:
Currently, the vips source adds a `new` method that, when used with `addTile`, can take a series of images (PIL or numpy or vips) and output a file in a few formats.  The `addTile` method species where in the single frame image the new data is located and allows for masking the data as it is added.

We want to expand this capability:

- [x] Support arbitrary data axes to have multiple frames.  The base image can have different sample depths (currently L, LA, RGB, RGBA, but it would be nice to support arbitrary hyperspectal sample channels OR indexed data).  Frames could be arbitrary OR could have specific axes: C (channel, which should have channel names as optional metadata), Z (vertical Z stack), T (time), XY (also called P or P and Q for physical sample position).  We don't currently handle other axes in any source, but having arbitrary axes would make reading things like netcdf more obvious.
- [x] Support arbitrary numpy dtypes for the data
- [x] Support a variety of inputs for data and masks: PIL, numpy arrays, vips images
- [x] Set a variety of metadata: pixel scale (see #1482).
- [x] Set image description (see #1483)
- [x] Set channel names (#1484)
- [x] Support cropping the data.
- [x] Support padding the data to different sizes (#1831)
- [x] ~Add data at different scales (e.g., a lower resolutions from the base).  This could do a nearest neighbor mapping to the base resolution, but we might want to eventually add an option for how the data is scaled.  The general form of this would be to accept an affine transform.~  This is the purview of the multi-source source, and shouldn't be part of this issue.
- [ ] For geospatial data, specify a base projection and corner locations in some manner, and add an optional different projection when adding data.
- [x] Have entry-point based file output plugins for this.
- [x] Add associated images (#1485)
- [ ] Add arbitrary additional metadata.
- [x] Allow new images to have a optional path so that they could be created over multiple program runs.  This would require having a close method (preferably implicit) that saves enough state to resume image creation (#1487)

One method of storing this data internally would be a hdf5 file via the h5py module.  We could make a new hdf5 source (limited to just files we create) to allow reading the data as it is being created.

We always want the property that the last data added "wins", overwriting any existing data at its location.  If the added data has a mask, it should only change the appropriate locations.  If it has an alpha channel, this should be applied as it is added.  As part of this, we may want to have a default alpha channel that is transparent until any data is added.  Note that if data is added to the same location multiple times with alpha values that are 0 < a < 1, we need to make sure that they mix in the expected proportion.

[...]

PR #1882, opened 2025-04-21
Zarr sink: Set projection and GCPs

annehaley:
This PR allows users to set projection and GCPs on an image written with the zarr sink. Example below. This PR will complete the last remaining item in #1002.

```
sink.addTile(np.random.random((256, 256, 3)) * 10, x=0, y=0)
sink.projection = 'epsg:4326'
sink.gcps = [(-77.050, 38.888, 0, 0), (-77.049, 38.889, 256, 256)]
sink.write('example.tiff')
```

### DIFF
--- a/sources/zarr/large_image_source_zarr/__init__.py
+++ b/sources/zarr/large_image_source_zarr/__init__.py
@@ -1324,30 +1324,21 @@ class ZarrFileTileSource(FileTileSource, metaclass=LruCacheMetaclass):
             self._validateZarr()  # refresh self._levels before continuing
 
     def _applyGeoReferencing(self, path):
-        if self.projection is not None and self.gcps is None:
+        if self.projection and not self.gcps:
             msg = (
                 'Projection was specified but GCPs were not specified. '
-                'Output will not be georeferenced.'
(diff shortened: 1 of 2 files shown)


sources/zarr/large_image_source_zarr/__init__.py
```diff
@@ -1249,6 +1323,32 @@ def _generateDownsampledLevels(self, resample_method):
...
+        if (
+            self.projection is not None and
+            self.gcps is not None and
+            len(self.gcps)
+        ):
+            import tifftools
+
+            tifftools.tiff_set(path, setlist=[
+                ('projection', self.projection),
+                ('gcps', self.gcps),
```
manthey, line 1349: I've used gcps without a projection.  Specifically, I have two images and I want to warp one image to match the other.  Neither is geospatial.  I can add gcps on one image and then the gdalwarp will properly reproject the image as I desire (and, eventually, I'd like to make the multisource use scikit-image do this as a tps or affine warp, too).  I'd change this to not complain if there is no projection, and to just add gcps when there is no projection specified.

PR #1884, opened 2025-04-22
979 annotation list checkboxes

### LINKED ISSUE #979: Annotation list checkboxes
In the Girder interface, for annotation lists in items, it would be convenient to have check boxes on each row and a top-level check box to check/uncheck all and then have the top-level controls (permissions, delete, download) apply only to those annotations that are checked.  This would make it easier to delete multiple but not all annotations at a time, which would be useful in some of the guided labelling work.

marySalvi:
Closes #979 
In addition to adding the functionality of being able to select all annotations this also updates the global actions to perform on any selected annotations. As such the global actions are disabled if there are no annotations selected. Finally the global actions have been moved to the table header to match the view annotations action styling.


girder_annotation/girder_large_image_annotation/web_client/templates/annotationListWidget.pug
```diff
@@ -5,21 +5,12 @@
...
-        i.icon-cancel
 
 if annotations.length
   table.g-annotation-list.table.table-hover.table-condensed
     thead
-      //
-        th.g-annotation-select
-          input.g-select-all(type='checkbox')
+      th.g-annotation-select
+        input(type='checkbox' title='Select all annotations for bulk actions' class='pr-2')
```
manthey, line 13: I think linting will demand this to be `input.pr-2(type='checkbox', title='Select all annotations for bulk actions')`

manthey:
You can fix the tests with a diff like
```
diff --git a/girder_annotation/test_annotation/web_client_specs/annotationListSpec.js b/girder_annotation/test_annotation/web_client_specs/annotationListSpec.js
index db2bc6a3..d7f4f536 100644
--- a/girder_annotation/test_annotation/web_client_specs/annotationListSpec.js
+++ b/girder_annotation/test_annotation/web_client_specs/annotationListSpec.js
@@ -372,8 +372,9 @@ describe('AnnotationListWidget', function () {
             }, 'the upload to finish');
         });
         it('test header permission editor link', function () {
-            var $el = $('.g-annotation-list-header');
-            expect($el.find('.g-annotation-permissions').length).toBe(1);
+            var $el = $('.g-annotation-list');
+            expect($el.find('th .g-annotation-permissions').length).toBe(1);
+            $el.find('th.g-annotation-select input').click();
             $el.find('.g-annotation-permissions').click();

             girderTest.waitForDialog();
@@ -384,8 +385,9 @@ describe('AnnotationListWidget', function () {
             girderTest.waitForLoad();
         });
         it('test header delete link', function () {
-            var $el = $('.g-annotation-list-header');
-            expect($el.find('.g-annotation-delete').length).toBe(1);
+            var $el = $('.g-annotation-list');
+            expect($el.find('th .g-annotation-delete').length).toBe(1);
+            $el.find('th.g-annotation-select input').click();
             $el.find('.g-annotation-delete').click();

             girderTest.waitForDialog();
```
This shows up that deleting multiple annotations fails when we try to show the delete dialog.

Also, if _no_ annotations are selected, the permissions dialog can still be shown.

manthey:
The current error is because https://github.com/girder/large_image/blob/979-annotation_list_checkboxes/girder_annotation/girder_large_image_annotation/web_client/views/annotationListWidget.js#L262 fails when there is more than one item to delete.

girder_annotation/girder_large_image_annotation/web_client/views/annotationListWidget.js
```diff
@@ -182,24 +193,95 @@ const AnnotationListWidget = View.extend({
...
-            return;
+            for (let i = 0; i < checkedAnnotations.length; i++) {
+                const annotationId = $(checkedAnnotations[i]).parents('.g-annotation-row').data('annotationId');
+                checkedAnnotationIds.push(annotationId);
+            }
+            if (checkedAnnotations.length !== 0) {
+                confirm({
+                    text: `<h3>Are you sure you want to delete the following annotations?</h3>
+                        <ul
+                          style="max-height: 200px; padding-left: 0; overflow-y: auto;"
```
manthey, line 259: It is better to style via css classes than inline, because we can reuse components or have them appear in different places.  Also, I think this is better without the max-height restriction, since the dialog's max-height will be restricted elsewhere.  Otherwise, use a form like `calc(100vh - 200px)` so it can fill more of the screen (when I have a lot of annotations, it seems sub-optimal to scroll when it could all be visible).

girder_annotation/test_annotation/web_client_specs/annotationListSpec.js
```diff
@@ -384,8 +385,9 @@ describe('AnnotationListWidget', function () {
             girderTest.waitForLoad();
         });
         it('test header delete link', function () {
-            var $el = $('.g-annotation-list-header');
-            expect($el.find('.g-annotation-delete').length).toBe(1);
+            var $el = $('.g-annotation-list');
+            expect($el.find('th .g-annotation-delete').length).toBe(1);
+            $el.find('th.g-annotation-select input').click();
             $el.find('.g-annotation-delete').click();
```
manthey, line 391: This needs to be changed to `$el.find('th .g-annotation-delete').click();` (and we should do the same for the permissions check above).  Otherwise, we are trying to click ALL of the delete controls at the same time.  Whichever happens to respond first will be the dialog that appears, and if isn't the header control, then we only delete one annotation, not all of them.

girder_annotation/test_annotation/web_client_specs/annotationListSpec.js
```diff
@@ -374,8 +374,8 @@ describe('AnnotationListWidget', function () {
         it('test header permission editor link', function () {
             var $el = $('.g-annotation-list');
             expect($el.find('th .g-annotation-permissions').length).toBe(1);
-            $el.find('th.g-annotation-select input').click();
-            $el.find('.g-annotation-permissions').click();
+            $el.find('th .g-annotation-select input').click();
```
manthey, line 377: You _don't_ want the space here.  The `.g-annotation-select` class is on the `th` element (whereas the `.g-annotations-permissions` class is on the control within its `th` element.

manthey:
When downloading multiple annotations, this behaves unexpectedly: before, selecting download all annotations downloaded them in a single file with a name based on the item name, and downloading one annotation downloaded a single file named based on the annotation.  Now, selecting multiple (or all) downloads multiple files.

At the very least we should download a single file if all of the annotations are selected.  I _think_ that we should always download a single file which would just be a list of the annotations, but, if we don't currently have an endpoint to do so, downloading multiple files is fine (but definitely want to download a single file when downloading all of them).

girder_annotation/girder_large_image_annotation/web_client/views/annotationListWidget.js
```diff
@@ -182,24 +193,104 @@ const AnnotationListWidget = View.extend({
...
-                        method: 'DELETE'
-                    }).done(() => {
-                        this.collection.fetch(null, true);
+            if ($el.find('input').is(':checked')) {
+                allChecks.prop('checked', true);
+            }
+            if (!$el.find('input').is(':checked')) {
+                allChecks.prop('checked', false);
+            }
+        }
```
manthey, line 217: After checking the check-all box, if you uncheck another box, the check-all box is still checked.  I thin we want an else to the `if (!id)` that would change the check-all box to be checked or unchecked based on if all of the other boxes are checked.

marySalvi: updated in 256499718c11e0d6c10c437ab357c77d7d67e8ff

girder_annotation/girder_large_image_annotation/web_client/views/annotationListWidget.js
```diff
@@ -182,24 +193,109 @@ const AnnotationListWidget = View.extend({
...
-                text: `Are you sure you want to delete <b>ALL</b> annotations?`,
-                escapedHtml: true,
-                yesText: 'Delete',
-                confirmCallback: () => {
-                    restRequest({
-                        url: `annotation/item/${this.model.id}`,
-                        method: 'DELETE'
-                    }).done(() => {
-                        this.collection.fetch(null, true);
+            if ($el.find('input').is(':checked')) {
```
manthey, line 211: `is(':checked')` only checks the first element, so I'm not sure what this is actually doing (at least rechecking all the individual boxes doesn't turn the all-box back on.  I probably would have done `allChecks.prop('checked', $el.find('.g-annotation-row .g-annotation-select input:not(:checked)').length === 0)` instead of the pair of if statements (I didn't test my code).

PR #1885, opened 2025-04-22
Upload cached thumbnails and images in threads to parallelize

manthey:
This allows file writes to happen in the background, return the results more quickly.

### DIFF
--- a/girder/girder_large_image/models/image_item.py
+++ b/girder/girder_large_image/models/image_item.py
@@ -39,6 +39,9 @@ from .. import constants, girder_tilesource
 
 
 class ImageItem(Item):
+    _cacheSaveDataLock = threading.RLock()
+    _cacheSaveDataThreads = []
+
     # We try these sources in this order.  The first entry is the fallback for
     # items that antedate there being multiple options.
     def initialize(self):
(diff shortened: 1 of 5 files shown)


girder/girder_large_image/models/image_item.py
```diff
@@ -39,6 +39,9 @@
 
 
 class ImageItem(Item):
+    _cacheSaveDataLock = threading.RLock()
```
zachmullen, line 42: @manthey can we use Girder's cache abstraction here (dogpile) so that this plugin is scalable and portable to different execution environments? The use of threading makes me nervous about portability.

manthey: We are trying to add an associated file to the database, so that a subsequent query doesn't recompute something.  This should be persistent (e.g., a histogram isn't going change once computed), but, since this a cache, a failure isn't particularly critical (it is functionally a cache miss).  These artifacts could be big (e.g., a tile frames object could be an 8k x 8k jpeg).  In the field on an s3 assetstore, the slow point was often storage, so we wanted to return the result when it was ready but lazily cache it so if another requests asks for it, we don't recompute it.

I don't actually know what dogpile is doing or how (or if) it persists data between process restarts.

zachmullen: I'd suggest using a cache abstraction that will scale (e.g. memchaced or redis), and preventing stampedes using a key in that cache indicating that the computation is in progress and doesn't need to be repeated. That way we can ensure this code is useful even in multi-process / multi-node deployments.

PR #1915, opened 2025-05-21
In many instances, save annotation element via a PATCH call

manthey:
This greatly reduces serialization time and web traffic.


manthey:
Further refinements could be to (a) not save the primary annotation records unless they have changed and (b) allow sending partial elements to reduce traffic for things like large pixelmap updates.

PR #1922, opened 2025-05-27
`getGeospatialRegion` Function

annehaley:
This PR adds `getGeospatialRegion` as a function on the `TileSource` base class and adds an example to the documentation.

Example usage:
```
import large_image
source = large_image.open('sample.tiff')
source.getGeospatialRegion(
    'epsg:3857', 
    [
        (-8578909.8696,4704125.8132, 0, 0),
        (-8571075.0742,4710164.3385, source.sizeX, source.sizeY),
    ],
    'epsg:4326', 
    {
          'left': -77.015104,
          'top': 38.887642,
          'right': -77.005877,
          'bottom': 38.892235,
    },
    frame=0,
    format='numpy',
)
```

### DIFF
--- a/large_image/tilesource/base.py
+++ b/large_image/tilesource/base.py
@@ -2180,7 +2180,6 @@ class TileSource(IPyLeafletMixin):
         import pyproj
         import rasterio
 
-
         if any(len(gcp) != 4 for gcp in src_gcps):
             msg = 'Ground control points must contain four values in the form (cx, cy, px, py).'
             raise ValueError(msg)
--- a/large_image/tilesource/base.py
+++ b/large_image/tilesource/base.py
@@ -2202,11 +2202,16 @@ class TileSource(IPyLeafletMixin):
         ) for gcp in converted_gcps]
 
         # transform dest_region to pixel coords
-        pixel_to_world = rasterio.transform.from_gcps(gcps)
-        world_to_pixel = ~pixel_to_world
-        px1, py1 = world_to_pixel * (dest_region.get('top'), dest_region.get('left'))
-        px2, py2 = world_to_pixel * (dest_region.get('bottom'), dest_region.get('right'))
-        pixel_region = dict(left=px1, top=py1, right=px2, bottom=py2)
+        transformer = rasterio.transform.GCPTransformer(gcps)
+        px1, py1 = transformer.rowcol(dest_region.get('top', 0), dest_region.get('left', 0))
+        px2, py2 = transformer.rowcol(dest_region.get('bottom', 0), dest_region.get('left', 0))
+        px3, py3 = transformer.rowcol(dest_region.get('top', 0), dest_region.get('right', 0))
+        px4, py4 = transformer.rowcol(dest_region.get('bottom', 0), dest_region.get('right', 0))
+        left = min(px1, px2, px3, px4)
+        top = min(py1, py2, py3, py4)
+        right = max(px1, px2, px3, px4)
+        bottom = max(py1, py2, py3, py4)
+        pixel_region = dict(left=left, top=top, right=right, bottom=bottom)
 
         # send pixel_region into getRegion
         return self.getRegion(region=pixel_region, **kwargs)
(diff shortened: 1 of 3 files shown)


large_image/tilesource/base.py
```diff
@@ -2147,6 +2147,65 @@ def tileFrames(
...
+        :param src_gcps: A list of ground control points describing projected coordinates for
+            certain pixel coordinates in the image. Each GCP can be a list or tuple with the
+            following format: (cx, cy, px, py) where (cx, cy) is a projected coordinate in the
+            coordinate reference system described by src_projection and (px, py) is a pixel
+            coordinate within the extents of the image.
+        :param dest_projection: A string describing the coordinate reference system used for
+            dest_region. This string can be an EPSG code or other format accepted
+            by pyproj.CRS.from_string.
+        :param dest_region: A dictionary describing the desired region to retrieve from the image.
+            Should specify values for "top", "bottom", "left", and "right" in the projected
```
manthey, line 2176: I'd change the wording from `Should` to `Must` (or `Shall`), as these aren't optional.   But, see other comment.

large_image/tilesource/base.py
```diff
@@ -2147,6 +2147,65 @@ def tileFrames(
...
+            y=gcp[1],
+            row=gcp[2],
+            col=gcp[3],
+        ) for gcp in converted_gcps]
+
+        # transform dest_region to pixel coords
+        pixel_to_world = rasterio.transform.from_gcps(gcps)
+        world_to_pixel = ~pixel_to_world
+        px1, py1 = world_to_pixel * (dest_region.get('top'), dest_region.get('left'))
+        px2, py2 = world_to_pixel * (dest_region.get('bottom'), dest_region.get('right'))
```
manthey, line 2203: This assumes that the projected coordinates are not rotated with respect to the original pixel space.  But, if they are rotated, this could have very surprising results (e.g., rotate a square 45 degrees, any you will end up with a region with a 0 width or a 0 height depending on the rotation direction).

I think that, rather, you could do
```
px1, py1 = world_to_pixel * (dest_region.get('top', 0), dest_region.get('left', 0))
px2, py2 = world_to_pixel * (dest_region.get('bottom', 0), dest_region.get('left', 0))
px3, py3 = world_to_pixel * (dest_region.get('top', 0), dest_region.get('right', 0))
px4, py4 = world_to_pixel * (dest_region.get('bottom', 0), dest_region.get('right', 0))
left = min(px1, px2, px3, px4)
... etc
```

And, a question, is rasterio _guaranteed_ to expect the projection to go from y, x to x, y?  I thought (perhaps erroneously) that this depended on the selected projection.

large_image/tilesource/base.py
```diff
@@ -2147,6 +2147,65 @@ def tileFrames(
...
+        import pyproj
+        import rasterio
+
+        # convert gcps to dest_projection
+        crs_transform = pyproj.Transformer.from_crs(src_projection, dest_projection)
+        converted_gcps = [
+            [
+                *crs_transform.transform(gcp[0], gcp[1]),
+                gcp[2], gcp[3],
+            ] for gcp in src_gcps if len(gcp) == 4
```
manthey, line 2189: I think we should raise an error if len(gcp) != 4

manthey (changes requested):

large_image/tilesource/base.py
annehaley: Added check in [f12a77e](https://github.com/girder/large_image/pull/1922/commits/f12a77e09c18feb324012101389495af11ab1cf9)

large_image/tilesource/base.py
annehaley: Good point. I applied your suggested change and I also updated the definition of the transform in [5dc09e8](https://github.com/girder/large_image/pull/1922/commits/5dc09e87240042218b810f366e32d7645296de0c). I followed [these docs](https://rasterio.readthedocs.io/en/stable/topics/transforms.html#using-affine-transformation-matrix), which explain that we can use the `rowcol` function to be more explicit about x-y ordering.

PR #1947, opened 2025-07-02
Multi-source Thin Plate Spline Warping

annehaley:


sources/multi/large_image_source_multi/__init__.py
```diff
@@ -649,8 +661,14 @@ def _sourceBoundingBox(self, source, width, height):
         m[1][2] = pos.get('y', 0)
         if 'warp' in pos and skimage_transform is not None:
             warp = pos.get('warp')
-            warp_src = np.array(warp.get('src') or []).astype(float)
-            warp_dst = np.array(warp.get('dst') or []).astype(float)
+            warp_src = warp.get('src')
+            warp_dst = warp.get('dst')
+            if isinstance(warp_src, dict):
+                warp_src = list(warp_src.values())
```
manthey, line 667: There is no guarantee that the keys are in the same order.  If src and dst are both dicts, we want to correspond them by key and skip the points without matching keys in the other dict.  If they are both dicts, this might be:
```
keys = list(set(warp_src.keys()) & set(warp_dst.keys()))
warp_src = [warp_src[key] for key in keys]
warp_dst = [warp_dst[key] for key in keys]
```

annehaley: Should we not accept a schema where only one is specified as a dict (e.g. src points as a dict and dst points as a list)?

manthey: Yes, let's prohibit one list and one dict in the schema -- that is bound to be confusing.

annehaley: Added validation and a test in [16ce154](https://github.com/girder/large_image/pull/1947/commits/16ce154731cd62dc70bd79475bdae047ea8929ad)

sources/multi/large_image_source_multi/__init__.py
```diff
@@ -270,6 +285,41 @@ def _lazyImport():
                     # TODO: Add polygon option
                     # TODO: Add postTransform option
                 },
+                'warp': {
+                    'type': 'object',
+                    'properties': {
+                        'src': {
```
manthey, line 292: We could enforce the both-must-be-the-same type in the schema rather than in the code:
```
                'warp': {
                    'type': 'object',
                    'oneOf': [{
                        'properties': {
                            'src': {
                                'type': 'array',
                                'items': {
                                    'type': 'array',
                                    'items': {'type': 'number'},
                                    'minItems': 2,
                                    'maxItems': 2,
                                },
                            },
                            'dst': {
                                'type': 'array',
                                'items': {
                                    'type': 'array',
                                    'items': {'type': 'number'},
                                    'minItems': 2,
                                    'maxItems': 2,
                                },
                            },
                        },
                        'required': ['src', 'dst'],
                        'additionalProperties': False,
                    }, {
                        'properties': {
                            'src': {
                                'type': 'object',
                                'additionalProperties': {
                                    'type': 'array',
                                    'items': {'type': 'number'},
                                    'minItems': 2,
                                    'maxItems': 2,
                                },
                            },
                            'dst': {
                                'type': 'object',
                                'additionalProperties': {
                                    'type': 'array',
                                    'items': {'type': 'number'},
                                    'minItems': 2,
                                    'maxItems': 2,
                                },
                            },
                        },
                        'required': ['src', 'dst'],
                        'additionalProperties': False,
                    }],
                },
```

manthey: Good point.

annehaley: Descriptions added in [20c4e6a](https://github.com/girder/large_image/pull/1947/commits/20c4e6a7dd1b44955865a3d289aba8ff368d3cea)

PR #1959, opened 2025-07-22
Jupyter widget warp editor mode

annehaley:


large_image/tilesource/jupyter.py
```diff
@@ -157,6 +163,15 @@ def iplmap(self) -> Any:
...
             return self._map.map
 
+    @property
+    def warp_points(self):
+        return self._map.warp_points
+
+    @property
+    def image_path(self):
+        #  TODO: is there a better way to get the path value?
+        return str(self._initValues[0][0])  # type: ignore[attr-defined]
```
manthey, line 173: Tile sources have `self.largeImagePath` (though, strictly, this could be an object since you can pass a python dictionary and the multi-source can open in).

annehaley: Switched to using `largeImagePath` in [586c88d](https://github.com/girder/large_image/pull/1959/commits/586c88d1a56e342196874fb611132d48b9c58346).

PR #1987, opened 2025-09-24
Skip adjacent DICOM file search during import to improve performance

PaulHax:
When noCache=True, skip searching for adjacent files in the folder. This avoids creating hundreds of S3 adapters per file during import.


manthey:
This is fine, but after this gets merged, I'll make an issue to revisit this.  We have a canRead method which really just tries to open an image and checks for failure.  In the girder instance, we open with the noCache option.  The virtue of canRead and the default no-cache open just being regular open is that one avoids issues where an image will partially open but fail when more fully opened (for instance, there are some multi-file ome tiffs that can open one file but not all of them).  There is clearly a balance between open-check/canRead speed and fully opening a file for use.

PR #2072, opened 2026-05-14
Aaron/eager iter

amrosado:


amrosado:
@manthey can you inspect the failing on these tests? it looks like it is running into issues with the circleci environment and it maybe hitting some limits.

manthey:
If you merge master (or at least #2074), the type error in CI will go away.  mypy had an update that made of our casts needless (it always should have been).

amrosado:
@manthey  This is ready for your review.

amrosado:
@manthey there will not be additional changes until I have the time to make example notebooks and the other things we discussed

large_image/tilesource/eager_utils/eager_image_modifications.py
```diff
@@ -0,0 +1,355 @@
...
+        out = pad_color(out, top, bottom, left, right, constant_color)
+    else:
+        msg = "Invalid pad_mode value. Must be 'wsi_edge' or 'equal'."
+        raise ValueError(msg)
+
+    return out
+
+
+def return_constant_color(image: np.ndarray, pad_fill_mode: Any):
+    # Determine the color that will be used for paadding
```
manthey: This comment seems redundant

large_image/tilesource/eager_utils/eager_fn.py
```diff
@@ -0,0 +1,42 @@
...
+from typing import Callable, Optional
+
+_eager_iter_transform: Optional[Callable] = None
+_eager_iter_transform_scale: Optional[Callable] = None
+
+
+def set_transform(fn: Callable) -> None:
+    """Store the process-local eager transform callable.
+
+    :param fn: Callable to apply to eager tiles in worker processes.
```
manthey: Can you document what the function takes and returns (and also the scale function)?

There is no need to change it, but rather than using `global var`, in other places we used a construct like `_eager_fns = {'transform': None, 'transform_scale': None}`, and then set the dictionary entries instead of the package global values

test/eager/dtype/test_numpy_dtype.py
```diff
@@ -0,0 +1,19 @@
+import numpy as np
+import pytest
+
+from ...datastore import datastore
+from ..eager_helpers import run_eager_iterator_numpy_dtype
+
+
+@pytest.mark.singular
```
manthey: Is there a reason these tests have been marked singular?  And, would switching to a smaller source file speed up the tests and still be sufficient?

manthey:
There are two images that are added at the base repo level: `low_res_img.png` and `mask.png`.  I don't see where these are being used.  If they are needed they should be placed in the `test/test_files` folder (though I'd argue that the low_res_img.png is large enough it should be in the pooch datastore).  And, unless it will change the test, I prefer to run oxipng on all pngs before checking them in to make them as small as possible.

large_image/tilesource/eager_utils/eager_read_args.py
```diff
@@ -0,0 +1,389 @@
...
+    # Determine if the grid is complete or incomplete even in the case of edges being removed
+    if (sorted_tiles.shape[0] + diff_from_edge) == n_possible_tiles:
+        chunks = gen_read_args_complete_grid(sorted_tiles, chunk_mult)
+    else:
+        chunks = gen_read_args_incomplete_grid(sorted_tiles, chunk_size)
+
+    return chunks
+
+
+def default_region_coords_and_target_scale_from_read_args(
```
manthey, line 335: This function has no coverage.  If it is quick to add, a test would be great.

manthey:
@amrosado Go ahead and make adjustments as per my comments.  I'm sure there will be future tweaks, but let's get this in.

large_image/tilesource/eager_utils/eager_image_modifications.py
amrosado: Removed

large_image/tilesource/eager_utils/eager_fn.py
amrosado: Changed.

large_image/tilesource/eager_utils/eager_read_args.py
amrosado: Added coverage.

test/eager/dtype/test_numpy_dtype.py
amrosado: Removed singular here.

amrosado:
@manthey  This looks like it checks out now.  I tried to incorporate all your suggestions.

manthey (approved):

manthey:
@amrosado I will do a squash-and-merge on this to flatten the history; the full commit history still exists but it will main the master branch a little clearer to look through.

manthey:
This is now in large_image 1.35.0, DOI https://doi.org/10.5281/zenodo.21133125

PR #2101, opened 2026-07-22
Fix subtile dx/dy slicing

Harold2017:
- fix subtile dx/dy slicing
- use tile.getbands() instead of tile.mode


manthey:
Thank you, @Harold2017.  We should have always been using `getbands()` instead of mode -- I think there is one other place it needs to change.  I'll add a test after we get this merged (CI should pass once merged with master).

manthey (approved):

PR #2120, opened 2026-08-27
Find public config files blocked by private ones

naglepuff:
Prevents returning a `401`/`403` when retrieving a YAML config if an attempt is made to read a config file in a private folder.

Instead, the traversal of the hierarchy skips folders that are inaccessible to the current user, and continues to find publicly available config files (up to and including the root collection/user and the .config directory).

### DIFF
--- a/girder/girder_large_image/__init__.py
+++ b/girder/girder_large_image/__init__.py
@@ -596,9 +596,13 @@ def yamlConfigFile(folder, name, user):
     """
     addConfig = None
     last = False
+    isGlobalConfig = False
     while folder:
         item = Item().findOne({'folderId': folder['_id'], 'name': name})
-        if item:
+        # If the user can't access the folder and it is not a global config mechanism,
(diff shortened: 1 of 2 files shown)


girder/girder_large_image/__init__.py
```diff
@@ -596,9 +596,13 @@ def yamlConfigFile(folder, name, user):
...
     addConfig = None
     last = False
+    isGlobalConfig = False
     while folder:
         item = Item().findOne({'folderId': folder['_id'], 'name': name})
-        if item:
+        # If the user can't access the folder and it is not a global config mechanism,
+        # skip this folder. Keep walking up the chain to merge other publically available
+        # configuration in the hierarchy.
+        if item and (isGlobalConfig or Folder().hasAccess(folder, user, AccessType.READ)):
```
manthey, line 605: It took me a while to remember the logic here. We always permit the global config.  This comment is really just a note to my future self.

manthey (approved):

